Issue #3, opened 2024-02-07
Latent processes


SamuelBrand1:

SamuelBrand1:
_Background concept_: The main logic/workhorse API of `Turing` is the macro `@model`. `@model` is expecting a set of code that looks like a standard function definition but with a stan-like `~` symbol.

The meaning of `X ~ dist` where `dist` is a [distribution](https://juliastats.org/Distributions.jl/stable/) is either:
1. If `X` has been defined with a value, it is treated as an observation with likelihood `pdf(dist, X)`.
2. If `X` has either not been defined or defined as having the value `missing` then it is treated as an unknown parameter/latent variable with prior `dist`.

**NB: this is the same as the numpyro concept of variables having either a determined value (obs data) or value `None` (latent variable)**.

Example:

```julia
@model function my_model(...some args...)
 X ~ some_dist_obj
...some code...
return generated_quantities
end
```

tl; dr version: Turing is a mash up of stan-like syntactic sugar and Numpyro-like data = value/variable = `None` idea.

_Latent process concept_: 

The latent processes that we want to test are each defined as a Turing model. This mean they can be sampled from independently and tested using the standard API to `model` usage.

However, for the overall inference problem against data they are given as arguments to another Turing model which covers the data observation process, and we use the [`@submodel`](https://turinglang.org/DynamicPPL.jl/stable/api/#DynamicPPL.@submodel) macro which allows nesting of Turing models inside other Turing models (honestly this just works).

Reason for this choice is code modularity. We can define and test each model separately, then glue the bits together (and end-to-end test).

SamuelBrand1:
`random_walk` is defined [here](https://github.com/CDCgov/Rt-without-renewal/blob/3c06d369eadd532976a268962b2fb715f5e7f99e/src/latent-processes.jl#L16).

The model defines sampling from a RW std value (default is HalfNormal(0.05)) and generating a RW. Unit test is [here](https://github.com/CDCgov/Rt-without-renewal/blob/main/test/test_latent-processes.jl).

SamuelBrand1:
##### Latent processes to do:

- [ ] RW
- [ ] Spline
- [ ] Difference AR(1)

seabbs:
> Reason for this choice is code modularity. We can define and test each model separately, then glue the bits together (and end-to-end test).

Agree this seems like a natural approach to me and is a good showcase of Julia imo. 

> Latent processes to do:

[...]

PR #5, opened 2024-02-07
Bring `random_walk` into main

SamuelBrand1:
[`random_walk`](https://github.com/CDCgov/Rt-without-renewal/blob/ec5b11ad713f482dd311dab1b78e0b82daa4ede7/src/latent-processes.jl#L1C1-L27C4) brings a random walk with random noise variance into main.

Unit test is [here](https://github.com/CDCgov/Rt-without-renewal/blob/SamuelBrand1/issue2/test/test_latent-processes.jl).

Prior predictive checking is [here](https://github.com/CDCgov/Rt-without-renewal/blob/SamuelBrand1/issue2/test/prior_predictive_checking/ppc-latent-processes.jl).

This is aimed at addressing the RW task [here](https://github.com/CDCgov/Rt-without-renewal/issues/3#issuecomment-1931801274).


src/latent-processes.jl
```diff
@@ -1,27 +1,43 @@
 """
-    random_walk(n, ϵ_t = missing; latent_process_priors = (σ_RW_dist = truncated(Normal(0., 0.05), 0., Inf),), ::Type{T} = Float64) where {T <: Real}
+    random_walk(n, ϵ_t = missing; latent_process_priors = (var_RW_dist = truncated(Normal(0., 0.05), 0., Inf),), ::Type{T} = Float64) where {T <: Real}
```
seabbs, line 2: Why do we want to define this in terms of the variance and not the standard deviation?

src/latent-processes.jl
```diff
@@ -1,27 +1,43 @@
...
     rw = Vector{T}(undef, n)
     ϵ_t ~ MvNormal(ones(n))
-    σ_RW ~ latent_process_priors.σ_RW_dist
+    σ²_RW ~ latent_process_priors.var_RW_dist
+    σ_RW = sqrt(σ²_RW)
     rw .= cumsum(σ_RW * ϵ_t)
     return rw, (; σ_RW)
 end
+
+# @model function cubic_spline_basis(
```
seabbs, line 29: is this an example? Why is it commented?

seabbs: If we can I woiuld like to avoid commented out code so this could be in an issue

test/prior_predictive_checking/ppc-latent-processes.jl
```diff
@@ -0,0 +1,67 @@
...
+
+plt_rw = plot(
+    plt_ppc_rw,
+    σ_hist,
+    layout = (1, 2),
+    size = (800, 400),
+    left_margin = 3mm,
+    bottom_margin = 3mm,
+)
+savefig(plt_rw, joinpath(@__DIR__(), "assets", "ppc_rw.png"))
```
seabbs, line 67: missing end of line?

seabbs (commented):
Looks overall good with a few comments

src/latent-processes.jl
SamuelBrand1: Its standard for HalfNormals to be priors for variance rather than std. Since it doesn't make much different I think going with the literature norm is probably best upon reflection.

src/latent-processes.jl
SamuelBrand1: Thats an ooops... because I stopped in the middle of writing the next gen process.

src/latent-processes.jl
seabbs: okay that makes sense though I would note its not standard in other PPLs (i.e. stan) and so if aiming for interoperability this could be confusing. I am happy with either choice though.

SamuelBrand1: Ah right, I thought that was the stan default.

seabbs (approved):
LGTM.

seabbs (approved):
LGTM.

Issue #7, opened 2024-02-07
Better double censoring approach to discretising continuous distributions


SamuelBrand1:
The appropriate way to discretise a continuous distribution's pdf into a probability vector is still a somewhat open question.

Our default was the approach at [epinowcast](https://package.epinowcast.org/articles/distributions.html), however, it would be preferable to go to the more recent solution implemented [here](https://github.com/epinowcast/epinowcast/blob/3b2e94f269690dbbc4d1170db0aafffdb025253b/R/model-module-helpers.R#L293).

This can be directly translated into Julia.

seabbs:
> The appropriate way to discretise a continuous distribution's pdf into a probability vector is still a somewhat open question.

I'm not sure this is true based on https://www.medrxiv.org/content/10.1101/2024.01.12.24301247v1 its more that most of the current implementations of the theory have issues that have yet to be corrected or those corrections have yet to make it into releases.

That being said above point doesn't impact the work proposed here.

SamuelBrand1:
I would say 

> current implementations of the theory have issues that have yet to be corrected or those corrections have yet to make it into releases.

== appropriate way is still an open Q

SamuelBrand1:
Question for @seabbs @zsusswein 

Given that we have some distributions where the delay after censoring can be `0` (e.g. primary and secondary events occur on same day), do you want arrays that index on `[0]`  using [OffsetArrays](https://github.com/JuliaArrays/OffsetArrays.jl)?

It doesn't hurt anything code wise and might avoid some gotchas when indexing?

seabbs:
Hmm I see the interpretability benefit but in my head, I am now very much conditioned to think about up to day 1 now!

I would be happy to stick with the simple option of having to fudge the indexes a bit but I don't feel that strongly

SamuelBrand1:
Closed by #10

seabbs:
In the futuree if you say something like close #issue-number it will auto close the related issue when its merged in

PR #10, opened 2024-02-08
Add double censoring discretization 

SamuelBrand1:
The issue this is addressing is discussed [here](https://github.com/CDCgov/Rt-without-renewal/issues/7).

#### Method
Method implementation is [here](https://github.com/CDCgov/Rt-without-renewal/blob/03cf3bbb2ba7683a379f994aa3a3009445bccca9/src/utilities.jl#L84C1-L95C4). The concept is that the (by assumption) the primary time $P$ which gets censored to some observation time $t$ occurs in the interval $[t, t + \Delta d)$ (default is $\Delta d = 1$). 

The secondary time less the primary censored observation time $t$ is $S - t = X + U$ where $U \sim \mathcal{U}(0, \Delta d)$ and $X$ is from some continuous non-negative distribution with CDF $F_X$.

Therefore, the probability that the censored secondary is assigned to observation time $s = t + \tau$ where $\tau = 0, \Delta d, 2 \Delta d, \dots$ is:

```math
P(s) = F_{X + U}(\tau+ \Delta d) -  F_{X+U}(\tau). 
```

This is solved using numerical [quadature](https://juliamath.github.io/QuadGK.jl/stable/) on the integral

```math
 F_{X + U}(s) = \int_0^{\Delta d} {F_{X}(s - u) \over \Delta d} du.
```

A maximum value for $\tau$ is implemented using standard conditioning.

Users can also access a "basic" discretisation 

```math
P(\tau) = F_{X}(\tau + \Delta d) -  F_{X}(\tau). 
```

by dispatching on `Val(:basic)`, e.g.

```julia
Δd = 1
ts = (0.0:Δd:(D-Δd)) |> collect
pmf1 = create_discrete_pmf(cont_dist, Val(:basic); Δd = Δd, D = D) # Basic discretisation 
pmf2 = create_discrete_pmf(cont_dist; Δd = Δd, D = D) # Double censoring
```
The intention is to give a comparison.

#### Testing and visual evaluation

Unit testing is [here](https://github.com/CDCgov/Rt-without-renewal/blob/03cf3bbb2ba7683a379f994aa3a3009445bccca9/test/test_utilities.jl#L30).

Visualisation is [here](https://github.com/CDCgov/Rt-without-renewal/blob/SamuelBrand1/issue7/test/prior_predictive_checking/discretized_pmfs.jl). In particular, this shows fairly big differences when censoring is daily compared to hourly.

_EDIT: added $s = t + \tau$ bit_
_EDIT: rewritten a bit for correctness and clarity_


SamuelBrand1:
OK I'm assuming that if there was no censoring at all then we observe

$P$ and $S = P + X$ where $X$ is the delay from some distribution.

My understanding of the censoring model is that it `floor` based, e.g if $P \in [t, t +\Delta d)$ we observe $t$ as our observed censored primary time.

If we assume that, conditional on observing $t$, then $S = t + X + U$ where $U$ is uniform. This is the assumption that the primary is uniformly distributed within its censor interval. 

Then the probability that $S$ is reported in the same interval is 

```math
P(X + U \in [0, \Delta d)) = P(X + U < \Delta d) - P(X + U \leq 0)
``` 

the next interval is 

```math
P(X + U \in [\Delta d, 2 \Delta d)) = P(X + U < 2 \Delta d) - P(X + U \leq \Delta d).
``` 

and so on.

Am I missing something about censoring here?

zsusswein:
This may be an @seabbs question, but I think something is not-quite-right here based on a prior SamA conversation:

> The concept is that the (by assumption) the primary time $P$ which gets censored to some observation time $t$ occurs in the interval $[t, t + \Delta d)$ (default is $\Delta d = 1$). 

My understanding was that the approximation looked at $p(t - \Delta d \le P < t + \Delta d)$. For example, the density binned into g[0] would be from [-1, 1). Then g[1] would correspond to [0, 2) and there's some probability of a primary event at $t = 0.75$ being binned into g[0] and some probability it gets binned $g[1]$. Is that understanding incorrect?

SamuelBrand1:
Right! 

OK, thats easy to do, so long as I've got the precise mathematical description of what we want.

zsusswein:
Ok chatted with SamA and glanced at the paper (page 18, eqs. 12-14). We landed at (1) the notation here isn't quite right (2) the code might be right anyway (3) some cleanup and testing against the epinowcast monte carlo implementation would be good

### (1) Notation

Following [page 18 at equations 12-14](https://www.medrxiv.org/content/10.1101/2024.01.12.24301247v1.full.pdf), let $\pi \in \mathbb{R}^+$ be the time of the primary event and $\sigma \in  \mathbb{R}^+$ be the time of the secondary event. Define $f(x | \pi)$ as the conditional density of a delay of length $x$ for a primary event $\pi$ (the forward process). Define $g(\pi)$ to be the probability density function of primary event times over the domain of all $\pi$.

> [!NOTE]
> The manuscript uses the notation $g_P(x)$ and defines it as "the probability distribution of the primary event time". I find this notation pretty confusing because in the rest of the paper $a_I ( \cdot)$ is used to mean $a ( \cdot  | I = i)$. But here, I don't think it makes sense to talk about conditioning the distribution of the primary event on the primary event so I'm dropping the subscript/conditioning.

 Let $\pi_L$ be the lower bound and $\pi_R$ be the upper bound of the censoring on $\pi$ due to the observation process. Likewise, let $\sigma_L$ and $\sigma_R$ respectively be the bounds on censoring due to observation on the secondary event. Note that it is not necessarily true that $P(\sigma_L \le \pi) = 0$ nor $P(\sigma \le \pi) = 0$. This is the example of the negative serial interval.

For notational clarity, define
```math
 P(\pi^\star) = P( \pi_L < \pi < \pi_R |  \pi_L,  \pi_R)$ = \int_{\pi_L}^{\pi_R} g(a) da
```

 And define $P(\sigma^\star) =  P(\sigma_L < \sigma < \sigma_R |  \sigma_R,  \sigma_L)$. These probabilities are each the probability that the respective event of interest is in the interval defined by the specified lower and upper bounds.

To determine the probability distribution of secondary events for a given primary event with single censoring, we integrate over the forward distribution of delayed primary events, from $S_L$ to $S_R$:
```math
P(\sigma^\star | \Pi = \pi) \propto \int_{\sigma_L}^{\sigma_R} f_\pi (y - \pi) dy
```

When the primary event is also censored, we have double censoring. We can see that the single censoring approach can be modified to account for double censoring by integrating over the PDF of primary events within the censoring interval, $\pi_L$ and $\pi_R$

```math
P(\sigma^\star | \pi^\star) \propto \int_{\pi_L}^{\pi_R} \int_{\sigma_L}^{\sigma_R} g(x) f_\pi (y - x) dy dx
```

We can find the appropriate normalizing constant by Bayes rule:

$$ P(\sigma^\star | \pi^\star) P(\pi^\star) = P(\pi^\star | \sigma^\star) P(\sigma^\star)$$

$$ P(\sigma^\star | \pi^\star) = \frac{P(\pi^\star | \sigma^\star) P(\sigma^\star)}{P(\pi^\star)}$$

Plugging in the relationships found above for $P(\pi^\star)$

$$ P(\sigma^\star | \pi^\star) = \frac{P(\pi^\star | \sigma^\star) P(\sigma^\star)}{\int_{\pi_L}^{\pi_R} g(a) da}$$

And for $P(\sigma^\star | \pi^\star)$

```math
P(\sigma^\star | \pi^\star)  = \frac{ \int_{\pi_L}^{\pi_R}\int_{\sigma_L}^{\sigma_R} g(x) f_\pi (y - x) dy dx}{\int_{\pi_L}^{\pi_R} g(a) da}
```

Because we're observing a forward process, we by convention fix $\pi_L := 0$ and $\pi_R := 1$. 

```math
P(\sigma^\star | \pi^\star)  = \frac{ \int_{0}^{1}\int_{\sigma_L}^{\sigma_R} g(x) f_\pi (y - x) dy dx}{\int_{0}^{1} g(a) da}
```


92 is creating a single censored CDF, 94 is using the single censored CDF to censor again so you can use 94 as a distribution that is single censored to the primary event

might be equivalent because 92-94 is doing the int then differencing which is equivalent to doing [0, 1), [1, 2) and assigning to the upper bound. but it works because you're dropping the first element which is equivalent to assigning to the lower bound.

Either way, this is confusing enough code that it would be nice to clean it up and make the individual

SamuelBrand1:
Hey, I'll go line by line when I get a chance... but to cut through a lot of maths here.

Eq 14 in SWP's notation is:

```math
\int_{P_L}^{P_R} \int_{S_L}^{S_R} g_P(x|P_L, P_R) f_x(y-x)dy dx
```

1. swap integration (this is fine because finite quants [Fubini thm](https://en.wikipedia.org/wiki/Fubini%27s_theorem).
2. Note that $\int_{P_L}^{P_R} g_P(x|P_L, P_R) f_x(y-x) dx$ is the density at y of the random variable $X + Y$, $f_{X+Y}(y)$, where $X$ is the delay and $Y$ is the primary conditioned on being betwee $P_L$, $P_R$, due to [independence](https://en.wikipedia.org/wiki/Convolution_of_probability_distributions).
3. The approximation that $Y$ is actually uniform then recovers my original argument.

This maths discussion has gone a bit full [LOTUS](https://en.wikipedia.org/wiki/Law_of_the_unconscious_statistician)!

SamuelBrand1:
> We assume the density is uniform. In reality, this density varies with the epidemic growth rate (epidemic phase bias) but we see in practice that this contamination is small (Figure 2)

Yeah, this is an approx method.

SamuelBrand1:
Actually the code does have a subtle bug if $D$ is not a multiple of $\Delta d$, because it then assumes that final interval is $\Delta d$ width... I'll fix that tommorow.

zsusswein:
Yeah the context here is that I needed to spend some time working through this anyway, so I seized my opportunity.

seabbs:
That is a really nice walkthrough @zsusswein (which I personally found quite a lot clearer than @SamuelBrand1 💯 ).  Now that we have that I feel more clear that what @SamuelBrand1 is describing is the same thing and agree that the actual code implementations matches my expectations. 

I also think we should check via simulation in a test though just for extra safety.

I think the main notational confusion for me in all this was the use of the same interval for both the primary and secondary events. If that had been donated I think it would have been much clearer what was going on.

SamuelBrand1:
I've made a purely stylistic change. I don't promote using too many unicode characters, but this is an internal function and IMO maybe makes it more easy to connect directly to maths.

https://github.com/CDCgov/Rt-without-renewal/blob/5cbcab2eefa17361a67e45a07c876b430128bee5/EpiAware/src/utilities.jl#L91-L95

SamuelBrand1:
> That is a really nice walkthrough @zsusswein (which I personally found quite a lot clearer than @SamuelBrand1 💯 ). Now that we have that I feel more clear that what @SamuelBrand1 is describing is the same thing and agree that the actual code implementations matches my expectations.
> 
> I also think we should check via simulation in a test though just for extra safety.
> 
> I think the main notational confusion for me in all this was the use of the same interval for both the primary and secondary events. If that had been donated I think it would have been much clearer what was going on.

I'm about to put a unit test in against actual solvable distributions.

EpiAware/src/utilities.jl
```diff
@@ -28,12 +28,14 @@ function scan(f, init, xs)
...
-    create_discrete_pmf(dist; Δd = 1.0, D)
+    create_discrete_pmf(dist::Distribution, ::Val{:basic}; Δd = 1.0, D)
 
-Create a discrete probability mass function (PMF) from a given distribution.
+Create a discrete probability mass function (PMF) from a given distribution, assuming that the
+primary event happens at edge of an censoring inteval.
 
 Arguments:
 - `dist`: The distribution from which to create the PMF.
+- ::Val{:basic}: A dummy argument to dispatch to this method.
```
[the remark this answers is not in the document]
SamuelBrand1: It works, there are pitfalls at run-time if you try and leverage this kind of dispatch uncarefully e.g. [here](https://docs.julialang.org/en/v1/manual/performance-tips/#man-performance-value-type).

In this case, there is basically a good way of doing this, and a bad way we are keeping for historical/comparison purposes. And the code is nearly the same (you can't dispatch a solution method on type). So my thought was that dispatch on Val type is like a barrier. You only use this method if you _really_ want to look at single interval censoring rather than by mistake.

SamuelBrand1:
Unit test:

Easiest distribution I could solve analytically but was also not completely trivial was $X \sim \text{Exp}(1)$ day delay with daily interval censoring. W.l.o.g. Primary censored obs time is $t = 0$

In case I made a mistake at the top, I'll go straight from SWO (no Right truncation).

For above example the secondary censored obs time can be $s = 0, 1, 2,...$ days. The probability mass function is:

```math
P_S(s) = \int_0^1 \int_s^{s+1} f_X(y-x)dy dx.
```

This splits into two cases: $s = 0$ and $s \geq 1$.

### $s=0$

```math
P_S(0) = \int_0^1 \int_x^1 \exp(-(y - x)) dy dx = \exp(-1).
```

_NB: the density is zero for negative values._
### $s>1$
```math
P_S(s) = \int_0^1 \int_s^{s+1} \exp(-(y - x)) dy dx = (1 - \exp(-1)) (\exp(1) - 1) \exp(-s).
```

### Checking

Just to check that the above is a discrete prob distribution.

```math
\begin{align}
 \sum_{s \geq 1} P_S(s)&= (1 - \exp(-1)) (\exp(1) - 1) \sum_{s \geq 1}  \exp(-s)  \\
&=  (1 - \exp(-1)) (\exp(1) - 1)  {\exp(-1) \over 1 - \exp(-1)} \\
& = 1 -  \exp(-1).
\end{align}
```
Therefore,

```math
 \sum_{s \geq 0} P_S(s) = 1.
```

SamuelBrand1:
New unit test against exact solution at `1e-15` absolute tolerance (NB: its faster than MC sampling too):

https://github.com/CDCgov/Rt-without-renewal/blob/298eb33ed75f5e4346ceb1e8a67fd3bd958b5596/EpiAware/test/test_utilities.jl#L57-L67

SamuelBrand1:
Killed off having arbitrary `D` value.

EpiAware/src/utilities.jl
```diff
@@ -28,12 +28,14 @@ function scan(f, init, xs)
 end
 
 """
-    create_discrete_pmf(dist; Δd = 1.0, D)
+    create_discrete_pmf(dist::Distribution, ::Val{:basic}; Δd = 1.0, D)
 
-Create a discrete probability mass function (PMF) from a given distribution.
+Create a discrete probability mass function (PMF) from a given distribution, assuming that the
```
zsusswein, line 33: I may be misunderstanding, but isn't this implementation the simple single censoring case? With the assumption that we assign to the floor of the bin rather than the center?

Either way, would you mind adding some more clarification to the docstring about what this corresponds to re: single vs double censoring and "traditional" implementations?

EpiAware/test/test_utilities.jl
```diff
@@ -45,16 +45,33 @@ end
...
         @test pmf ≈ expected_pmf atol = 1e-15
     end
 
-end
+    # Test case 5: Testing output against expected PMF basic version - double
+    # interval censoring
+    @testset "Test case 5" begin
+        dist = Exponential(1.0)
+        expected_pmf_uncond = [exp(-1);
+                        [(1 - exp(-1)) * (exp(1) - 1) * exp(-s) for s = 1:9]
```
zsusswein, line 62: Can you throw a link to where you derive this? Or put it somewhere findable so that its origin doesn't get lost? (I like this test and am planning on re-using it)

zsusswein (approved):
This is really nice Sam. Left two small documentation things inline, but I don't need to look again after those are addressed,

EpiAware/test/test_utilities.jl
seabbs: yes I agree - ideally we would have literate docs containing the above as I think it might be hard for people to follow

SamuelBrand1: Its a pretty straightforward calc; its with probability 1 been written down somewhere.

Without a link, I could add some latex to the docstring?

EpiAware/src/utilities.jl
SamuelBrand1: Yes, thats the one. As mentioned above, the main reason I dispatch with a Val type is that I don't want an easy option here; you have to select this version quite hard.

EpiAware/test/test_utilities.jl
zsusswein: I think using [this link](https://github.com/CDCgov/Rt-without-renewal/pull/10#issuecomment-1935759767) is plenty

EpiAware/src/utilities.jl
zsusswein: Coolio -- if you could say that in the function documentation that would be helpful

EpiAware/test/test_utilities.jl
seabbs: I don't think we want to link out to a PR discussion tbh as if we version it it can be improved in the future. Can it just be included as a markdown help file and linked to that way?

SamuelBrand1: I've added the analytical derivation to the predictive checking for this method and checked that it renders to nice markdown with `Literate`, which means it'll eventually be part of the docs without further modification [here](https://github.com/CDCgov/Rt-without-renewal/blob/147e97d2e883ad86b9ec4128b47ae84da9335913/EpiAware/test/predictive_checking/discretized_pmfs.jl#L11).

EpiAware/src/utilities.jl
SamuelBrand1: I've added a bit more description and slightly extended so its default mid-point approximation for the primary event (but flexible to be any point).

Issue #12, opened 2024-02-08
Change repo directory structure


SamuelBrand1:
The current repo is structured in a default Julia package set up, but is intended to be a multi-use repo with a pipeline and other analysis as well as the `EpiAware` package code.

We want to change the repo so that each main branch of the project has its own sub-directory from the root repo directory.

PR #13, opened 2024-02-08
Proposed new directory structure

SamuelBrand1:
Addresses issue #12 

Pretty straightforward, except maybe we want different names for the main directories?

I've gone with:
- `/analysis`
- `/pipeline`
- `EpiAware`

NB: If you're running Julia with VS-Code defaults it will no longer automatically start in the project environment, but thats not hard to activate.


seabbs (approved):
LGTM. Agree project not autostarting is a bit of a pain but otherwise this seems worth it?

SamuelBrand1:
Actually, I think its probably all good. The VS-Code integration is so nice that you forget how to do CL Julia.

seabbs:
sounds like in my case it will be never learn

PR #17, opened 2024-02-09
Adds install-julia local action

seabbs:
This PR adds a install-julia action that uses caching to speed up julia installation and julia package installation. It also tweaks precommit configuration settings to give fewer false positives with julia.

I tested that the precommit action now works as expected (including with caching) locally using `act`.

In order to get pecommit to pass I ran it across all files (`pre-commit run --all-files`)


zsusswein (approved):
Thanks for figuring this out. 

If possible, I'd like to keep the python and R pre-commit hooks. I imagine we'll have some plotting or glue code pretty soon?

seabbs:
> If possible, I'd like to keep the python and R pre-commit hooks. I imagine we'll have some plotting or glue code pretty soon?

I killed these for now as the R ones especially were failing due to the lack of Rscript and I didn't want to add a install r step that would just slow down CI and not be used.

I agree we should add back in as soon as there is any R or python code and am happy to cover this issue when needed.

zsusswein:
Fair enough. Thanks for thinking it through.

PR #27, opened 2024-02-12
Add TestItemRunner.jl to enable VScode testing integration and support running individual unit tests

seabbs:
Adds `TestItemRunner.jl` and checks it works with both vscode testing framework and via `test`. Also cleans up some duplicate tests and improves test names.

This change was made as `TestItemRunner` works with the vscode testing framework and allows users in other IDEs to run specific tests using a Julia macro.


zsusswein (commented):
Two asks: 
1. Can you fix the broken CI? From a quick glance, it looks like `TestItems` is added successfully but `TestItemRunner` needs to be added as well 
2. Can you document in a commit message or something why you made the switch?

PR #31, opened 2024-02-12
Add Aqua.jl - check package "quality" indicators

seabbs:
`Aqua.jl` checks a range of package "quality" measures and is integrated within the `TestItems` framework. I have added it here.

Note that this PR is a work in progress as `Aqua.test_all()` flags a range of package issues though some of these may be false positives. 

I will resolve this but if any stick out as being obviously wrong please do flag ASAP.


SamuelBrand1:
Looking at the CI error `Aqua.jl` might not be compatible with the function generation that is defined by the macro `@model` from Turing. But not sure.

SamuelBrand1:
Looks like its working now! LGTM.

SamuelBrand1 (approved):

Issue #15, opened 2024-02-09
Make one out of log-infections, $r_t$ or $R_t$ models functional.


SamuelBrand1:

SamuelBrand1:
`EpiModel` struct now ingests continuous distributions for generation interval and delay.

- Generation interval gets left truncated so zero delay infections (at the time discretisation size) don't apply and distribution is renormed.
- Delay distribution has double censoring with the uniform primary event within censoring interval approximation.

Issue #14, opened 2024-02-09
Toy model and output


SamuelBrand1:
Following a discussion with @seabbs I think we should move more rapidly towards an end-to-end example with a standard output. The reason for this, as opposed to engineering each latent process we want and then moving on is that the sooner we can run one full example, the sooner we can reason on the pipeline and manuscript format (e.g. plotting).

##### Sub-tasks
- [x] #15
- [x] #43
- [x] #55

PR #33, opened 2024-02-12
Toy model generate data

SamuelBrand1:
As per Issue #14  and issue #15 it is necessary to have at least one fully defined model.

In this case, each model has two main parts:

1. A `Turing` model defining how an underlying infection time series is generated from a latent process, and how the underling infection time series defines the log-likelihood of any observed case data, e.g. after a delay.
2. Another `Turing` model defining the latent process, which interacts with the main model via `@submodel`.

The PR covers defining 1. directly generating infections by `exp` transforming the latent process, and 2. that latent process is a random walk with random step size variance and random initial condition. This also acts to test the underlying logic of using submodels to create modularity.

Unit tests are [here](https://github.com/CDCgov/Rt-without-renewal/blob/toy-model-generate-data/EpiAware/test/test_latent-processes.jl) and [here](https://github.com/CDCgov/Rt-without-renewal/blob/toy-model-generate-data/EpiAware/test/test_models.jl).

An example script that runs in `TestDev` mode is [here](https://github.com/CDCgov/Rt-without-renewal/blob/toy-model-generate-data/EpiAware/test/predictive_checking/toy_model_log_infs_RW.jl).


EpiAware/src/EpiAware.jl
```diff
@@ -30,13 +30,11 @@ using Distributions,
...
 
+# Exported utilities
 export scan,
-    create_discrete_pmf,
-    growth_rate_to_reproductive_ratio,
-    generate_observation_kernel,
-    EpiModel,
-    log_daily_infections,
-    random_walk
+    create_discrete_pmf, growth_rate_to_reproductive_ratio, generate_observation_kernel
```
seabbs, line 35: why is scan on its own line but none of the others are? That aside splitting it out like this is nice

EpiAware/src/epimodel.jl
```diff
@@ -1,24 +1,24 @@
 abstract type AbstractEpiModel end
 
 
+
 """
-    struct EpiModel{T}
+    struct EpiModel{T<:Real} <: AbstractEpiModel
 
-The `EpiModel` struct represents a basic renewal model.
+EpiModel represents an epidemiological model with generation intervals, delay intervals, and observation delay kernel.
```
seabbs, line 8: Can we at some point give this a cooler name?

EpiAware/src/epimodel.jl
```diff
@@ -1,24 +1,24 @@
...
-- `delay_int::Vector{T}`: Discrete delay distribution.
-- `delay_kernel::SparseMatrixCSC{T,Integer}`: Sparse matrix representing the action of convoluting
-    a series of infections with observation delay.
-- `cluster_coeff::T`: Cluster coefficient for negative binomial distribution of observations.
-- `len_gen_int::Integer`: Length of `gen_int` vector.
-- `len_delay_int::Integer`: Length of `delay_int` vector.
+- `gen_int::Vector{T}`: Discrete generation inteval, runs from 1, 2, ... to the end of the vector.
+- `delay_int::Vector{T}`: Discrete delay distribution runs from 0, 1, ... to the end of the vector less 1.
+- `delay_kernel::SparseMatrixCSC{T,Integer}`: Sparse matrix representing the observation delay kernel.
+- `cluster_coeff::T`: Cluster coefficient for negative binomial observations.
```
seabbs, line 14: can we add some help here for people so they can understand its 1 / overdispersion

EpiAware/src/epimodel.jl
```diff
@@ -28,14 +28,15 @@ struct EpiModel{T<:Real} <: AbstractEpiModel
...
+    function EpiModel(gen_int, delay_int, cluster_coeff, time_horizon::Integer)
         @assert all(gen_int .>= 0) "Generation interval must be non-negative"
         @assert all(delay_int .>= 0) "Delay interval must be non-negative"
         @assert sum(gen_int) ≈ 1 "Generation interval must sum to 1"
         @assert sum(delay_int) ≈ 1 "Delay interval must sum to 1"
-        #construct observation delay kernel
+
         K = generate_observation_kernel(delay_int, time_horizon)
 
         new{eltype(gen_int)}(
```
seabbs, line 42: can you explain to me like I am 15 what this bit does?

EpiAware/src/latent-processes.jl
```diff
@@ -1,3 +1,7 @@
+const STANDARD_RW_PRIORS =
```
seabbs, line 1: what is the logic for defining this like this? On the face of it it seems quite ugly?

seabbs: and if doing so shouldn't it have its own documentation? In R I would guess this would be a little function

EpiAware/src/models.jl
```diff
@@ -1,10 +1,39 @@
...
+
+## Arguments
+- `y_t`: Observed data.
+- `epimodel`: Epidemiological model.
+- `latent_process`: Latent process model.
+- `latent_process_priors`: Priors for the latent process model.
+- `transform_function`: Function to transform the latent process into infections. Default is `exp`.
+- `n_generate_ahead`: Number of time steps to generate ahead. Default is `0`.
+- `pos_shift`: Positive shift to avoid zero values. Default is `1e-6`.
+- `neg_bin_cluster_factor`: Missing value for the negative binomial cluster factor. Default is `missing`.
```
seabbs, line 25: In R we would manage the docs over lap with EpiModel by inheriting documentation is there some kind of mechanism for that in Julia and if so can we make an issue to use it.

EpiAware/src/models.jl
```diff
@@ -1,10 +1,39 @@
+"""
+    log_infections(y_t, epimodel::EpiModel, latent_process;
```
seabbs, line 2: In Julia do you really need to manually write out the arguments of the function in the docs that are defined below? It seems like a bit of duplication that could lead to errors?

EpiAware/src/models.jl
```diff
@@ -13,16 +42,17 @@
...
+    time_steps = epimodel.time_horizon
     @submodel _I_t, latent_process_parameters =
-        latent_process(data_length; latent_process_priors = latent_process_priors)
+        latent_process(time_steps; latent_process_priors = latent_process_priors)
 
     #Transform into infections
     I_t = transform_function.(_I_t)
 
     #Predictive distribution
     case_pred_dists =
```
seabbs, line 53: ultimately can we also express this as a submodel?

EpiAware/src/models.jl
```diff
@@ -1,10 +1,39 @@
...
+- `transform_function`: Function to transform the latent process into infections. Default is `exp`.
+- `n_generate_ahead`: Number of time steps to generate ahead. Default is `0`.
+- `pos_shift`: Positive shift to avoid zero values. Default is `1e-6`.
+- `neg_bin_cluster_factor`: Missing value for the negative binomial cluster factor. Default is `missing`.
+- `neg_bin_cluster_factor_prior`: Prior distribution for the negative binomial cluster factor. Default is `Gamma(3, 0.05 / 3)`.
+
+## Returns
+A named tuple containing the generated quantities `I_t` and `latent_process_parameters`.
+"""
 @model function log_infections(
```
seabbs, line 31: I was sort of expecting each model to inherit more from the EpiModel (i.e to be an instantiation of it) as it stands this seems like you will need to duplicate quite a lot of code?

EpiAware/test/predictive_checking/toy_model_log_infs_RW.jl
```diff
@@ -0,0 +1,123 @@
...
+r &\sim \text{Gamma}(3, 0.05/3).
+\end{align}
+```
+
+## Load dependencies in `TestEnv`
+
+=#
+
+split(pwd(), "/")[end] != "EpiAware" && begin
+    cd("./EpiAware")
```
seabbs, line 55: I don't love to see cd in code. Can we just assume that the user is in the root of the project directory for this to work or leave them to activate the right environment themselves?

EpiAware/test/predictive_checking/toy_model_log_infs_RW.jl
```diff
@@ -0,0 +1,123 @@
...
+#=
+## Define a log-infections model
+The log-infections model is defined by a Turing model `log_infections`.
+
+In this case we don't have observed data, so we use `missing` value for `y_t`.
+=#
+toy_log_infs = log_infections(
+    missing,
+    epimodel,
+    random_walk;
```
seabbs, line 100: I am a bit confused about how epimodel interacts with log_infections. Specifically why they both share arguments and how that gets internally determined.

seabbs (commented):
I've just done a sight read and this generally looks good. Overall there is more model structure in log_infections than I was expecting (I thought it would just be a version of EpiModel with a defined map from the latent process to infections but it looks like it has more custom code than that. 

Will check out and run in a bit and circle back

EpiAware/src/epimodel.jl
SamuelBrand1: Yes!

SamuelBrand1:
> I've just done a sight read and this generally looks good. Overall there is more model structure in log_infections than I was expecting (I thought it would just be a version of EpiModel with a defined map from the latent process to infections but it looks like it has more custom code than that.
> 
> Will check out and run in a bit and circle back

This is actually not a bad suggestion.

If the `EpiModel` struct also defines the latent process -> underlying infections transformation then we're done.

EpiAware/src/EpiAware.jl
SamuelBrand1: The JuliaFormatter basic format standard is to let tuple arguments be flat until they hit a certain length and then change into column format.

SamuelBrand1: We can fiddle that if its annoying.

EpiAware/src/models.jl
SamuelBrand1: No, you can use https://github.com/JuliaDocs/DocStringExtensions.jl .

EpiAware/src/models.jl
SamuelBrand1: If you want to shift y_t then yes. Or we can have a function which outputs the predictive distribution if you prefer.

EpiAware/src/epimodel.jl
SamuelBrand1: Structs come with a default constructor. You can also define inner and outer constructors. Inner constuctors come with a reserved keyword `new` and the chance to give a type for parametric typed structs.

The docs are [here](https://docs.julialang.org/en/v1/manual/constructors/)

EpiAware/src/epimodel.jl
SamuelBrand1: Lets do this in a follow on refactor

EpiAware/src/models.jl
SamuelBrand1: Lets use doc string extension as mentioned elsewhere.

seabbs:
In a f2f conversation we decided to proceed with this as is and to then open additional issues to improve documentation (#35) and to make the model implementation better show case Julia features and to be more DRY.

seabbs (commented):
LGTM - still waiting on an issue for the revised approach to the model framework

seabbs (approved):
LGTM - still waiting on an issue for the revised approach to the model framework

SamuelBrand1:
#38

Issue #32, opened 2024-02-12
git ignore manifest.toml


seabbs:
When EpiAware is stable community posts suggest that good practice is to depend on the `Package.toml` and to therefore `.gitignore` the manifest. This is also helpful as it avoids frequent merge issues during development due to differing package versions etc.

SamuelBrand1:
Agreed.

seabbs:
Due you have a read on when we should do this? I almost think it should be sooner rather than later?

PR #36, opened 2024-02-13
Issue 32: First remove Manifest.toml and gitignore

seabbs:
As discussed in #32 this PR adds all `Manifest.toml` files to the `.gitignore` for the `EpiAware` submodule. This is motivated by continuing issues with merge issues due to small manifest changes and because as we are aiming to produce a reusable package toolkit it makes sense that the `EpiAware.jl` tooling works robustly without needing exact versions.


SamuelBrand1 (approved):
LGTM, and runs.

Issue #35, opened 2024-02-13
Improve docstrings


seabbs:
Can we improve our docstring practice to reduce duplication, improve clarity and generally set a standard?

I see this suggestion for python which seems to make sense to me: https://numpydoc.readthedocs.io/en/latest/example.html#example

In particular I like the lack of repetition of the function arguments, the use of doctests, and the use of a clear summary section and notes section for further details

Are there resources from the Julia community we could point at to justify/improve our practice?

SamuelBrand1:
https://github.com/JuliaDocs/DocStringExtensions.jl

SamuelBrand1:
Since we are using the SciML style for formatting; I assume we are going to use the SciML preferred documentation [conventions](https://github.com/SciML/SciMLStyle?tab=readme-ov-file#documentation)?

seabbs:
what do you think? I think yes and its good to follow established standards and it seems like a high quality ecosystem

SamuelBrand1:
> This makes me wonder if we do need this template builder for the doc strings or not?

Which template builder? `DocStringExtensions`? 

I think its handy.

SamuelBrand1:
@seabbs @zsusswein 
I've started putting this in on this [branch](https://github.com/CDCgov/Rt-without-renewal/tree/81-document); although far from finished version.

But have a look to see if that the level of doc you want.

seabbs:
> Which template builder? DocStringExtensions?

Yes and to be clear i would like to  use it but not going to force the it on anyone

seabbs:
@SamuelBrand1 could you make make some sub issues for parts that still need docs so someone else can pick this up?

SamuelBrand1:
> @SamuelBrand1 could you make make some sub issues for parts that still need docs so someone else can pick this up?

Yes, and not sure why I didn't just do that!

SamuelBrand1:
Basically, I'm happy on that branch _until_ the bit documenting the callable of `Renewal`... thats clunky due to LaTeX strings (sure there is a better way).

Fairly random thoughts:
1. `DocStringExtensions` is probably most useful deployed asap, because it reduces the work load of redocumenting breaking changes.
2. I haven't used [`DocStringExtensions.@template`](https://docstringextensions.juliadocs.org/stable/#DocStringExtensions.@template) but it looks handy? Again the kind of thing to set up before extensive documentation to save time.

seabbs:
[...]

PR #39, opened 2024-02-14
Implement a proposed modular API to specify a data (i.e. case) generating process

### LINKED ISSUE #15: Make one out of log-infections, $r_t$ or $R_t$ models functional.

### LINKED ISSUE #38: Modular interface between fixed data of inference problem and the `Turing` model implementing the inference problem
## Interface

Atm, the toy model merged in #33 ingests an `EpiModel` struct which contains data that we are treating as fixed for the purposes of this project into a `Turing` model called `log_infections`.

This fixes how any latent process generates underlying infections at the level of separate `Turing models` and therefore implies a substantial amount of repeated code.

## Change to interface

The proposal is for the generative process to dispatch on _either_ the type of the fixed data struct _or_ from a function which is a field of the fixed data struct.

### Options

1. Define concrete subtypes, e.g. `DirectInfections <: AbstractEpiModel` which dispatch a particular method for latent process -> infections.
2. Add another field to `EpiModel` which is the function that does latent process -> infections.

Personal preference is 1. because I think that the pattern

```julia
... data wrangling...

epimdl = DirectInfections(my_distributions...)
my_inference_results = do_inference(data, epimdl...)
```
where `do_inference` is an abstraction on top of `Turing` sampling from the model is easier to understand than

```julia
... data wrangling...

epimdl = EpiModel(my_distributions..., direct_infections_func)
my_inference_results = do_inference(data, epimdl...)
```

But I don't have a very strong preference here.

SamuelBrand1:
This PR introduces the modular API to the data generating process discussed in #38 .

### Major additions
Diagrammatically, the "modules" (in workflow rather than Julia sense) that are implemented are below. Dotted lines fill in the currently unimplemented steps.

Unit tests are in test scripts [here](https://github.com/CDCgov/Rt-without-renewal/tree/38-modular-inference/EpiAware/test).

A script to run in `TestDev` mode which demonstrates the currently implemented steps is [here](https://github.com/CDCgov/Rt-without-renewal/blob/38-modular-inference/EpiAware/test/predictive_checking/toy_model_log_infs_RW.jl).

### Minor additions
- Made the RW distributional test slightly more [precise](https://github.com/CDCgov/Rt-without-renewal/blob/bd386d2a2b9309e7f84d9316ae60861d96f80bbd/EpiAware/test/test_latent-processes.jl#L21)
- Moved the random walk default priors from a `const` object to a simple [construction function](https://github.com/CDCgov/Rt-without-renewal/blob/bd386d2a2b9309e7f84d9316ae60861d96f80bbd/EpiAware/src/latent-processes.jl#L1C10-L1C27)

### Missing feature
I have not implemented docstrings in some functions here ahead of addressing #35 

```mermaid
flowchart TD

    A["Underlying dists.
---------------------
EpiData"]

    B["Choice of target 
for latent process
---------------------
DirectInfections
    ExpGrowthRate
    Renewal"]

C["Observational Data
---------------------
Obs. cases y_t"]
D["Latent processes
---------------------
Random Walk"] 
E[Turing model constructor]
F["Latent Process priors"]
G[Posterior draws]
H[Posterior checking]
I[Post-processing]
DataW[Data wrangling and QC]

    A --> EpiModel
    B --> EpiModel
    EpiModel -->E
    C-->E
    D-->|random_walk| E
F-->|default_rw_priors|E
E-.->|sample...NUTS...| G
G-.->H
H-.->I
DataW-.->C
```
Closes #38 
Closes #15


EpiAware/src/epimodel.jl
```diff
@@ -1,96 +1,83 @@
...
+
+struct ExpGrowthRate <: AbstractEpiModel
+    data::EpiData
+end
 
-# Arguments
-- `recent_incidence`: Array of recent incidence values.
-- `Rt`: Reproduction number.
+function (epi_model::ExpGrowthRate)(recent_incidence, rt)
+    new_incidence = recent_incidence * exp(rt)
```
seabbs, line 72: does it make any difference in julia to do this as exp(log_recent_incidence + rt) vs current suggestion?

seabbs: Also this implementation suggests it will be used interatively (in the scan function). Are we clear that this is more efficient than the usual approach to implementation (exp(log_init_incidence + cumsum(rt)))?

EpiAware/src/latent-processes.jl
```diff
@@ -1,32 +1,25 @@
...
-"""
-    random_walk(n, ϵ_t = missing; latent_process_priors = (var_RW_dist = truncated(Normal(0., 0.05), 0., Inf),), ::Type{T} = Float64) where {T <: Real}
-
-Constructs a random walk model.
-
-# Arguments
-- `n`: The number of time steps.
-- `ϵ_t`: The random noise vector. Defaults to `missing`, in which case it is sampled from the standard multivariate normal distribution.
-- `latent_process_priors`: The prior distribution for the latent process parameters. Defaults to `(var_RW_dist = truncated(Normal(0., 0.05), 0., Inf),)`.
+function default_rw_priors()
```
seabbs, line 1: So this was a suggest for what I would expect to see in R. Does it make sense in Julia or was the original more normal?

EpiAware/src/epimodel.jl
SamuelBrand1: No. That would be fine.

EpiAware/src/latent-processes.jl
```diff
@@ -1,32 +1,25 @@
...
+) where {T<:AbstractFloat}
     rw = Vector{T}(undef, n)
     ϵ_t ~ MvNormal(ones(n))
     σ²_RW ~ latent_process_priors.var_RW_dist
     init_rw_value ~ latent_process_priors.init_rw_value_dist
     σ_RW = sqrt(σ²_RW)
-    rw .= init_rw_value .+ cumsum(σ_RW * ϵ_t)
-    return rw, (; σ_RW, init_rw_value)
+
+    rw[1] = init_rw_value + σ_RW * ϵ_t[1]
```
seabbs, line 20: why the change here?

EpiAware/src/models.jl
```diff
@@ -1,39 +1,8 @@
...
-- `neg_bin_cluster_factor_prior`: Prior distribution for the negative binomial cluster factor. Default is `Gamma(3, 0.05 / 3)`.
-
-## Returns
-A named tuple containing the generated quantities `I_t` and `latent_process_parameters`.
-"""
-@model function log_infections(
+@model function make_epi_inference_model(
     y_t,
-    epimodel::EpiModel,
+    epimodel::AbstractEpiModel,
```
seabbs, line 3: Could you just briefly explain why we are handling the renewal etc differently to the underling latent process? I.e. why is one a function and one a AbstractEpiModel struct?

seabbs: I guess I am wondering if the priors for the latent process should be part of the latent process object in particular? Or if that is a pattern that doesn't make a lot of sense?

EpiAware/src/models.jl
```diff
@@ -42,16 +11,16 @@ A named tuple containing the generated quantities `I_t` and `latent_process_para
...
         latent_process(time_steps; latent_process_priors = latent_process_priors)
 
     #Transform into infections
-    I_t = transform_function.(_I_t)
+    I_t, _ = scan(epimodel, latent_process_parameters.init, latent_process)
 
     #Predictive distribution
     case_pred_dists =
-        (epimodel.delay_kernel * I_t) .+ pos_shift .|>
+        (epimodel.data.delay_kernel * I_t) .+ pos_shift .|>
```
seabbs, line 23: As currently coded isn't the neg_bin_cluster_factor part of EpiData?

EpiAware/src/models.jl
seabbs: And a similar comment for the observation model.

seabbs (commented):
This looks really nice. It feels a little still like there is a mixture of styles going on here for the different parts of the model (as in the issue) it would be great to have a little more clarity on that. 

I think we might be best places though to review this design as it stands and then do another round if issue/PR iteration? 

I've only sight read for now so will circle back once I have looked over the code locally.

EpiAware/src/epimodel.jl
SamuelBrand1: In fact, I just noted that I should have been treating the initial value generated by `random_walk` and the initialiser for the infection time series. 

So above was wrong. Just committed the fix.

seabbs: For both?

EpiAware/src/latent-processes.jl
SamuelBrand1: > Also this implementation suggests it will be used interatively (in the scan function). Are we clear that this is more efficient than the usual approach to implementation (exp(log_init_incidence + cumsum(rt)))?

The function is iteratively applied but was _also_ supposed to do the transformation back from the unconstrained domain (default apply `exp`). It wasn't doing that so I've committed a fix.

There won't be an efficiency gain for vectorising this call in Julia (it compiles down to same thing).

The main utility of `scan` here is that all the models come down to the same set of operations.

SamuelBrand1: If we were only doing `rt` I'd just use the pattern above.

SamuelBrand1: Another option is to apply the (default) `exp` transform inside `random_walk` and abstract away `scan` or not using `scan` inside the `EpiModel` callable definition.

SamuelBrand1: Or a mixture of those strategies

EpiAware/src/latent-processes.jl
SamuelBrand1: I was declaring `rw` into memory, _then_ declaring another RHS vector and assigning. I wasn't sure if the compiler was abstracting that away (got to start some proper profiling...), but this way it definitely isn't.

EpiAware/src/models.jl
SamuelBrand1: That should probably be `true_neg_bin_cluster_factor`; I was thinking about storing all the data required to generate data from a "true model".

EpiAware/src/models.jl
SamuelBrand1: > I guess I am wondering if the priors for the latent process should be part of the latent process object in particular? Or if that is a pattern that doesn't make a lot of sense?

Open to suggestions here!

So the mental model of `latent_process` is its a function that makes objects (in this case Turing models) rather than an object itself. So my first pass was to include some priors to give to it that says "Make Turing model with these priors". This interfaces nicely with `@submodel` to say "actually don't make that Turing model, but add it to the one you're declared inside of".  

Of course we could wrap these together so there is a self-contained object/struct that has "here is how we build latent processes" field and "these are the priors" field.

EpiAware/src/epimodel.jl
SamuelBrand1: > For both?

Hmmmm...... Actually we think abit about the `init` for renewal. As ever.

EpiAware/src/latent-processes.jl
SamuelBrand1: This way is fine/better. Goal should be zero `const` objects if possible.

seabbs (approved):
I think this is good to go once issues have been resolved (i.e. moved to issues) as is. I do think there is a bit more conversation to be had about the precise modular construction of the model but I think we can iterate on this over time and should move on to postprocessing and formulating the different models likely first.

EpiAware/src/models.jl
seabbs: I'm still a bit confused about this as I don't think it should be part of the `EpiData` as its an observation process but  I think we should discuss in another issue.

PR #42, opened 2024-02-16
Initialisation for epidemic inference

SamuelBrand1:
This PR is for creating an appropriate initialisation for inference. In practice, this applies to the renewal model approach encapsulated by the `Renewal` model.

#### Major changes

- An new expectation is that any latent process model should pass an `init` kw value and this can have its own priors e.g. here:
https://github.com/CDCgov/Rt-without-renewal/blob/dece834bbbdba185d4e10282c250d1a6ff8dc18c/EpiAware/src/latent-processes.jl#L17-L24
- The callable function associated with an object which is a subtype of `AbstractEpiModel` is now fixed as defining the latent process -> infections transformation. Before it defined an updating function to be used by `scan`; however that was not the natural approach in (say) direct infection modelling. Instances of `Renewal` model now use `scan` under the hood. eg here:
https://github.com/CDCgov/Rt-without-renewal/blob/dece834bbbdba185d4e10282c250d1a6ff8dc18c/EpiAware/src/models.jl#L15-L19
- This change means that initialisation is also dealt with by the callable function. The only complex one is for `Renewal` [here](https://github.com/CDCgov/Rt-without-renewal/blob/dece834bbbdba185d4e10282c250d1a6ff8dc18c/EpiAware/src/epimodel.jl#L79C1-L93C4).
- For `Renewal` initialisation is done by treating `init` as defining a time zero incidence. The pre-time zero incidence is backwards exponential with the exponential growth rate implied by `R_t[1]` from the latent process and the GI. Note that this matches @dylanhmorris [here](https://github.com/cdcent/cfa-stf-team-materials/discussions/24). The difference is I use a quick Newton method with an informed guess to do the $\mathcal{R}_0$ to $r$ transformation; this is faster than using a generic root finder because you can solve the structure of the problem [here](https://github.com/CDCgov/Rt-without-renewal/blob/999bd8e2bfdd108a63216fca0c7de908e07a6c70/EpiAware/src/utilities.jl#L133C1-L146C4) and gets to high accuracy in a couple of steps: evidence [here](https://github.com/CDCgov/Rt-without-renewal/blob/40-initialisation-infection-generation-processes/EpiAware/test/predictive_checking/fast_approx_for_r.jl).

#### Minor change

- I've changed an `Aqua` kwarg to `persistent_tasks = false` because locally a Turing extension was precompiling out of sequence (TBD to burrow into that).

### DIFF
--- a/EpiAware/src/EpiAware.jl
+++ b/EpiAware/src/EpiAware.jl
@@ -35,16 +35,19 @@ export scan,
     create_discrete_pmf,
     growth_rate_to_reproductive_ratio,
     generate_observation_kernel,
-    default_rw_priors
+    default_rw_priors,
+    neg_MGF,
+    dneg_MGF_dr,
+    fast_R_to_r_approx
 
 # Exported types
-export EpiData, Renewal, ExpGrowthRate, DirectInfections
+export EpiData, Renewal, ExpGrowthRate, DirectInfections, AbstractEpiModel
 
 # Exported Turing model constructors
 export make_epi_inference_model, random_walk
 
-include("utilities.jl")
 include("epimodel.jl")
+include("utilities.jl")
 include("models.jl")
 include("latent-processes.jl")
 
--- a/EpiAware/src/epimodel.jl
+++ b/EpiAware/src/epimodel.jl
@@ -60,24 +60,34 @@ struct DirectInfections <: AbstractEpiModel
     data::EpiData
 end
 
-function (epi_model::DirectInfections)(recent_incidence, unc_I_t)
-    nothing, epi_model.data.transformation(unc_I_t)
+function (epimodel::DirectInfections)(_It, latent_process_aux)
+    epimodel.data.transformation.(_It)
 end
 
 struct ExpGrowthRate <: AbstractEpiModel
     data::EpiData
 end
 
-function (epi_model::ExpGrowthRate)(unc_recent_incidence, rt)
-    new_unc_recent_incidence = unc_recent_incidence + rt
-    new_unc_recent_incidence, epi_model.data.transformation(new_unc_recent_incidence)
+function (epimodel::ExpGrowthRate)(rt, latent_process_aux)
+    latent_process_aux.init .+ cumsum(rt) .|> exp
 end
 
 struct Renewal <: AbstractEpiModel
     data::EpiData
 end
 
-function (epi_model::Renewal)(recent_incidence, Rt)
-    new_incidence = Rt * dot(recent_incidence, epi_model.data.gen_int)
-    [new_incidence; recent_incidence[1:(epi_model.data.len_gen_int-1)]], new_incidence
+function (epimodel::Renewal)(_Rt, latent_process_aux)
+    I₀ = epimodel.data.transformation(latent_process_aux.init)
+    Rt = epimodel.data.transformation.(_Rt)
+
+    r_approx = fast_R_to_r_approx(Rt[1], epimodel)
+    init = I₀ * [exp(-r_approx * t) for t = 0:(epimodel.data.len_gen_int-1)]
+
+    function generate_infs(recent_incidence, Rt)
+        new_incidence = Rt * dot(recent_incidence, epimodel.data.gen_int)
+        [new_incidence; recent_incidence[1:(epimodel.data.len_gen_int-1)]], new_incidence
+    end
+
+    I_t, _ = scan(generate_infs, init, Rt)
+    return I_t
 end
--- a/EpiAware/src/latent-processes.jl
+++ b/EpiAware/src/latent-processes.jl
@@ -14,12 +14,12 @@ end
     rw = Vector{T}(undef, n)
     ϵ_t ~ MvNormal(ones(n))
     σ²_RW ~ latent_process_priors.var_RW_dist
-    init_rw_value ~ latent_process_priors.init_rw_value_dist
+    init ~ latent_process_priors.init_rw_value_dist
     σ_RW = sqrt(σ²_RW)
 
-    rw[1] = init_rw_value + σ_RW * ϵ_t[1]
+    rw[1] = σ_RW * ϵ_t[1]
     for t = 2:n
         rw[t] = rw[t-1] + σ_RW * ϵ_t[t]
     end
-    return rw, (; σ_RW, init_rw_value, init = rw[1])
+    return rw, (; σ_RW, init)
 end
--- a/EpiAware/src/models.jl
+++ b/EpiAware/src/models.jl
@@ -12,11 +12,11 @@
 
     #Latent process
     time_steps = epimodel.data.time_horizon
-    @submodel latent_process, latent_process_parameters =
+    @submodel latent_process, latent_process_aux =
         latent_process(time_steps; latent_process_priors = latent_process_priors)
 
     #Transform into infections
-    I_t, _ = scan(epimodel, latent_process_parameters.init, latent_process)
+    I_t = epimodel(latent_process, latent_process_aux)
 
     #Predictive distribution
     case_pred_dists =
@@ -27,5 +27,5 @@
     y_t ~ arraydist(case_pred_dists)
 
     #Generate quantities
-    return (; I_t, latent_process_parameters)
+    return (; I_t, latent_process, latent_process_aux)
 end
--- a/EpiAware/src/utilities.jl
+++ b/EpiAware/src/utilities.jl
@@ -17,7 +17,7 @@ value. This is similar to the JAX function `jax.lax.scan`.
 - `ys`: An array containing the result values of applying `f` to each element of `xs`.
 - `carry`: The final accumulator value.
 """
-function scan(f, init, xs)
+function scan(f::Function, init, xs::Vector{T}) where {T<:Union{Integer,AbstractFloat}}
     carry = init
     ys = similar(xs)
     for (i, x) in enumerate(xs)
@@ -110,6 +110,46 @@ function create_discrete_pmf(dist::Distribution; Δd = 1.0, D)
 end
 
 """
+    neg_MGF(r, w::AbstractVector)
+
+Compute the negative moment generating function (MGF) for a given rate `r` and weights `w`.
+
+# Arguments
+- `r`: The rate parameter.
+- `w`: An abstract vector of weights.
+
+# Returns
+The value of the negative MGF.
+
+"""
+function neg_MGF(r, w::AbstractVector)
+    return sum([w[i] * exp(-r * i) for i = 1:length(w)])
+end
+
+function dneg_MGF_dr(r, w::AbstractVector)
+    return -sum([w[i] * i * exp(-r * i) for i = 1:length(w)])
+end
+
+function fast_R_to_r_approx(R₀, w::Vector{T}; newton_steps = 1) where {T<:AbstractFloat}
+    mean_gen_time = dot(w, 1:length(w))
+    # Small r approximation as initial guess
+    r_approx = (R₀ - 1) / (R₀ * mean_gen_time)
+    # Newton's method
+    for _ = 1:newton_steps
+        r_approx +=
+            (1 - R₀ * neg_MGF(r_approx, w)) * neg_MGF(r_approx, w) /
+            dneg_MGF_dr(r_approx, w)
+    end
+    return r_approx
+end
+
+function fast_R_to_r_approx(R₀, epimodel::AbstractEpiModel; newton_steps = 3)
+    fast_R_to_r_approx(R₀, epimodel.data.gen_int; newton_steps = newton_steps)
+end
+
+
+
+"""
     growth_rate_to_reproductive_ratio(r, w)
 
 Compute the reproductive ratio given exponential growth rate `r`
@@ -123,7 +163,7 @@ Compute the reproductive ratio given exponential growth rate `r`
 - The reproductive ratio.
 """
 function growth_rate_to_reproductive_ratio(r, w::AbstractVector)
-    return 1 / sum([w[i] * exp(-r * i) for i = 1:length(w)])
+    return 1 / neg_MGF(r, w::AbstractVector)
 end
 
 
--- a/EpiAware/src/utilities.jl
+++ b/EpiAware/src/utilities.jl
@@ -136,14 +136,12 @@ function fast_R_to_r_approx(R₀, w::Vector{T}; newton_steps = 1) where {T<:Abst
     r_approx = (R₀ - 1) / (R₀ * mean_gen_time)
     # Newton's method
     for _ = 1:newton_steps
-        r_approx +=
-            (1 - R₀ * neg_MGF(r_approx, w)) * neg_MGF(r_approx, w) /
-            dneg_MGF_dr(r_approx, w)
+        r_approx -= (R₀ * neg_MGF(r_approx, w) - 1) / (R₀ * dneg_MGF_dr(r_approx, w))
     end
     return r_approx
 end
 
-function fast_R_to_r_approx(R₀, epimodel::AbstractEpiModel; newton_steps = 3)
+function fast_R_to_r_approx(R₀, epimodel::AbstractEpiModel; newton_steps = 4)
     fast_R_to_r_approx(R₀, epimodel.data.gen_int; newton_steps = newton_steps)
 end
 
--- a/EpiAware/test/predictive_checking/fast_approx_for_r.jl
+++ b/EpiAware/test/predictive_checking/fast_approx_for_r.jl
@@ -1,0 +1,78 @@
+#=
+# Fast approximation for `r` from `R₀`
+I use the negative moment generating function (MGF).
+
+Let
+```math
+G(r) = \sum_{i=1}^{\infty} w_i e^{-r i}.
+```
+and
+```math
+f(r, \mathcal{R}_0) = {1 \over G(r)} - \mathcal{R}_0.
+```
+then the connection between `R₀` and `r` is given by
+```math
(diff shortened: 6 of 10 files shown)


EpiAware/src/epimodel.jl
```diff
@@ -60,24 +60,34 @@ struct DirectInfections <: AbstractEpiModel
...
 struct Renewal <: AbstractEpiModel
     data::EpiData
 end
 
-function (epi_model::Renewal)(recent_incidence, Rt)
-    new_incidence = Rt * dot(recent_incidence, epi_model.data.gen_int)
-    [new_incidence; recent_incidence[1:(epi_model.data.len_gen_int-1)]], new_incidence
+function (epimodel::Renewal)(_Rt, latent_process_aux)
+    I₀ = epimodel.data.transformation(latent_process_aux.init)
+    Rt = epimodel.data.transformation.(_Rt)
```
seabbs, line 81: This is fine for now but potentially dangerous. Likely a new issue to address in the later future (i.e it doesn't touch our analysis).

EpiAware/src/epimodel.jl
```diff
@@ -60,24 +60,34 @@ struct DirectInfections <: AbstractEpiModel
...
+
+    r_approx = fast_R_to_r_approx(Rt[1], epimodel)
```
seabbs, line 83: we don't need to call this fast and it isn't informative. I suggest `R_to_r` and then it can have methods later on or a name that explicitly touches on the method being being used.

EpiAware/src/latent-processes.jl
```diff
@@ -14,12 +14,12 @@ end
...
+    init ~ latent_process_priors.init_rw_value_dist
     σ_RW = sqrt(σ²_RW)
 
-    rw[1] = init_rw_value + σ_RW * ϵ_t[1]
+    rw[1] = σ_RW * ϵ_t[1]
     for t = 2:n
         rw[t] = rw[t-1] + σ_RW * ϵ_t[t]
     end
-    return rw, (; σ_RW, init_rw_value, init = rw[1])
+    return rw, (; σ_RW, init)
```
seabbs, line 24: I can see why you have coupled initialisation like this to the random walk but in a more general setting it might not be the correct choice (where you have random walks that stop and start on top of some other latent process). I suggest we leave as is and address if and when it comes up (which I don't think it will).

EpiAware/src/models.jl
```diff
@@ -12,11 +12,11 @@
 
     #Latent process
     time_steps = epimodel.data.time_horizon
-    @submodel latent_process, latent_process_parameters =
+    @submodel latent_process, latent_process_aux =
         latent_process(time_steps; latent_process_priors = latent_process_priors)
 
     #Transform into infections
-    I_t, _ = scan(epimodel, latent_process_parameters.init, latent_process)
+    I_t = epimodel(latent_process, latent_process_aux)
```
seabbs, line 19: Thanks for updating. I much prefer this structure.

EpiAware/src/utilities.jl
```diff
@@ -109,6 +109,44 @@ function create_discrete_pmf(dist::Distribution; Δd = 1.0, D)
...
+    # Small r approximation as initial guess
+    r_approx = (R₀ - 1) / (R₀ * mean_gen_time)
+    # Newton's method
+    for _ = 1:newton_steps
+        r_approx -= (R₀ * neg_MGF(r_approx, w) - 1) / (R₀ * dneg_MGF_dr(r_approx, w))
+    end
+    return r_approx
+end
+
+function fast_R_to_r_approx(R₀, epimodel::AbstractEpiModel; newton_steps = 4)
```
seabbs, line 144: you have inconsistent naming with the function below. I suggest standardising on R and r or growth and reproductive ratio.

EpiAware/src/utilities.jl
```diff
@@ -109,6 +109,44 @@ function create_discrete_pmf(dist::Distribution; Δd = 1.0, D)
...
+"""
+function neg_MGF(r, w::AbstractVector)
+    return sum([w[i] * exp(-r * i) for i = 1:length(w)])
+end
+
+function dneg_MGF_dr(r, w::AbstractVector)
+    return -sum([w[i] * i * exp(-r * i) for i = 1:length(w)])
+end
+
+function fast_R_to_r_approx(R₀, w::Vector{T}; newton_steps = 1) where {T<:AbstractFloat}
```
seabbs, line 133: no docs? Some docs would be better than one here I think even if we are going to update them at a later date.

EpiAware/src/utilities.jl
```diff
@@ -109,6 +109,44 @@ function create_discrete_pmf(dist::Distribution; Δd = 1.0, D)
...
+    mean_gen_time = dot(w, 1:length(w))
```
seabbs, line 134: This assumes that the generation time is indexed from 1 which we enforce as an input? If so we need to say that somewhere.

EpiAware/test/predictive_checking/fast_approx_for_r.jl
```diff
@@ -0,0 +1,73 @@
+#=
+# Fast approximation for `r` from `R₀`
+I use the negative moment generating function (MGF).
```
seabbs, line 3: Why isn't this in the docs for the function?

EpiAware/test/predictive_checking/fast_approx_for_r.jl
```diff
@@ -0,0 +1,73 @@
...
+r_0 = { \mathcal{R}_0 - 1 \over  \mathcal{R}_0 \langle W \rangle }.
+```
+where $\langle W \rangle$ is the mean of the generation interval.
+
+To rapidly improve the estimate for `r` we use Newton steps given by
+```math
+r_{n+1} = r_n - {\mathcal{R}_0 G(r) - 1\over \mathcal{R}_0 G'(r)}.
+```
+
+=#
```
seabbs, line 32: Putting testenv.activate in these scripts means they are very annoying to run more than once. Perhaps there is a better approach here?

EpiAware/test/predictive_checking/fast_approx_for_r.jl
```diff
@@ -0,0 +1,73 @@
...
+TestEnv.activate()
+using EpiAware
+using Distributions
+using StatsPlots
+
+# Create a discrete probability mass function (PMF) for a negative binomial distribution
+# with left truncation at 1.
+
+w =
+    create_discrete_pmf(NegativeBinomial(2, 0.5), D = 20.0) |>
```
seabbs, line 44: is it good style to start this on a new line?

EpiAware/test/predictive_checking/fast_approx_for_r.jl
```diff
@@ -0,0 +1,73 @@
+#=
+# Fast approximation for `r` from `R₀`
```
seabbs, line 2: How are you rendering these quickly? I am having to using literate.jl -> rename to jmd -> weave.jl and its pretty dull

EpiAware/src/utilities.jl
seabbs: your two fast_R_to_r_approx functions have inconsistent defaults for newton_steps. This should likely be standarised. It also seems from your testing that 4 steps is really more than is required for the level of numerical accuracy I assume we actually benefit from here. 2 seems more than enough?

EpiAware/test/predictive_checking/fast_approx_for_r.jl
```diff
@@ -0,0 +1,73 @@
...
+```math
+f(r, \mathcal{R}_0) = \mathcal{R}_0 G(r) - 1.
+```
+then the connection between `R₀` and `r` is given by
+```math
+f(r, \mathcal{R}_0) = 0.
+```
+Given an estimate of $\mathcal{R}_0$ we implicit solve for $r$ using a root
+finder algorithm. In this note, I test a fast approximation for $r$ which
+should have good autodifferentiation properties. The idea is to start from the
```
seabbs, line 19: you say should here and you say it does in the comment on Dylan's PR but I see not testing evidence?

EpiAware/test/test_models.jl
```diff
@@ -28,14 +28,90 @@
...
+    @test gen.I_t ≈ expected_I_t
+end
+
+@testitem "Renewal with RW latent process" begin
+    using Distributions, Turing, DynamicPPL
+    # Define test inputs
+    y_t = missing # Data will be generated from the model
+    data = EpiData([0.2, 0.3, 0.5], [0.1, 0.4, 0.5], 0.8, 10, exp)
+    latent_process_priors = default_rw_priors()
+    transform_function = exp
```
seabbs, line 86: I don't see this used?

EpiAware/test/test_utilities.jl
```diff
@@ -82,6 +82,7 @@ end
 end
 
 @testitem "Testing growth_rate_to_reproductive_ratio function" begin
+    using EpiAware
```
seabbs, line 85: why is this needed here but not elsewhere?

EpiAware/test/test_utilities.jl
```diff
@@ -119,3 +120,51 @@ end
     end
 
 end
```
seabbs, line 122: space

EpiAware/test/test_utilities.jl
```diff
@@ -119,3 +120,51 @@ end
...
+@testitem "Testing neg_MGF function" begin
```
seabbs, line 123: no test for the overall approximation?

seabbs (changes requested):
This is really nice. I like the reorg. Some general comments about naming, docs etc which feel a little messy here. 

I really like the use of $R_0$ here to define the initial growth rate and the approximation approach makes sense and seems correct. From a reviewer side though this really feels like another PR and would have been easier to review if that were the case. I can see why you included it here though. 

In terms of the renaming etc I think I would again suggest we move many of these comments to issues so we can keep up the velocity. The only change I would really like to see here is an explicit test of the approximation approach.

EpiAware/src/EpiAware.jl
```diff
@@ -35,16 +35,19 @@ export scan,
     create_discrete_pmf,
     growth_rate_to_reproductive_ratio,
     generate_observation_kernel,
-    default_rw_priors
+    default_rw_priors,
+    neg_MGF,
```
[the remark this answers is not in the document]
SamuelBrand1: Don't have to; its a purely stylistic choice.

EpiAware/src/epimodel.jl
SamuelBrand1: The reason for `fast_` is that other options were eigenvalue or more black box root finding.

EpiAware/src/models.jl
SamuelBrand1: Me too. `scan` is only really useful for the Renewal formulation.

EpiAware/test/predictive_checking/fast_approx_for_r.jl
SamuelBrand1: The test script is [here](https://github.com/CDCgov/Rt-without-renewal/blob/40-initialisation-infection-generation-processes/EpiAware/test/predictive_checking/fast_approx_for_r.jl).

I'm not sure how much testing this needs? Its a solvable small $r$ approximation to start a small number of Newton method steps. The main failure mode is incorrectly writing the adjoint method for the PGF of the generation distribution which you can diagnose visually (because it will diverge or go to wrong value fast).

EpiAware/test/test_models.jl
SamuelBrand1: Good spot. This is left over from a previous API to the main inference function.

EpiAware/src/utilities.jl
SamuelBrand1: Good spot

EpiAware/src/utilities.jl
SamuelBrand1: Added some docs.

EpiAware/src/utilities.jl
SamuelBrand1: Yes we need to make that clear somewhere.

EpiAware/test/predictive_checking/fast_approx_for_r.jl
SamuelBrand1: I run as a script saving any figures as an asset and then render literate -> github flavoured markdown.

The reason for using literate is that when we go to Documenter it renders to documenter flavoured md which is advantageous.

If there wasn't an eventual Documenter target, I'd probably either literate -> jupyter OR use Pluto on local Pkg mode

EpiAware/test/predictive_checking/fast_approx_for_r.jl
SamuelBrand1: I've added some now.

EpiAware/test/predictive_checking/fast_approx_for_r.jl
SamuelBrand1: I hadn't noticed! I run that bit as a code block once; then develop interactively on VS-Code.

I'll have a think.

EpiAware/test/test_utilities.jl
SamuelBrand1: Its not.

EpiAware/test/test_utilities.jl
SamuelBrand1: done

EpiAware/test/test_utilities.jl
SamuelBrand1: As in the R -> r approximation?

EpiAware/src/epimodel.jl
SamuelBrand1: But I'll defer to your wisdom here!

EpiAware/test/predictive_checking/fast_approx_for_r.jl
SamuelBrand1: As someone you does a dev pattern of 

`julia runscript.jl`

what would work best for you?

EpiAware/test/predictive_checking/fast_approx_for_r.jl
SamuelBrand1: Fixed

EpiAware/test/predictive_checking/fast_approx_for_r.jl
SamuelBrand1: I've updated the doc to the `R_to_r` function.

EpiAware/test/test_models.jl
SamuelBrand1: Fixed

EpiAware/test/test_utilities.jl
SamuelBrand1: Removed

EpiAware/src/epimodel.jl
SamuelBrand1: I've changed the name.

EpiAware/src/EpiAware.jl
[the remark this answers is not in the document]
seabbs: I'd generally go with only exporting things we want people to use/we will use elsewhere but I don't mind and I don't understand the landscape

In R this wouldn't be a stylistic choice as you are making a statement that you expect users to make use of a given function and therefore wont remove or break it without depreciation notice.

EpiAware/src/epimodel.jl
seabbs: my point is more that we should be able to dispatch to different methods from the same function right? vs naming the function and then having to change that name

seabbs: also "wisdom" - love it

EpiAware/test/predictive_checking/fast_approx_for_r.jl
seabbs: So what I am doing is just rerunning the whole script in the REPL - maybe that implies that I am just using the wrong workflow here?

EpiAware/test/predictive_checking/fast_approx_for_r.jl
seabbs: my complaint here is specifically about evidence that "it works well under AD" but I am happy that it probable does (can see little reason it wouldn't)

EpiAware/test/test_utilities.jl
seabbs: yes

seabbs: as in you have it your interactive check but not in an actual test. I think we should just make a new issue for moving these prior and posterior checks into the docs and unit tests tbh

EpiAware/test/predictive_checking/fast_approx_for_r.jl
SamuelBrand1: Ah right! OK that is based on more general principles, e.g. that this function pre-defines the number of iterations it will do rather than basing it on abstol etc.

seabbs: got to explain more Sam for us slow folk :)

EpiAware/src/EpiAware.jl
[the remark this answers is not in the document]
seabbs: > It is common to export names which form part of the API (application programming interface). In the above code, the export list suggests that users should use nice and DOG. However, since qualified names always make identifiers accessible, this is just an option for organizing APIs: unlike other languages, Julia has no facilities for truly hiding module internals.

This makes me think internal functions shouldn't be in the export list. I will make an issue to address.

seabbs (approved):
LGTM. Making an issue for internal functions now.

PR #44, opened 2024-02-19
Add observation process as a Turing model

SamuelBrand1:
This PR modifies PR #42 to add the observation model as a `Turing` model instead of hard baking the observation model into the overall inference process.

The main change is [here](https://github.com/CDCgov/Rt-without-renewal/blob/a6ecd501c404ae2264837a8342656deec6ecadc5/EpiAware/src/models.jl#L5), and the observation model is defined [here](https://github.com/CDCgov/Rt-without-renewal/blob/a6ecd501c404ae2264837a8342656deec6ecadc5/EpiAware/src/observation-processes.jl#L1) with test [here](https://github.com/CDCgov/Rt-without-renewal/blob/41-observation-process-as-a-turing-model/EpiAware/test/test_observation-processes.jl).

The evolving toy model script [here](https://github.com/CDCgov/Rt-without-renewal/blob/41-observation-process-as-a-turing-model/EpiAware/test/predictive_checking/toy_model_log_infs_RW.jl) has been updated and runs.

### DIFF
--- a/EpiAware/src/models.jl
+++ b/EpiAware/src/models.jl
@@ -1,15 +1,11 @@
 @model function make_epi_inference_model(
     y_t,
     epimodel::AbstractEpiModel,
-    latent_process;
+    latent_process,
+    observation_process;
     latent_process_priors,
     pos_shift = 1e-6,
-    neg_bin_cluster_factor = missing,
-    neg_bin_cluster_factor_prior = Gamma(3, 0.05 / 3),
 )
-    #Prior
-    neg_bin_cluster_factor ~ neg_bin_cluster_factor_prior
-
     #Latent process
     time_steps = epimodel.data.time_horizon
     @submodel latent_process, latent_process_aux =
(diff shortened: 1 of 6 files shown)


EpiAware/src/models.jl
```diff
@@ -1,15 +1,11 @@
 @model function make_epi_inference_model(
     y_t,
     epimodel::AbstractEpiModel,
-    latent_process;
+    latent_process,
+    observation_process;
     latent_process_priors,
```
seabbs, line 6: this how now become all priors? Does this name really still make sense?

seabbs (commented):
This is nice. Not clear why its aimed at another PR vs being modular (ease?). All looks good but one complaint about how priors are now being passed. I'd suggest waiting to resolve issues with #42 prior to merging this one

seabbs (approved):
As its just the prior issue I will open another issue for that and okay this.

EpiAware/src/models.jl
seabbs: Wee #45

PR #47, opened 2024-02-19
Reduce number of exported functions

### LINKED ISSUE #46: Decide on what functions should be exported
As per the title we need to decide which functions should be exported (i.e. are designed to be used by a module user vs inside the module). We should also decide on a naming scheme for internal methods and check that all current code works as expected when these are not exported. 

This issue was based on discussion in #40 where it was noted that several of the functions appeared to be for internal use only but were still being exported (i.e. we had a by default export all policy).

SamuelBrand1:
All functions were being exported by `EpiAware`. Following discussion, in particular #46 , I've decided to expose _only_ functions that:

- Define the model (defined as latent process, latent process to infections function, observation process, their priors).
- Are going to be used in inference.

This PR is my first pass on this. Tests for internal functions have `EpiAware.my_function` as prefix in the `test` folder.

closes #46

### DIFF
--- a/EpiAware/src/EpiAware.jl
+++ b/EpiAware/src/EpiAware.jl
@@ -31,18 +31,10 @@ using Distributions,
     QuadGK
 
 # Exported utilities
-export scan,
-    create_discrete_pmf,
-    growth_rate_to_reproductive_ratio,
-    generate_observation_kernel,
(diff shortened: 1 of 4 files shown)


seabbs (approved):
LGTM. Nice one

Issue #45, opened 2024-02-19
Improve the prior interface for `make_epi_model_inference` 


seabbs:
As the title says we should be able to improve the prior interface. The simplest improvement would be to generalise the argument name for the latent process priors to reflect that it is used across all submodules. An additional improvement would be to think more deeply about the coupling of priors to model submodules.

SamuelBrand1:
> The simplest improvement would be to generalise the argument name for the latent process priors to reflect that it is used across all submodules.

Yes, atm I just use `merge` on the `NameTuple`s of each set of prior distributions.

> An additional improvement would be to think more deeply about the coupling of priors to model submodules.

I'm going to try and prototype this with the easy fix as back up. It _shouldn't_ be very hard. Basic idea is to wrap the function that makes a `Turing` model for each process with its priors in a struct. **If** you think thats cleaner than having a `NamedTuple` of priors lets do that.

The essential question is:

- Should all the priors "live" together in code as one object?
- Should the priors of a particular process "live" with the process definition as one object?

seabbs:
> I'm going to try and prototype this with the easy fix as back up. It shouldn't be very hard. Basic idea is to wrap the function that makes a Turing model for each process with its priors in a struct. If you think thats cleaner than having a NamedTuple of priors lets do that.

Yes I think this is a nice idea. 

> Should the priors of a particular process "live" with the process definition as one object?

This seems cleanest to me and I think likely to have fewer failure states.

PR #54, opened 2024-02-20
Do inference on generated data

### LINKED ISSUE #43: Do inference on generated data using the "toy model" in conjunction with the (already existing) RW latent process.

SamuelBrand1:
This PR demonstrates adding Bayesian inference to a combined `EpiModel`, latent process and observation process together for a [toy model](https://github.com/CDCgov/Rt-without-renewal/blob/43-do-inference-on-generated-data-using-this-in-conjunction-with-the-already-existing-rw-latent-process/EpiAware/test/predictive_checking/toy_model_log_infs_RW.jl).

#### Specific options

- `EpiModel` is `DirectInfections`; that is the latent process is aimed at generating the log-infections directly (as opposed to $r_t$ or $\mathcal{R}_t$.
- The latent process is `random_walk`; that is the latent process is a random walk with unknown step size variance.
- The observation process is `delay_observations`; that is the infections are delayed by a delay kernel and then cases are observed using negative binomial sampling.

#### Minor additional fixes

- `latent_process_priors` is renamed to `process_priors`. This contains the prior distributions of all processes rather than just the latent generative process; this is the simplest in code (see #45 ), but might be changed soon.
- Tweaked `random_walk` to engage type stability.
- Fixed `Direct_Infections` so that it started at the inferred `init` value.

Closes #43


EpiAware/src/epimodel.jl
```diff
@@ -61,7 +61,7 @@ struct DirectInfections <: AbstractEpiModel
 end
 
 function (epimodel::DirectInfections)(_It, latent_process_aux)
-    epimodel.data.transformation.(_It)
+    epimodel.data.transformation.(latent_process_aux.init .+ _It)
```
seabbs, line 64: Noting that passing in a list  feels like a bit of a bad pattern here. I would usually have thought specifiying the argumments explicitly and then calling the list of args on the function would be a better approach?

EpiAware/src/latent-processes.jl
```diff
@@ -5,18 +5,14 @@ function default_rw_priors()
...
-    ::Type{T} = Float64;
-    latent_process_priors = default_rw_priors(),
-) where {T<:AbstractFloat}
-    rw = Vector{T}(undef, n)
+@model function random_walk(n; latent_process_priors = default_rw_priors())
     ϵ_t ~ MvNormal(ones(n))
     σ²_RW ~ latent_process_priors.var_RW_dist
     init ~ latent_process_priors.init_rw_value_dist
     σ_RW = sqrt(σ²_RW)
+    rw = Vector{eltype(ϵ_t)}(undef, n)
```
seabbs, line 13: can you say a bit more about what motivates this change?

EpiAware/src/latent-processes.jl
```diff
@@ -5,18 +5,14 @@ function default_rw_priors()
...
 
+    # rw = σ_RW * cumsum(ϵ_t)
```
seabbs, line 15: either remove or justify why we need commented out code

seabbs (approved):
Looks good to me!

seabbs (commented):
(once above conversations are resolved this will auto-merge)

seabbs (commented):
(once above conversations are resolved this will auto-merge)

EpiAware/src/epimodel.jl
SamuelBrand1: The reason for using this pattern is that its flexible enough to be represented as:

https://github.com/CDCgov/Rt-without-renewal/blob/2e18d00529a716226088a2df1d1016059b884385/EpiAware/src/models.jl#L13-L16

for any combination of latent process, and method for going from latent process -> actual infections.

At the moment, we are assuming that the generative distribution for initialisation "lives" inside the latent process as a random value `init` among whatever auxillary values are generated by the latent process. I could see an argument for this being moved elsewhere.

Passing a `NamedTuple` is pretty common, I don't think its a bad pattern. For example, in [LogDensityProblems.jl](https://github.com/tpapp/LogDensityProblems.jl) the standard pattern is to define a data struct contain the fixed data of an inference problem with a callable function taking a `NamedTuple` arg and outputing the lpd.

seabbs: So we could instead use `...args` here? The advantage of that is then all the input requirements of a specific latent process would be clearly documented within the function specification. 

If you test and document the `NamedTuple` input requirements then this is less of an issue but I've seen it go wrong a fair few times (and people just forget what inputs functions are using etc).

seabbs: To be clear I am not opposed to this but just stating some potential issues. I think if we move to a struct for submodules this becomes less of an issue as the object that contains the function will be unique vs being a generic shared prior structure.

EpiAware/src/latent-processes.jl
SamuelBrand1: The previous `function(...args...) where T` pattern didn't work with `@submodel`. 

Rather than burrow hard into that this pattern is flexible enough to give type information to the compiler in the case that `ϵ_t` values aren't Floats (this can happen in forward mode autodiff where `Dual` type numbers are used).

EpiAware/src/latent-processes.jl
SamuelBrand1: That was a mistake leave. Pushed the removed.

EpiAware/src/epimodel.jl
SamuelBrand1: > To be clear I am not opposed to this but just stating some potential issues. I think if we move to a struct for submodules this becomes less of an issue as the object that contains the function will be unique vs being a generic shared prior structure.

I think this is the way to go ala #45 .

SamuelBrand1: > So we could instead use `...args` here? The advantage of that is then all the input requirements of a specific latent process would be clearly documented within the function specification.
> 
> If you test and document the `NamedTuple` input requirements then this is less of an issue but I've seen it go wrong a fair few times (and people just forget what inputs functions are using etc).

I think the callable function from `EpiModel` subtype to actual infections only needs 
1. A latent process input
2. An `init` value (which will virtually always be unconstrained/transformed time zero incidence).

SamuelBrand1: I'm going to alter the pattern to reflect that.

SamuelBrand1: I've just pushed a minor change where the `init` value is:
- Directly outputted by the latent process
- used as a direct arg in `epimodel`

rather than in both cases being "hidden" as the optional field of a `NamedTuple` object (with the potential gotchas you mention).

Doesn't change the reasons for #45 though.

PR #57, opened 2024-02-20
New function `spread_draws` for creating `tidybayes` compliant MCMC output

### LINKED ISSUE #55: Pipe output to a `tidybayes` compliant output

SamuelBrand1:
This PR proposes a new function: `spread_draws`

https://github.com/CDCgov/Rt-without-renewal/blob/048fd0a5c01b547eeae119f6fcb9d83dae28dd0a/EpiAware/src/utilities.jl#L236-L255

This converts a `Chains` object return from running Bayesian inference using the `Turing` `sample` method into a `DataFrame` which conforms to the standard output expected from the `tidybayes` function of the [same name](https://mjskay.github.io/tidybayes/reference/spread_draws.html).

#### Added dependencies:
- `DataFramesMeta`. This exports `DataFrames` functionality as well as additional macros for `dplyr`-like interaction with dataframe objects. In particular, the constructor `DataFrame` dispatches on `Chains` objects to create a DataFrame that is _nearly_ `tidybayes` compliant.
- `CSV`. This exports utilities for using `.csv` files. This has been added to the `/test` folder environment to expose `CSV.write`.

Closes #55

### DIFF
--- a/EpiAware/test/predictive_checking/toy_model_log_infs_RW.jl
+++ b/EpiAware/test/predictive_checking/toy_model_log_infs_RW.jl
@@ -57,7 +57,7 @@ TestEnv.activate()
 
 =#
 
-# using TestEnv # Run in Test environment mode
+## using TestEnv # Run in Test environment mode
 # TestEnv.activate()
 
 using EpiAware
@@ -67,6 +67,9 @@ using StatsPlots
 using Random
 using DynamicPPL
 using Statistics
+using DataFramesMeta
+using CSV # For outputting the MCMC chain
+
 Random.seed!(0)
 
 #=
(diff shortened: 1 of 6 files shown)


EpiAware/test/predictive_checking/toy_model_log_infs_RW.jl
```diff
@@ -67,6 +67,9 @@ using StatsPlots
 using Random
 using DynamicPPL
```
seabbs, line 68: Is this intensionally missing things from the test environmment? Like StatsPlots?

SamuelBrand1: `StatsPlots` is in the test env but not a dependency of `EpiAware`; which I think is the right move.

PR #62, opened 2024-02-21
Issue 58: Add new GitHub Actions

### LINKED ISSUE #58: Setup suggest Julia actions
Setup suggested Julia actions using `julia-actions`

seabbs:
This PR closes #58 by adding a range of GitHub Actions. Note that until all actions are passing it is not likely worht reviewing.


seabbs (commented):
Note it also moves us to use the `sciml` style for `JuliaFormatter` and runs this everywhere.

seabbs (commented):
Note it also moves us to use the `sciml` style for `JuliaFormatter` and runs this everywhere.

SamuelBrand1:
LGTM!

seabbs:
Note I am not 100% that coverage will work properly because of the workarounds I've had to put in place to get this all working in a monorepo. 

This might be starting to indicate that we want to move to a multi-repo structure but we can discuss that down the line I think

SamuelBrand1 (approved):

PR #65, opened 2024-02-23
Submodels take `kwargs` variable splits rather than `NamedTuple`s.

### LINKED ISSUE #63: Moving away from the `NamedTuple` to instead use argument splitting.
> ...we may move away from the `NamedTuple` to instead use argument splitting.

_Originally posted by @seabbs in https://github.com/CDCgov/Rt-without-renewal/pull/61#pullrequestreview-1894180933_

SamuelBrand1:
## Goal

This PR is aimed at addressing the problem outlines in #63 .

## API changes

None at top level API (e.g. running this [script](https://github.com/CDCgov/Rt-without-renewal/blob/main/EpiAware/test/predictive_checking/toy_model_log_infs_RW.jl). 

Where `NamedTuple`s are used at lower level we can splat their fields as `kwargs` e.g.

https://github.com/CDCgov/Rt-without-renewal/blob/8035d6032fcdc0eef9bbaa55cbb7802b38bab8e2/EpiAware/src/models.jl#L17-L26

Closes #63

### DIFF
--- a/EpiAware/src/observation-processes.jl
+++ b/EpiAware/src/observation-processes.jl
@@ -6,13 +6,14 @@ end
         y_t,
         I_t,
         epimodel::AbstractEpiModel;
-        kwargs...
+        neg_bin_cluster_factor_prior,
+        pos_shift
 )
     #Parameters
-    neg_bin_cluster_factor ~ kwargs[:neg_bin_cluster_factor_prior]
+    neg_bin_cluster_factor ~ neg_bin_cluster_factor_prior
 
     #Predictive distribution
-    case_pred_dists = (epimodel.data.delay_kernel * I_t) .+ kwargs[:pos_shift] .|>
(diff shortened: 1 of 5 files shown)


EpiAware/src/observation-processes.jl
```diff
@@ -6,14 +6,13 @@ end
         y_t,
         I_t,
         epimodel::AbstractEpiModel;
-        observation_process_priors = default_delay_obs_priors(),
-        pos_shift = 1e-6
+        kwargs...
```
seabbs, line 9: hm this isn't quite what I had in mind. Can you not specify the precise arguments here (i.e neg_bin_cluster_factor_prior) and then use splatting when calling the function to pattern match to these arguments? This doesn't really help docwise etc?

SamuelBrand1: Yes you can!

This was a mini brain fade.

seabbs (approved):
LGTM

PR #67, opened 2024-02-23
rename `_dist` postfixes to `_prior` when used as a prior

### LINKED ISSUE #66: standardise use of `_dist`  and  `_prior`
I've just had a thought that I'm using a postfix of `_dist` in some places and `_prior` in others to indicate the prior/generative distribution of parameters... We should choose one and go with that (happy for this to become a new issue).

_Originally posted by @SamuelBrand1 in https://github.com/CDCgov/Rt-without-renewal/issues/65#issuecomment-1961384648_

SamuelBrand1:
Simple PR.

When a `Distribution` is used as a prior in any `Turing` model the postfix `_prior` is used. Other uses, e.g. the underlying continuous distribution of the generation interval which is discretised have `_dist` postfix.

Closes #66

### DIFF
--- a/EpiAware/src/latent-processes.jl
+++ b/EpiAware/src/latent-processes.jl
@@ -1,14 +1,14 @@
 function default_rw_priors()
     return (
-        var_RW_dist = truncated(Normal(0.0, 0.05), 0.0, Inf),
-        init_rw_value_dist = Normal()
+        var_RW_prior = truncated(Normal(0.0, 0.05), 0.0, Inf),
+        init_rw_value_prior = Normal()
     )
 end
 
-@model function random_walk(n; var_RW_dist, init_rw_value_dist)
+@model function random_walk(n; var_RW_prior, init_rw_value_prior)
     ϵ_t ~ MvNormal(ones(n))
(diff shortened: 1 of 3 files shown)


EpiAware/src/latent-processes.jl
```diff
@@ -1,14 +1,14 @@
 function default_rw_priors()
     return (
-        var_RW_dist = truncated(Normal(0.0, 0.05), 0.0, Inf),
-        init_rw_value_dist = Normal()
+        var_RW_prior = truncated(Normal(0.0, 0.05), 0.0, Inf),
```
seabbs, line 3: note we have a lot of repetition in our name. We have a function that is called default_rw_priors and then everything in it is called rw as are all the args in the random walk function.

seabbs (approved):
LGTM. A comment unrelated to this PR about duplicative variable names.

EpiAware/src/latent-processes.jl
SamuelBrand1: I was just thinking this too. I'll spin out your comment to an issue.

Issue #75, opened 2024-02-26
Refactor to promote greater use of multiple dispatch and one paradigm programming


SamuelBrand1:
## Problem

In the code base as it exists there is a mix of coding paradigms; for example `<: AbstractEpiModel` objects implement latent process to latent infection functions as callables on their type; but otoh, a single `LatentProcess` type struct contains the latent process as a struct function.

This makes the codebase confusing and harder to extend, as well as violating the design principles of the Julia language.

## Solution

Move to a design principle where each part of the model has its _behaviour_ determined by a dispatch on a type and its _implementation_ determined by a method being dispatched upon.

seabbs:
Sounds good. Note this was motivated by a f2f discussion and agreed.

Once deciding issue is if methods that dispatch on a specific struct/dict should dispatch down to another method has the specific required arguments. 

This would meet my design goal of making all functions that aren't wrappers explicitly list their arguments (due to ease of testing, preventing potential bugs, and ease of documentation) but could be clunky and not required given that documentation etc could happen in the paired dict/struct. it also isn't clear to me that it would make sense to dispatch on the same type twice (with and without arguments). However this could open up a workflow where a struct partially contains a functions arguments with the rest passed when initialised (its not clear to me you would actually want this at all).

seabbs:
Note @SamuelBrand1 is working on this in https://github.com/CDCgov/Rt-without-renewal/tree/refactor-for-more-multi-dispatch

dylanhmorris:
> But could be clunky and not required given that documentation etc could happen in the paired dict/struct.

Isn't part of the appeal of type-determined function dispatch [^1] that it's reasonable for a function to "know" about the mandatory attributes of a/the struct(s) it gets passed? 

So this kind of Python should be avoided:

```python
def dist_from_origin(xy_coords: dict):
    return math.sqrt(xy_coords["x"] ** 2 + xy_coords["y"] ** 2)
```

Whereas this Julia seems fine to me:

```Julia
struct XYCoords
    x::Float32
    y::Float32
end

function dist_from_origin(xy_coords::XYCoords)
    return sqrt(xy_coords.x^2 + xy_coords.y^2)
end
```

Am I missing something?

[^1]: Multiple nor not

SamuelBrand1:
@dylanhmorris Absolutely, the refactor is to do more of this!

The other possible benefit is separating behaviour from implementation.

So to extend your example:

```julia
#Parametric type of XYCoords
[...]

PR #78, opened 2024-02-27
Refactor with method hierarchy

seabbs:
This is a draft PR that implements an alternative workflow (building on that discussed in #75) which uses multiple dispatch to allow use of both a struct and a function that takes arguments. 

Implementation:

https://github.com/CDCgov/Rt-without-renewal/blob/refactor-struct-to-arg/EpiAware/src/latent-processes.jl

See example use: 

```julia
priors = default_rw_priors()
n = 10  # Number of steps in the random walk

# Directly using the tag-based dispatch
latent_rw = latent_process(RandomWalkLatentProcessArg(), n, var_prior = truncated(Normal(0.0, 0.05), 0.0, Inf),
	init_prior = Normal())

# Create a RandomWalkLatentProcess instance with priors
rw_lp = RandomWalkLatentProcess(priors[:init_rw_value_prior], priors[:var_RW_prior])

# Using the struct-based dispatch
latent_rw_struct = latent_process(rw_lp, n)

# Attempting to call with a generic AbstractLatentProcess
latent_process(AbstractLatentProcess(), n)

# Attempting to call with a generic AbstractLatentProcessArg
latent_process(AbstractLatentProcessArg(), n)
```

## Con

- Adds an additional method that needs to be defined
- May not be a standard approach
- Leads to a slightly more complex stack trace
- Has no code impact for our use case

## Plus

- Can document the arguments in the funciton docs and then link out to the struct that carries them
- May make resuse of low level functions easier
- Increases code modularity (?)

### DIFF
--- a/EpiAware/src/latent-processes.jl
+++ b/EpiAware/src/latent-processes.jl
@@ -1,24 +1,41 @@
 abstract type AbstractLatentProcess end
+abstract type AbstractLatentProcessArg end
 
 struct RandomWalkLatentProcess{D <: Sampleable, S <: Sampleable} <: AbstractLatentProcess
     init_prior::D
     var_prior::S
 end
 
+struct RandomWalkLatentProcessArg <: AbstractLatentProcessArg end
+
 function default_rw_priors()
     return (:var_RW_prior => truncated(Normal(0.0, 0.05), 0.0, Inf),
         :init_rw_value_prior => Normal()) |> Dict
 end
 
-function generate_latent_process(latent_process::AbstractLatentProcess, n; kwargs...)
-    @info "No concrete implementation for generate_latent_process is defined."
+function latent_process(lp::AbstractLatentProcess, n;
+        kwargs...)
+    return latent_process(
+        AbstractLatentProcessArg(), n; var_prior = lp.var_prior, init_prior = lp.init_prior)
+end
+
+function latent_process(lp::AbstractLatentProcess, n; kwargs...)
+    @info "No concrete implementation for latent_process is defined."
     return nothing
 end
 
-@model function generate_latent_process(latent_process::RandomWalkLatentProcess, n)
+function latent_process(lp::RandomWalkLatentProcess, n)
+    return latent_process(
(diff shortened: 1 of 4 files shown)


EpiAware/src/latent-processes.jl
```diff
@@ -1,24 +1,41 @@
...
 
-@model function generate_latent_process(latent_process::RandomWalkLatentProcess, n)
+function latent_process(lp::RandomWalkLatentProcess, n)
+    return latent_process(
+        RandomWalkLatentProcessArg(), n; var_prior = lp.var_prior,
+        init_prior = lp.init_prior
+    )
+end
+
+@model function latent_process(lp::RandomWalkLatentProcessArg, n;
```
SamuelBrand1, line 34: It feels like the point of the above is to lower down to this function; but you can get the same effect by specifying the fields of `RandomWalkLatentProcess` (i.e. not using a `NamedTuple` or `Dict` of priors of potentially any name) and dispatching on that.

I'd avoid that by having `RandomWalkLatentProcess` have determined typed fields and saving a bunch of code repetition.

SamuelBrand1:
Its a pretty standard pattern to pass structs and dispatch on type information in Julia Packages; which I think you are trying hard to avoid here?

For example, in `DifferentialEquations.jl` a workflow might be:

1. create a `DEProblem` struct (with different behaviours being `ODEProblem`, `SDEProblem` etc.
2. Choose a solution method struct, e.g. `Tsit5`
3. pass the above to `solve` e.g. `solve(prob; Tsit5())` along with any other key words.

For example, scraping an example from [here](https://docs.sciml.ai/DiffEqDocs/stable/tutorials/faster_ode_example/)

```julia
function lorenz!(du, u, p, t)
    du[1] = 10.0 * (u[2] - u[1])
    du[2] = u[1] * (28.0 - u[3]) - u[2]
    du[3] = u[1] * u[2] - (8 / 3) * u[3]
    nothing
end

u0 = [1.0; 0.0; 0.0]
tspan = (0.0, 100.0)
prob = ODEProblem(lorenz!, u0, tspan)
solve(prob, Tsit5());
```

PR #83, opened 2024-02-28
Change `scan` so that compiler specialisation will be active

### LINKED ISSUE #82: Activate specialisation for functions with function input
## Problem

I think it is safer performance wise to activate type specialisation for functions that ingest other functions, see [this](https://docs.julialang.org/en/v1/manual/performance-tips/#Be-aware-of-when-Julia-avoids-specializing). 

## Solution

Fix this for `scan` and check other possible usages.

SamuelBrand1:
Minor fix of `scan` to address #82 . Also removed docstring for `scan` because being actively developed in response to #35 .

Closes #82

### DIFF
--- a/EpiAware/src/utilities.jl
+++ b/EpiAware/src/utilities.jl
@@ -1,5 +1,32 @@
 
-function scan(f::F, init, xs) where {F}
+"""
+    scan(f::F, init, xs) where {F <: AbstractEpiModel}
+
+Apply `f` to each element of `xs` and accumulate the results.
+
+`f` must be a [callable](https://docs.julialang.org/en/v1/manual/methods/#Function-like-objects)
+    on a sub-type of `AbstractEpiModel`.
+
+### Design note
+`scan` is being restricted to `AbstractEpiModel` sub-types to ensure:
+    1. That compiler specialization is [activated](https://docs.julialang.org/en/v1/manual/performance-tips/#Be-aware-of-when-Julia-avoids-specializing)
+    2. Also avoids potential compiler [overhead](https://docs.julialang.org/en/v1/devdocs/functions/#compiler-efficiency-issues)
+    from specialisation on `f<: Function`.
+
+
+
+# Arguments
+- `f`: A callable/functor that takes two arguments, `carry` and `x`, and returns a new
+    `carry` and a result `y`.
+- `init`: The initial value for the `carry` variable.
+- `xs`: An iterable collection of elements.
+
+# Returns
+- `ys`: An array containing the results of applying `f` to each element of `xs`.
+- `carry`: The final value of the `carry` variable after processing all elements of `xs`.
+
+"""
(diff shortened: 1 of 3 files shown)


EpiAware/src/utilities.jl
```diff
@@ -1,23 +1,5 @@
-"""
-    scan(f, init, xs)
-
-Apply a function `f` to each element of `xs` along with an accumulator hidden state with intial
```
seabbs, line 4: I'm not clear why we would remove this due to another issue and not just update it in that PR? It seems like this information will be useful to have in any work that improves docstrings?

EpiAware/src/utilities.jl
```diff
@@ -1,23 +1,5 @@
...
-    hidden state.
-- `init`: The initial accumulator value.
-- `xs`: An iterable collection of elements.
 
-# Returns
-- `ys`: An array containing the result values of applying `f` to each element of `xs`.
-- `carry`: The final accumulator value.
-"""
-function scan(f, init, xs::Vector{T}) where {T <: Union{Integer, AbstractFloat}}
+function scan(f::F, init, xs) where {F}
```
seabbs, line 2: Could you give more info on why this is a fix?

seabbs (changes requested):
Could you give a bit more on why this is a fix and what the impact on the fix is (the latter only if its not that much trouble). It looks a bit subjective from what I have seen so far.

seabbs:
Also I assume you checked that it wasn't already specialising?

EpiAware/src/utilities.jl
SamuelBrand1: Thats a good point.

seabbs (approved):
This looks good to me and reasoning makes sense as to the change.

SamuelBrand1:
In light of discussion here and on #82 I changed `scan` implementation to take a `where F <: AbstractEpiModel` argument.

The reason for this change is to 

1. avoid potential problems with specialisation on generic functions (flagged by @seabbs in #82).
2. Create safety for the code; the only context we want `EpiAware.scan` used is as part of the latent infection dynamics. 
   - The behaviour of latent infs (e.g. model structure) is now determined by the type of `EpiModel`.
   - The fixed parameters (e.g. GI) are immutable fields of an `EpiModel` object.
   - Optionally, a callable/functor for the `EpiModel` can be defined which defines the dynamics and is used with `EpiAware.scan`.

Having said the above simple performance testing shows no change between doing this and `scan(f::F,...) where {F}` or even `scan(f,...)`. 

Therefore, point 1. should be considered fairly subjective/based on general principles at this point and point 2. is more persuasive.

seabbs:
> Create safety for the code; the only context we want

This is really nice by the way

seabbs:
> Having said the above simple performance testing 

Its interesting it should make a difference but in practice doesn't really. I wonder if its due to julia docs being a little out of step with improvements that have been made or just this case. Shows I guess that its hard to reason on.

PR #85, opened 2024-02-28
Partial documentation of `EpiAware` and also sets documentation standard

SamuelBrand1:
This PR gives partial documentation to `EpiAware`, mainly the module.

Also gives a standard for using `DocStringExtensions`.


seabbs (commented):
This is looking really nice and from a skim all looks good. Any idea what is going on in CI?

SamuelBrand1:
> This is looking really nice and from a skim all looks good. Any idea what is going on in CI?

The answer is that I was not being careful in adding deps!

> EpiAware [b2eeebe4-5992-4301-9193-7ebc9f62c855] does not declare a compat entry for the following deps:
1-element Vector{Base.PkgId}:
 DocStringExtensions [ffbed154-4ef7-542d-bbb7-c09d3a79fcae]

Will fix in the morning!

seabbs:
sorry I haven't yet got to this in detail and also for the mess that #89 will bring to this

seabbs:
shall i look at this now? Do you want to ping me to review (top right when ready?)

SamuelBrand1:
> shall i look at this now? Do you want to ping me to review (top right when ready?)

Yeah, there is not alot to review:

- Use of `DocStringExtensions`? Is it good/meh/bad.
- LaTex math... using escape (as per this PR) or some other solution, see [here](https://docs.julialang.org/en/v1/manual/documentation/#Writing-Documentation).

Second one I only just noticed.

seabbs:
> @ doc raw

This seems like a cleaner solution than escaping to me for Latex but we could address in its own issue. I definitely like `DocStringExtensons` (and we should make a template)

EpiAware/src/EpiAware.jl
```diff
@@ -1,35 +1,47 @@
...
+         model to data. Implemented concrete subtypes:
+    - `Renewal`: A renewal process model for the spread of an infectious disease.
+    - `ExpGrowthRate`: An exponential growth rate model for the spread of an infectious
+         disease.
+    - `DirectInfections`: A model for the spread of an infectious disease based on modelling
+        direct infections.
+2. `AbstractLatentProcess`: Subtypes of this abstract type represent different latent
+     processes that can be used in an epidemiological model. Implemented concrete subtype:
+    - `RandomWalkLatentProcess`: A random walk latent process.
+3. `AbstractObservationProcess`: Subtypes of this abstract type represent different
```
seabbs, line 25:
```suggestion
4. `AbstractObservationModel`: Subtypes of this abstract type represent different
```

EpiAware/src/EpiAware.jl
```diff
@@ -1,35 +1,47 @@
...
-- LinearAlgebra: for linear algebra operations.
-- SparseArrays: for working with sparse arrays.
-- Random: for generating random numbers.
-- ReverseDiff: for automatic differentiation.
-- Optim: for optimization algorithms.
-- Zygote: for automatic differentiation.
+An epidemiological model in `EpiAware` consists of composable structs with core abstract
+    types. The core types are:
+
+1. `AbstractEpiModel`: Subtypes of this abstract type represent different models for the
```
seabbs, line 13:
```suggestion
2. `AbstractEpiModel`: Subtypes of this abstract type represent different models for the
```

seabbs:
> I did a quick try on that, and the problem was that raw kills using the DocStringExtension interpolation... tricky!

ah okay. Lets port to another issue for discussion.

SamuelBrand1:
I think this is ok now

seabbs:
Aweesome will review now. Could you resolve out the things that are sorted?

EpiAware/Project.toml
```diff
@@ -17,15 +18,16 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 Turing = "fce5fe82-541a-59a6-adf8-730c64b5f9a0"
 
 [compat]
+DataFramesMeta = "0.14"
 Distributions = "0.25"
-LinearAlgebra = "1.9"
+DocStringExtensions = "0.9"
 LogExpFunctions = "0.3"
+LinearAlgebra = "1.9"
```
seabbs, line 25: This is now in the wrong order vs the deps. (logexpfunction and linearalgebra)

EpiAware/src/EpiAware.jl
```diff
@@ -1,35 +1,48 @@
...
-- LinearAlgebra: for linear algebra operations.
-- SparseArrays: for working with sparse arrays.
-- Random: for generating random numbers.
-- ReverseDiff: for automatic differentiation.
-- Optim: for optimization algorithms.
-- Zygote: for automatic differentiation.
+An epidemiological model in `EpiAware` consists of composable structs with core abstract
+    types. The core types are:
+1. `AbstractModel`: This overarching type is used to abstract `Turing` models and is inheried by the other abstract types we use.
+``
```
seabbs, line 13:
```suggestion

```

seabbs (approved):
A few minor issues but otherwise good to go.

PR #87, opened 2024-02-28
Issue 84: Export all abstract types

### LINKED ISSUE #84: Should we export all abstract types?
If people are going to build there own modules that we would want to export all thee abstract types? (this is really for a new issue)

_Originally posted by @seabbs in https://github.com/CDCgov/Rt-without-renewal/pull/83#discussion_r1505807188_

seabbs:
This PR closes #84 by exporting all current abstract types.


SamuelBrand1 (approved):
LGTM

Issue #86, opened 2024-02-28
Implement a new abstract type AbstractProcess to link the various current abstract types and enable scan with it


seabbs:
As the title we likely want to link all current abstract types in order to allow us to slightly increase the scope for scan whilst still being restrictive (and hence safe).

SamuelBrand1:
I think this is a good idea.

PR #89, opened 2024-02-28
Issue 88 and 86: Update naming to be more consistent and add new overarching AbstractModel

### LINKED ISSUE #88: Rename Abstract**Process to Abstract**Model for consistency?
As the title

seabbs:
As the title this PR closes #88 and #86. 

It: 

- Uses model rather than process consistently
- Renames `generate_latent_process` -> `generate_latent` to match the pattern of the other modules.
- Renames `RandomWalkLatentProcess` -> `RandomWalk` again to match the pattern from the other module.
- Renames `make_epi_inference_model` -> `make_epi_aware` again for naming consistency and as discussed (this change lacks a connected issue but I think is in scope of this PR.
- Renames `epimodel` to `epi_model` again for consistency.
- Moves all abstract types into a single file and adds a heirarchy type called `AbstractModel`
- Generalises `scan` to expect a callable of type `AbstractModel`

Note as this change touches a lot of files it may cause issues for #85 - sorry!


SamuelBrand1 (approved):
This looks good.

Using `DocStringExtensions` lowers the pain of renaming, because done correctly it should just work

PR #94, opened 2024-02-29
Add some examples for documentation

SamuelBrand1:
This PR builds on #93 with the goal of adding some example usage/or underlying maths of `EpiAware`.

### Main addition
Three `.jl` scripts which were in a subdirectory of `/test` are now in the `doc/example` subdir. These were rendered to Documenter flavour markdown by `Literate.markdown` (not a dep of this package) in the `doc/src/man` folder. This means that the blocks tagged `@example` are run when documenter renders `md` to HTML whilst creating the documentation.

These have been added as links in the sidebar of the main doc page.

### Minor changes/sharp bits
- Julia code is run from the `EpiAware/doc` environment in the build process. The main usage expects a package available from the Julia registry; the alternative as suggested by Documenter guide was to include `src` in the `LOAD_PATH` environmental variable. This meant adding all the deps of `EpiAware` to the `EpiAware/doc` `Project.toml` which "felt" clunky but works.
- Commented out `linkcheck`. This caused an error, which debugging should be an issue. I _think_ this is because we have placeholder links in the documentation.
- As suggested in the Documenter guide I changed the formal kwarg in `makedocs` to:
```julia
format = Documenter.HTML(
        prettyurls = get(ENV, "CI", nothing) == "true"
    )
```
This means that local rendering still has active links when view from your browser.


EpiAware/docs/Project.toml
```diff
@@ -1,4 +1,16 @@
 [deps]
 Changelog = "5217a498-cd5d-4ec6-b8c2-9b85a09b6e3e"
+DataFramesMeta = "1313f7d8-7da2-5740-9ea0-a2ca25f37964"
+Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
-EpiAware = "b2eeebe4-5992-4301-9193-7ebc9f62c855"
```
seabbs, line 4: Could you explain again (and maybe point at the docs) why this is needed because it is horrible.

EpiAware/docs/examples/toy_model_log_infs_RW.jl
```diff
@@ -143,7 +141,8 @@ truth_data = random_epidemic.y_t
 model = make_epi_aware(truth_data, time_horizon, ; epi_model = epi_model,
     latent_model_model = rwp, observation_model = obs_model,
     pos_shift = 1e-6)
-@time chn = sample(model,
+
+chn = sample(model,
```
seabbs, line 145: Not for this PR but I assume we are going to make some kind of EpiAware specific sample here that users pathfinder and then NUTs with our tested defaults here?

EpiAware/docs/make.jl
```diff
@@ -1,12 +1,23 @@
-using Documenter
+using Pkg
+Pkg.instantiate() #This expecting to be run within the EpiAware/docs environment
```
seabbs, line 2: This part makes me deeply sad

seabbs (commented):
Overall this looks very nice. One request for clarification and also are you manually creating the examples.jl? 

I assume we actually want those to be generated in the github action vs committed or in make.jl?

EpiAware/docs/examples/discretized_pmfs.jl
```diff
@@ -81,7 +81,7 @@ plt1 = let
         xlabel = "Days",
         ylabel = "Probability")
 end
-savefig(plt1, joinpath(@__DIR__(), "assets/", "discrete_pmf_daily.png"))
+## savefig(plt1, joinpath(@__DIR__(), "assets/", "discrete_pmf_daily.png"))
```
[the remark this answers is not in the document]
SamuelBrand1: No we don't... this is leftover.

EpiAware/docs/Project.toml
SamuelBrand1: First let me check if it is _actually_needed

EpiAware/docs/examples/toy_model_log_infs_RW.jl
SamuelBrand1: Yes, everytime I improve something this script acts as end-to-end testing of the inference on generated data

EpiAware/docs/Project.toml
SamuelBrand1: You can make this better by using `dev`; see the usage [here](https://pkgdocs.julialang.org/v1/managing-packages/#developing).

This is fine locally; just checked. But for CI we might want to run `dev` on `EpiAware` before making the documentation.

Essentially, engineering has been done that interacting with the `Test` environment is clever and understand that we need to load the _developing_  code of `EpiAware`. 

The `doc` environment engineering seems to come from a use-case of we've made a package, put it on Github and probably Julia registry, and _now_ we want nice docs.

Another possibility is to `add` `EpiAware` direct from this open repo; in CI this would always get the most recent version (I think).

SamuelBrand1:
```julia
using Pkg
 Pkg.add(url="https://github.com/CDCgov/Rt-without-renewal/", subdir="EpiAware")
```

SamuelBrand1: I might try that pattern actually

seabbs: > You can make this better by using dev; see the usage 

So this is what I thought I had done but maybe not?

seabbs: > direct from this open repo; in CI this would always get the most recent version (I think).

We will have the code checked out in the action so we could add it not with a URL. that would be more stable/less likely to break

seabbs:
Do you want some support to unstick this @SamuelBrand1? Note the discussion in #99  that we likely want to deal with that here.

SamuelBrand1:
This PR suffered from @seabbs and myself trying to fix CI issues simultaneously and is being closed in favour of working #109

PR #96, opened 2024-02-29
Issue 91: Add version 1.10 to testing matrix

### LINKED ISSUE #91: Either test `EpiAware` on Julia 1.9 or update `[compat]`
As title.

Atm, the Project file `[compat]` requirement for julia is [1.9+](https://github.com/CDCgov/Rt-without-renewal/blob/dbaad6b1a2203ec29049658191c4b8aa3549d617/EpiAware/Project.toml#L31).

I don't think we have tested on `1.9`; personally I would be ok with changing to `1.10`, but if we keep `1.9` (or even earlier) we should test.

seabbs:
As the title this PR closes #91 by adding Julia version 1.10 to our testing matrix. It also adds these checks to our required list so that they will need to pass in order for PRs to be merged.

Not that this increases our minimum version from 1.9 to 1.10 and this is required as otherwise we need to update our toml file.


SamuelBrand1 (approved):
LGTM

PR #106, opened 2024-03-01
Issuee 98: Review and update exported functions

### LINKED ISSUE #98: Review what is and isn't exported
currently it is unclear what should and shouldn't be exported. We should review this. Update the current code base and make a clear policy for the future that we document in the documentation.

seabbs:
This PR closes #98 by exporting all functions we expect a user (ourselves) to use if constructing a new model.


seabbs:
The Document issue is unrelated to this PR and I think an issue with the templating that likely shouldn't have made it to `main`. @SamuelBrand1 do you agree with the proposed exports here?

SamuelBrand1 (approved):
LGTM

PR #107, opened 2024-03-01
Issue 118: First pass at missing data support

seabbs:
This PR adds support for missing data in the observation model by moving from a vectorised to loop approach. It also renames the negative binomial observation model to match the distribution naming scheme.

### DIFF
--- a/EpiAware/docs/src/examples/getting_started.jl
+++ b/EpiAware/docs/src/examples/getting_started.jl
@@ -239,7 +239,24 @@ We do the inference by MCMC/NUTS using the `Turing` NUTS sampler with default wa
 # ╔═╡ c8ce0d46-a160-4c40-a055-69b3d10d1770
 truth_data = generated_obs
 
+# ╔═╡ 4a4c6e91-8d8f-4bbf-bb7e-a36dc281e312
+md"
+The observation model supports partially complete data. To test this we set some of the generated observations to be `missing`.
+"
+
+# ╔═╡ 259a7042-e74f-43c7-aeb4-97a3beeb7776
+let
+    truth_data = Union{Int, Missing}[truth_data...]
+    truth_data[vcat([3, 5], 10:20)] .= missing
+end
+
+# ╔═╡ 32638954-2c99-4d4e-8e03-52154030c657
+md"
+We now make the model but fixing the initial condition of the random walk to be 0.
+"
+
 # ╔═╡ b4033728-b321-4100-8194-1fd9fe2d268d
+
 inference_mdl = fix(
     make_epi_aware(truth_data, time_horizon; epi_model = epi_model,
         latent_model = rwp, observation_model = obs_model),
@@ -296,6 +313,8 @@ let
         layout = (1, 2),
         size = (700, 400))
 end
+
+# ╔═╡ 2293b711-0bd0-44d5-8a30-94e56c5e4c65
 
 # ╔═╡ fd6321b1-4c3a-4123-b0dc-c45b951e0b80
 md"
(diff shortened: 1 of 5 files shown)


seabbs:
@SamuelBrand1 any thoughts here? The issue I had here was in trying to give both a vectorised and non-vectorised likelihood option but perhaps just accepting it can't be vectorised is the way forward for now?

SamuelBrand1:
> @SamuelBrand1 any thoughts here? The issue I had here was in trying to give both a vectorised and non-vectorised likelihood option but perhaps just accepting it can't be vectorised is the way forward for now?
> 
> It all seems surprisingly clunky to me as well so perhaps I have just got the wrong end of the stick/.

Yeah, the underlying maths here is that vectorised == multivariate distribution and non-vectorised == conditionally independent. 

Having a vector with element type `Union{Missing, Integer}` doesn't make much sense to also use `arraydist`, whereas we are asking it to be multivariate and conditionally independent at the same time. The reason we want that is because we believe that the performance of the `ReverseDiff` AD is better that way.

So I'd suggest just trying out a non-vectorised approach and checking if it really does have a big performance hit. The AD systems evolve independent of the PPL and so what might have been true when they wrote their performance tips might not be true now. If it doesn't then a loop is more flexible, missing elements will just work as expected.

SamuelBrand1:
Another option is to make the concept of `missing` part of the generative process; e.g. there is a probability model for observed cases (we have that) and another that filters that and unobserved cases on a day are given value `-1` (for example, and maintaining common eltype in what I imagine is an `Integer` array).

Then the usage would be `missing -> -1`.

seabbs:
> Another option is to make the concept of missing part of the generative process; e.g. there is a probability model for observed cases (we have that) and another that filters that and unobserved cases on a day are given value -1 (for example, and maintaining common eltype in what I imagine is an Integer array).

> Then the usage would be missing -> -1 in data wrangling.

I really don't love this and it feels very clunky but I see it would work and enable vectorisation.

seabbs:
Note I have had to update the getting started examples to use `generated_y_t` vs `y_t` which I think is an internal Turing approach to account for having both observed data (i.e. y_t) and generated data (i.e generated_y_t) which can now happen due to the support for partial missingness?

SamuelBrand1:
Do you have any sense that this is noticeably bad for inference time?

seabbs:
the only real evidence I have for this is running the getting started example. Comparing CI runs there is perhaps a small slowdown (https://github.com/CDCgov/Rt-without-renewal/actions) but it doesn't seem massive (though hard to know how that would scale given so little of that CI check is from inference). 

For interest in `epinowcast` we have this nice benchmark CI that really helps with these kinds of questions: https://github.com/epinowcast/epinowcast/pull/442#issuecomment-1968886852

SamuelBrand1:
> the only real evidence I have for this is running the getting started example. Comparing CI runs there is perhaps a small slowdown (https://github.com/CDCgov/Rt-without-renewal/actions) but it doesn't seem massive (though hard to know how that would scale given so little of that CI check is from inference).
> 
> For interest in `epinowcast` we have this nice benchmark CI that really helps with these kinds of questions: [epinowcast/epinowcast#442 (comment)](https://github.com/epinowcast/epinowcast/pull/442#issuecomment-1968886852)

Very nice. We can do that with [BenchmarkTools](https://github.com/JuliaCI/BenchmarkTools.jl) I think

SamuelBrand1:
> > Another option is to make the concept of missing part of the generative process; e.g. there is a probability model for observed cases (we have that) and another that filters that and unobserved cases on a day are given value -1 (for example, and maintaining common eltype in what I imagine is an Integer array).
> 
> > Then the usage would be missing -> -1 in data wrangling.
> 
> I really don't love this and it feels very clunky but I see it would work and enable vectorisation.

We could have the not missing (or missing indices) as input-able data?

SamuelBrand1:
@seabbs The example script doesn't use a `missing` / Int mix.

I've tried that out [here](https://github.com/CDCgov/Rt-without-renewal/blob/missing-data-mixed-vector/EpiAware/docs/src/examples/getting_started.jl), but then generated quantities starts to chuck an error. So sadly this doesn't work yet :-(.

seabbs:
> The example script doesn't use a missing / Int mix.

That is why I was using it for benchmarking as it should show the difference in speed for complete data (which is what we care about for the benchmark). 

I tested the generated quantities in tests but yes ideally we would have a small example to had partially complete data at some point (an issue and not a priority for now I think).

seabbs:
> e tried that out [here](https://github.com/CDCgov/Rt-without-renewal/blob/missing-data-mixed-vector/EpiAware/docs/src/examples/getting_started.jl?rgh-link-date=2024-03-06T21%3A26%3A57Z), but then generated quantities starts to chuck an error. So sadly this doesn't work yet :-(.

Where did you try this? Shouldn't it be here: https://github.com/CDCgov/Rt-without-renewal/blob/02fcf744cdaaa64bc4f85493e2da749eec516448/EpiAware/docs/src/examples/getting_started.jl#L240

SamuelBrand1:
> > e tried that out [here](https://github.com/CDCgov/Rt-without-renewal/blob/missing-data-mixed-vector/EpiAware/docs/src/examples/getting_started.jl?rgh-link-date=2024-03-06T21%3A26%3A57Z), but then generated quantities starts to chuck an error. So sadly this doesn't work yet :-(.
> 
> Where did you try this? Shouldn't it be here:
> 
> https://github.com/CDCgov/Rt-without-renewal/blob/02fcf744cdaaa64bc4f85493e2da749eec516448/EpiAware/docs/src/examples/getting_started.jl#L240

Ooops I hadn't committed my change

seabbs:
I added a small example here as part of the getting started example: https://github.com/CDCgov/Rt-without-renewal/pull/107/commits/e28cc8c90379a0dc3344fc432b8ffd8c17688a7b

SamuelBrand1:
> I added a small example here as part of the getting started example: [e28cc8c](https://github.com/CDCgov/Rt-without-renewal/commit/e28cc8c90379a0dc3344fc432b8ffd8c17688a7b)

Nice one. LGTM now.

PR #113, opened 2024-03-05
Remove attempt to detect `fieldnames` for functions which lead to error

### LINKED ISSUE #108: Documenter broken on main
See 

https://github.com/CDCgov/Rt-without-renewal/actions/workflows/document-EpiAware.yaml

I'm not clear if this was picked up before merging or how this slipped through but Documenter has been broken by merging the template PR I think (i.e #104)

SamuelBrand1:
This aims to fix #108 .

The problem appears to be that we were setting a document `@template` for function, methods and macros which included fieldnames. Those objects don't have fieldnames, so this was chucking an error (this really should be a warning IMO...).

### DIFF
--- a/EpiAware/docs/make.jl
+++ b/EpiAware/docs/make.jl
@@ -7,7 +7,7 @@ include("pages.jl")
 makedocs(; sitename = "EpiAware.jl",
     authors = "Samuel Brand, Zachary Susswein, Sam Abbott, and contributors",
     clean = true, doctest = true, linkcheck = true,
-    warnonly = [:docs_block, :missing_docs],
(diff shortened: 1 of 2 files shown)


SamuelBrand1:
These link errors point at where we've manually included docstrings with the proposed template; they were warnings in local build upgraded to full errors in CI

SamuelBrand1:
This passes and has fixed the main error, but I recommend if merged an issue on why the link checking was erroring.

seabbs:
> OK new error

This looks to me to be because we aren't including functions in the internal or public API blocks at all at the moment (because they were erroring due to lack of docs).

seabbs (approved):
LGTM

PR #117, opened 2024-03-06
Issue 56: Add pathfinder init

### LINKED ISSUE #56: pathfinder based initialisation of Bayesian Inference
## Background on NUTS initialisation with Pathfinder

[Pathfinder](https://dl.acm.org/doi/abs/10.5555/3586589.3586895) is an approach to variational inference, that was initially envisioned as a method for [pre-heating NUTS-based Bayesian inference](https://github.com/mlcolab/Pathfinder.jl). Performance of NUTS is improved by:

1. Starting from an (approximate) draw from the posterior distribution of the Bayesian inference problem.
2. Starting KE matrix adaptation from (inverse) Hessian around the MAP point of the posterior distribution of the Bayesian inference problem.

Pathfinder supplies both of the above using its [`pathfinder`](https://mlcolab.github.io/Pathfinder.jl/stable/lib/public/#Pathfinder.pathfinder) multivariate normal approximation to the posterior distribution. Also, [`multipathfinder`](https://mlcolab.github.io/Pathfinder.jl/stable/lib/public/#Pathfinder.multipathfinder) supplies improved initial draw approximation (its samples from a mixture distribution of multiple single pathfinder approximations passed though a Pareto importance sampling filter).

## Proposed improvement

Implement the pathfinder initialisation for `Turing` described [here](https://mlcolab.github.io/Pathfinder.jl/stable/examples/turing/).

## Additional questions

Is it worth lowering the inference call to `AdvancedHMC` so we can use metric initialisation?

SamuelBrand1:
This is a new draft PR aimed at addressing #56 .

The main innovation for this PR occurs here:

https://github.com/CDCgov/Rt-without-renewal/blob/ac788d04a0494e87fd66ccd8c4943437aaba1276/EpiAware/docs/src/examples/getting_started_pf.jl#L245-L291

The aim is to ultimately have `_epi_aware` as an _internal_method as part of a wider workflow. But for this PR it is getting this functionality available.

The current first pass behaviour is for `_epi_aware` to take in the information required to act on `make_epi_aware`; then do `multipathfinder` and pass the initial parameters to `sample` (as per this [workflow](https://mlcolab.github.io/Pathfinder.jl/stable/examples/turing/)).

Close #56


EpiAware/src/inference-methods.jl
```diff
@@ -0,0 +1,51 @@
...
+        nchains,
+        pf_ndraws = 10,
+        pf_nruns = 10,
+        fixed_parameters = (;),
+        pos_shift = 1e-6,
+        executor = Transducers.ThreadedEx(),
+        adtype = AutoReverseDiff(true),
+        maxiters = 10,
+        kwargs...)
+    gen_mdl = make_epi_aware(missing, time_steps; epi_model,
```
seabbs, line 20: rather than all this repeating (which is dull and dangerous) can make this a small function?

EpiAware/src/models.jl
```diff
@@ -25,3 +25,38 @@
...
+            pos_shift = pos_shift)
+
+        #Generate quantities
+        return (;
+            generated_y_t,
+            I_t,
+            latent_model,
+            process_aux = merge(latent_model_aux, generated_y_t_aux))
+    catch
+        Turing.@addlogprob! -Inf
```
seabbs, line 59: It seems not ideal to have to copy and past like this. Would this be better having the contents in an internal function?

seabbs (approved):
This is all really nice and looks good to me. It would likely be nice to make this a bit more flexible (i.e not force the user to use pathfinder). Most of my comments were about deduplicating code which can be made into new issues and addressed in linked PRs.

seabbs:
This is just waiting on these commets -> issues if we agree they should be dealt with and then this can be merged (resolving the conversations should allow auto-merge to do this)

EpiAware/src/inference-methods.jl
```diff
@@ -0,0 +1,51 @@
+"""
+
+Do inference on an EpiAware model.
+"""
+function _epi_aware(y_t,
```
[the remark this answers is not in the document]
SamuelBrand1: I think we should use it for all internal methods. Its more a question of getting rock solid on whats internal and whats external

EpiAware/docs/src/examples/getting_started.jl
```diff
@@ -183,40 +190,37 @@ Having chosen an `EpiModel`, `LatentModel` and `ObservationModel`, we can implem
...
-# ╔═╡ b20c28be-7b07-410c-a33b-ea5ad6828c12
+# ╔═╡ 4a1fc7bb-82a0-4643-a18b-d331a31c1390
 random_epidemic = rand(cond_generative_model)
 
-# ╔═╡ d073e63b-62da-4743-ace0-78ef7806bc0b
+# ╔═╡ e571e7b6-0e26-4855-ae90-05a18be6ff38
 true_infections = generated_quantities(cond_generative_model, random_epidemic).I_t
 
-# ╔═╡ a04f3c1b-7e11-4800-9c2a-9fc0021de6e7
-generated_obs = generated_quantities(cond_generative_model, random_epidemic).generated_y_t
```
seabbs, line 217: has this intentionally gone?

EpiAware/docs/src/examples/getting_started.jl
```diff
@@ -236,42 +240,61 @@ However, we now treat the generated data as `truth_data` and make inference with
 We do the inference by MCMC/NUTS using the `Turing` NUTS sampler with default warm-up steps.
 "
 
-# ╔═╡ c8ce0d46-a160-4c40-a055-69b3d10d1770
-truth_data = generated_obs
+# ╔═╡ 7e48a4c5-cd30-4377-8a98-e0c23f2dc31e
+truth_data = random_epidemic.y_t
```
seabbs: this all looks like it just needs to be revert to what is on main?

EpiAware/docs/src/examples/getting_started.jl
SamuelBrand1: No that was a merge error.

EpiAware/docs/src/examples/getting_started.jl
SamuelBrand1: Yes.

EpiAware/docs/src/examples/getting_started.jl
```diff
@@ -236,42 +240,61 @@ However, we now treat the generated data as `truth_data` and make inference with
...
 
-# ╔═╡ 4a4c6e91-8d8f-4bbf-bb7e-a36dc281e312
+# ╔═╡ 4298f0ec-f6df-42ee-aa28-f7ed60f1e530
 md"
-The observation model supports partially complete data. To test this we set some of the generated observations to be `missing`.
+### Initialising inference
+
+It is possible for the default warm-up process for NUTS to get stuck in low probability or otherwise degenerate regions of parameter space.
+
+To make NUTS more robust we provide `manypathfinder`, which is built on pathfinder variational inference through the [Pathfinder.jl](https://mlcolab.github.io/Pathfinder.jl/stable/). `manypathfinder` runs `nruns` pathfinder processes on the inference problem and returns the pathfinder run with maximum estimated ELBO.
```
seabbs, line 264:
```suggestion
To make NUTS more robust we provide `manypathfinder`, which is built on pathfinder variational inference from [Pathfinder.jl](https://mlcolab.github.io/Pathfinder.jl/stable/). `manypathfinder` runs `nruns` pathfinder processes on the inference problem and returns the pathfinder run with maximum estimated ELBO.
```

EpiAware/docs/src/examples/getting_started.jl
```diff
@@ -236,42 +240,61 @@ However, we now treat the generated data as `truth_data` and make inference with
...
+# ╔═╡ cdd805e2-b00c-4522-9261-1819c6a195eb
+best_pf.draws_transformed
 
-inference_mdl = fix(
-    make_epi_aware(truth_data, time_horizon; epi_model = epi_model,
-        latent_model = rwp, observation_model = obs_model),
-    (rw_init = 0.0,)
-)
+# ╔═╡ e847b0b6-9d70-46ba-bec6-1e3fa676a33c
+init_params = collect.(eachrow(best_pf.draws_transformed.value[1:4, :, 1]))
```
seabbs, line 281: Can we have a helper function for this or make it part of the output of best_pf?  I don't love reading this or the thought of having to repeatedly write it down.

EpiAware/docs/src/examples/getting_started.jl
```diff
@@ -280,9 +303,9 @@ We can spaghetti plot generated case data from the version of the model _which h
 Because we are using synthetic data we can also plot the model predictions for the _unobserved_ infections and check that (at least in this example) we were able to capture some unobserved/latent variables in the process accurate.
 "
 
-# ╔═╡ e9df22b8-8e4d-4ab7-91ea-c01f2239b3e5
+# ╔═╡ e74fc652-cd5f-4764-a416-caa8bab0bf0c
 let
-    post_check_mdl = fix(full_epi_aware_mdl, (rw_init = 0.0,))
+    post_check_mdl = fix(full_epi_aware_mdl, (init_rw = 0.0,))
```
seabbs, line 311: can you remind me why we fix here?

EpiAware/docs/src/examples/getting_started.jl
SamuelBrand1: I've done that.

EpiAware/src/inference-methods.jl
```diff
@@ -0,0 +1,97 @@
+"""
+Run pathfinder multiple times and store the results in an array. Fails safely.
+
+# Arguments
+- `mdl::DynamicPPL.Model`: The `Turing` model to be used for inference.
```
seabbs, line 5: what has happened to templating? Going forward we should be trying to use this vs having to do catch up documentation Prs.

EpiAware/src/inference-methods.jl
```diff
@@ -0,0 +1,97 @@
...
+- `nruns`: The number of times to run the `pathfinder` function.
+- `kwargs...`: Additional keyword arguments passed to `pathfinder`.
+
+# Returns
+An array of `PathfinderResult` objects or `Symbol` values indicating success or failure.
+"""
+function _run_manypathfinder(mdl::DynamicPPL.Model; nruns, kwargs...)
+    @info "Running pathfinder $nruns times"
+    pfs = Vector{Union{Pathfinder.PathfinderResult, Symbol}}(undef, nruns)
+    Threads.@threads for i in 1:nruns
```
seabbs, line 15: is threads built in (hence not needing to import it?)

EpiAware/docs/src/examples/getting_started.jl
SamuelBrand1: Fixed now

EpiAware/docs/src/examples/getting_started.jl
SamuelBrand1: Thats an issue now

EpiAware/src/inference-methods.jl
```diff
@@ -0,0 +1,97 @@
...
+- `kwargs...`: Additional keyword arguments passed to `pathfinder`.
+
+## Returns
+- `best_pfs::PathfinderResult`: Best pathfinder result by estimated ELBO.
+"""
+function manypathfinder(mdl::DynamicPPL.Model; nruns = 4, ndraws = 10,
+        nchains = 4, maxiters = 50, max_tries = 100, kwargs...)
+    ndraws = max(ndraws, nchains)
+    _run_manypathfinder(mdl; nruns, ndraws, maxiters, kwargs...) |>
+    pfs -> _continue_manypathfinder!(pfs, mdl; max_tries, kwargs...) |>
```
seabbs, line 95: So this is if everything in `_run_manypathfinder` failed have another sequential go until one does work? We probably want to tell people if `_run_manypathfinder` all failed and its now just trying for one right as that could indicate some bad things?

EpiAware/src/inference-methods.jl
SamuelBrand1: Yeah I ditched this for now

EpiAware/src/models.jl
SamuelBrand1: Yep ditched for now

EpiAware/test/test_inference-methods.jl
```diff
@@ -0,0 +1,135 @@
+@testitem "Testing _run_manypathfinder function" begin
+    using Turing, Pathfinder
```
seabbs, line 2: are you testing the all failure case here?

EpiAware/test/test_inference-methods.jl
```diff
@@ -0,0 +1,135 @@
...
+        maxiters = 100
+
+        pfs = EpiAware._run_manypathfinder(
+            mdl; nruns = nruns, ndraws = ndraws, maxiters = maxiters)
+
+        @test length(pfs) == nruns
+        @test all(p -> p isa Union{PathfinderResult, Symbol}, pfs)
+    end
+end
+@testitem "Testing _continue_manypathfinder! function" begin
```
seabbs, line 36: again testing the all failure case?

EpiAware/test/test_inference-methods.jl
```diff
@@ -0,0 +1,135 @@
...
+    using Turing, Pathfinder
+
+    @testset "Test case 1" begin
```
seabbs, line 39: Can wee say what the difference is between. these cases? they look very similar?

EpiAware/test/test_inference-methods.jl
```diff
@@ -0,0 +1,135 @@
...
+    pfs = EpiAware._run_manypathfinder(
+        mdl; nruns = nruns, ndraws = ndraws, maxiters = maxiters)
+
+    best_pf = EpiAware._get_best_elbo_pathfinder(pfs)
+    @test best_pf isa PathfinderResult
+end
+@testitem "Testing manypathfinder function" begin
+    using Turing, Pathfinder
+
+    @model function test_model()
```
seabbs, line 101: should this really live somewhere (like smoe kind of testing setup or something? It seems a bit mad to keep recreating it)

seabbs (changes requested):
This is much nicer I think. Some safety and quality of life comments + the obvious big merge issue.

seabbs:
Once this has stabilised and we have taken it out for a spin it would be interesting to float this wrapper to `Pathfinder.jl`

EpiAware/test/test_inference-methods.jl
SamuelBrand1: Yeah! Thats a good point.

EpiAware/src/inference-methods.jl
SamuelBrand1: It does shout a warn is things fail for `max_tries`

EpiAware/docs/src/examples/getting_started.jl
SamuelBrand1: It is a non-identifiability between `DirectInfections` initialisation and the latent process (plus this is a typo I've just corrected which reveals a brittleness of needing to know the names of variables internally...)

In general we might want `rw_init` to represent the initial distribution of (say) $r_t$ or $\mathcal{R}_t$, and the initial incidence is a separate concept. But for log-transformed initial infections _and_ log-latent process we want one to be fixed.

NB: I actually didn't do that due to the typo and it all still ran so its not critical, but it will have been bad for the sampler.

EpiAware/src/inference-methods.jl
SamuelBrand1: This works with the templating I think? So my local help gives:

```julia-repl
help?> manypathfinder
search: manypathfinder

  manypathfinder Run multiple instances of the pathfinder algorithm and returns the pathfinder run with the largest ELBO estimate.

  Arguments
  =========

    •  mdl::DynamicPPL.Model: The model to perform inference on.

    •  nruns::Int: The number of pathfinder runs to perform.

    •  ndraws::Int: The number of draws per pathfinder run, readjusted to be at least as large as the number of chains.

    •  nchains::Int: The number of chains that will be initialised by pathfinder draws.

    •  maxiters::Int: The maximum number of optimizer iterations per pathfinder run.

    •  max_tries::Int: The maximum number of extra tries to find a valid pathfinder result.

    •  kwargs...: Additional keyword arguments passed to pathfinder.

  Returns
  =======

    •  best_pfs::PathfinderResult: Best pathfinder result by estimated ELBO.

  ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

  Signatures
  ≡≡≡≡≡≡≡≡≡≡

  manypathfinder(
      mdl::DynamicPPL.Model;
      nruns,
      ndraws,
      nchains,
      maxiters,
      max_tries,
      kwargs...
  ) -> Any
  


  ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

  Methods
  =======

  manypathfinder(
      mdl;
      nruns,
      ndraws,
      nchains,
      maxiters,
      max_tries,
      kwargs...
  )


  defined at /Users/samandfi/GitHub/CFA/Rt-without-renewal/EpiAware/src/inference-methods.jl:91
  (https://github.com/CDCgov/Rt-without-renewal/tree/957944a5eeb7d71113dd6543f8c2b3f1e39372ad//EpiAware/src/inference-methods.jl#L91).
```

EpiAware/src/inference-methods.jl
SamuelBrand1: Yes.

EpiAware/src/inference-methods.jl
SamuelBrand1: But being verbose at the beginning would be a good idea too.

EpiAware/src/inference-methods.jl
SamuelBrand1: We should change the template so that the function name is on a line before the docstrings but that is a new issue?

SamuelBrand1: Actually also signature before doc string as well but again new issue

EpiAware/test/test_inference-methods.jl
SamuelBrand1: Probably not...

EpiAware/src/inference-methods.jl
SamuelBrand1: I've put in a warning at the beginning that its had to fall back on trying again.

EpiAware/test/test_inference-methods.jl
SamuelBrand1: I've made that an issue

EpiAware/test/test_inference-methods.jl
SamuelBrand1: I've beefed up the testing a fair bit now.

SamuelBrand1:
I've increased the testing for `manypathfinder`:

- More expected failures.
- Verbose fail if retries all fail.
- KS test that the best pathfinder run is a good approximation to a simple normal (if it isn't then thats an obvious problem).

EpiAware/test/test_inference-methods.jl
SamuelBrand1: I've beefed up the fail case testing

EpiAware/docs/src/examples/getting_started.jl
seabbs: > In general we might want rw_init to represent the initial distribution of (say) 
 > or 
> , and the initial incidence is a separate concept. But for log-transformed initial infections and log-latent process > we want one to be fixed.

Yes I think we might want to think about it like this. Sounds like a new issue to me.

> NB: I actually didn't do that due to the typo and it all still ran so its not critical, but it will have been bad for the sampler.

Do you think this is a bug? It feels like a bug or at least a really bad developer choice. Worth an issue over on `DynamicPPL`?

EpiAware/src/inference-methods.jl
seabbs: I thought we were using the extension tools to manage the construction of the args documentation etc as you had in the original docs PR (and the other similar extension tools). It sounds like this is still in the air though so perhaps we need another discussion issue? What I mean is that yes we have the template but I thought when writing docstrings we would be automating large parts of them using the extension tools to make them less error prone (l;like the arg types).

seabbs (commented):
We discussed face to face reordering the inputs to match `multipathfinder` but think that has yet to happen? Once that is done and we have a new issue for falling back on more functionality from `multipathfinder` I think this is good to go. It also sounds like we have some confusion about the aim with the docstrings but again perhaps that needs a new issue to discuss..

SamuelBrand1:
> We discussed face to face reordering the inputs to match `multipathfinder` but think that has yet to happen? Once that is done and we have a new issue for falling back on more functionality from `multipathfinder` I think this is good to go. It also sounds like we have some confusion about the aim with the docstrings but again perhaps that needs a new issue to discuss..

I've changed the input to match `manypathfinder` input to `multipathfinder`, e.g. `manypathfinder(mdl, ndraws; kwargs...) -> PathfinderResult` is the current signature.

SamuelBrand1:
After a quick scoping, I don't have immediate plans to use more `multipathfinder` functionality under the hood anymore.

EpiAware/src/inference-methods.jl
SamuelBrand1: OK, I've made that a new issue.

seabbs:
> After a quick scoping, I don't have immediate plans to use more multipathfinder functionality under the hood anymore.

Why

seabbs (approved):
LGTM

SamuelBrand1:
> > After a quick scoping, I don't have immediate plans to use more multipathfinder functionality under the hood anymore.
> 
> Why

I thought that the `try catch` pattern in `manypathfinder` could be replaced by `fail_on_nonfinite = false` in pathfinder `kwargs`, but the behaviour of the `Distributions.NegativeBinomial` distribution constructor chucks an error if we find a `NaN` during warm-up. `manypathfinder` is resilient to that fail mode.

PR #123, opened 2024-03-07
Changing the prior for negative binomial observations

### LINKED ISSUE #29: Sampling noise prior
https://github.com/CDCgov/Rt-without-renewal/blob/f7bf1c6f34ae9f40e76b53ff17ad865f98b2ca78/EpiAware/src/models.jl#L10

Let's think about this prior a bit in a new issue. It's a reasonable prior, but I think the PC prior using the inverse square root is usually recommended? This is on 1/k so it's close but not exactly the same. But sampling from that prior has also behaved pretty poorly for me.....

@seabbs -- I know you have thoughts here.

_Originally posted by @zsusswein in https://github.com/CDCgov/Rt-without-renewal/pull/28#discussion_r1486369014_

SamuelBrand1:
This addresses #29 using the link given by @seabbs in that issue.

Main change is here

https://github.com/CDCgov/Rt-without-renewal/blob/a048da5fa2e0aab079589815667b6f28fed47a86/EpiAware/src/observation-models.jl#L46-L61

The point is that we want to have a half-normal family prior for a standard deviation. In this case, its the standard deviation of the latent Gamma process implicit in the Negative Binomial definition. All this is based on this classic [paper](https://projecteuclid.org/journals/bayesian-analysis/volume-1/issue-3/Prior-distributions-for-variance-parameters-in-hierarchical-models-comment-on/10.1214/06-BA117A.full).

I've update the getting started note and the sampler seems happy with this new prior.

Closes #29


EpiAware/docs/src/examples/getting_started.jl
```diff
@@ -171,7 +171,7 @@ We choose a simple observation model where infections are observed 0, 1, 2, 3 da
 obs_model = EpiAware.DelayObservations(
     fill(0.25, 4),
     time_horizon,
-    truncated(Gamma(5, 0.05 / 5), 1e-3, 1.0)
+    truncated(Normal(0, 0.1 * sqrt(pi) / sqrt(2)), 0.0, Inf)
```
seabbs, line 174: This is a bit painful to expect people to write. Maybe we need an issue for a helper? Its quite low priority though obviously.

EpiAware/docs/src/examples/getting_started.jl
```diff
@@ -149,7 +149,7 @@ as the action of a sparse kernel on the infections $I(t)$.
 ```math
 \begin{align}
 y_t &\sim \text{NegBinomial}(\mu = \sum_{s\geq 0} K[t, t-s] I(s), r), \\
-1 / r &\sim \text{Gamma}(3, 0.05/3).
+1 / \sqrt{r} &\sim \text{HalfNormal}\Big(0.1 \sqrt{{\pi \over 2}}\Big).
```
seabbs, line 152: we should probably give people a small hint as to the motivation for this

seabbs (approved):

EpiAware/docs/src/examples/getting_started.jl
SamuelBrand1: I agree. In the back of my head, this should be a constructor function for `DelayObservations`.

For example, we could have a kwarg like `varprior_mean = ...` and its understood that this creates a Half-Normal prior for the variance component of the model with the desire mean.

SamuelBrand1: Made this an issue.

EpiAware/docs/src/examples/getting_started.jl
SamuelBrand1: I gave this a bit of write up in the `getting_started` note.

Issue #126, opened 2024-03-08
Clarity on use of Doc extension tools 


SamuelBrand1:
I thought we were using the extension tools to manage the construction of the args documentation etc as you had in the original docs PR (and the other similar extension tools). It sounds like this is still in the air though so perhaps we need another discussion issue? What I mean is that yes we have the template but I thought when writing docstrings we would be automating large parts of them using the extension tools to make them less error prone (l;like the arg types).

_Originally posted by @seabbs in https://github.com/CDCgov/Rt-without-renewal/pull/117#discussion_r1516594731_

SamuelBrand1:
My understanding is that we aim for [this](https://docs.sciml.ai/SciMLStyle/stable/#Documentation) for type and function docstrings.

The way we get there is using docstringextensions, but descriptions like `#Arguments` need to go in the DOCSTRING for the template. The types of the arguments are found by TYPEDSIGNATURES.

seabbs:
> The way we get there is using docstringextensions, but descriptions like #Arguments need to go in the DOCSTRING for the template. The types of the arguments are found by TYPEDSIGNATURES.

Yes, issue is about using these which wasn't in the PR that this comment was pulled out of.

seabbs:
Noting that the lack of clarity here is starting to lead to increasing amounts of technical debt as some features are being implemented without any documentation and most have different standards being used. We should aim to resolve this ASAP.

SamuelBrand1:
I think we have clarity now.

`DocStringExtensions.jl` provides annotations in the the template macro but not used elsewhere.

PR #128, opened 2024-03-08
Issue 53: First pass at differenced AR

### LINKED ISSUE #53: AR(1) latent process
As part of #3 we need an AR(1) process to be implemented. Ideally, this would also be general enough to support differencing and potentially AR(p) processes. 

Potential inspiration (doesn't look possible to depend directly?): 

- https://github.com/LAMPSPUC/StateSpaceModels.jl/blob/master/src/models/sarima.jl

seabbs:
This PR is ready for review and closes #53


seabbs:
Any feedback before I push on?

SamuelBrand1:
I'm still parsing this

EpiAware/src/abstract-functions.jl
SamuelBrand1, line 1: I think this can go in with abstract types? They aren't really abstract functions, they are the most general methods. I think of them as fallbacks.

EpiAware/src/latent-models.jl
```diff
@@ -26,3 +21,90 @@ end
...
+
+    """A distribution representing the prior distribution of the initial values."""
+    init_prior::Priors
+
+    """
+    The order of the AR process, determined by the length of `damp_prior`.
+    """
+    p::Int
+
+    function AR(damp_prior::Priors, var_prior::Distribution, init_prior::Priors)
```
SamuelBrand1, line 43: I think making at least one constructors with kwargs might be nice?

seabbs: Can you dummy out what you mean?

seabbs: See #133

seabbs:
This is ready for review. 

I have added unit tests but given the blocker of #126 I think we need to wait for that for the documentation (which as noted is a serious issue). 

I have also opened a new PR for the differencing (#134 which I will address after this) and made an issue for the lack of consistency for positional args (#133).

EpiAware/src/latentmodels/autoregressive.jl
SamuelBrand1, line 1: In #130 I'm changing to `std_prior` which I think is better (agreed f2f) so probably worth changing here.

EpiAware/src/latentmodels/autoregressive.jl
```diff
@@ -0,0 +1,53 @@
+struct AR <: AbstractLatentModel
```
SamuelBrand1, line 1: Its usually better to use concrete types for fields of structs. This looks like you should make it a [parameteric](https://docs.julialang.org/en/v1/manual/types/#Parametric-Types) typed struct.

SamuelBrand1 (dismissed):
This looks great; love the self-referencing approach to getting generic AR/diff AR!

I think the main `AR` struct needs to be parametric, thats the only change I recommend.

seabbs:
I've updated w/ all suggestions. 

Note I have used `@with_kw` from parameters.jl vs the base implementation as it allows for assets to be placed inside the inner constructor and so makes the implementation cleaner. I have also removed the specification of `p` as this doesn't work well when using defaults as it starts to conflict with the inner constructor.

seabbs:
@SamuelBrand1 pinging for review

SamuelBrand1:
> @SamuelBrand1 pinging for review

Reviewing this now

EpiAware/src/latentmodels/autoregressive.jl
```diff
@@ -0,0 +1,46 @@
+@doc raw"
+The autoregressive (AR) model struct.
+"
+@with_kw struct AR{D <: Priors, S <: Distribution, I <: Priors} <: AbstractLatentModel
```
SamuelBrand1, line 4: I like this use of `Parameters.@with_kw`. Happy to have `Parameters.jl` as a dep.

EpiAware/test/test_autoregressive.jl
```diff
@@ -0,0 +1,62 @@
...
+    std_prior = truncated(Normal(0.0, 0.05), 0.0, Inf)
+    init_prior = Normal()
+    ar_process = EpiAware.AR(damp_prior, std_prior, init_prior)
+
+    @test ar_process.damp_prior == damp_prior
+    @test ar_process.std_prior == std_prior
+    @test ar_process.init_prior == init_prior
+end
+
+@testitem "Test AR defaults" begin
```
SamuelBrand1, line 14: Could you test this for creating a vector length `damp` variable?

SamuelBrand1 (dismissed):
LGTM, only concern is that the method isn't robust to generating vector shaped `damp` or `ar_init`.

If this gets tested and works then I'm happy

EpiAware/test/test_autoregressive.jl
seabbs: Good spot your right it isn't. I will follow the pattern in #134 and reping for a review.

seabbs:
This should be updated now following the approach suggested in #134 and rebased on main. Note I had to copy `docstring`z templates everywhere due to submodules (a good downside we should have flagged) but I think the duplication isn't that bad as its something that gets touched rarely. 

Moving on to #134 now.

EpiAware/src/EpiLatentModels/utils.jl
```diff
@@ -0,0 +1,6 @@
+function _expand_dist(dist::Vector{D} where {D <: Distribution})
```
SamuelBrand1, line 1: Nice addition

SamuelBrand1 (approved):
I like this.

Issue #127, opened 2024-03-08
Nicer constructor for variance priors


SamuelBrand1:
This is a bit painful to expect people to write. Maybe we need an issue for a helper? Its quite low priority though obviously.

_Originally posted by @seabbs in https://github.com/CDCgov/Rt-without-renewal/pull/123#discussion_r1516558535_

SamuelBrand1:
- [ ] Make a method for giving _standard deviation_ a halfnormal prior.
- [ ] Make constructor functions for structs that use this.

PR #130, opened 2024-03-08
Create `_make_halfnormal_prior` function and change to priors on standard deviations

SamuelBrand1:
This doesn't finish #127 but is step one to doing it.

## Main change
Adds `_make_halfnormal_prior` function, intended for internal usage. This function takes in a desired prior mean for a _standard deviation_ and returns the halfnormal distribution with that mean. This is a special case of a half-t distribution, which is recommended as a prior for variance/std parameters.

_EDIT_: slightly increasing the scope of this PR to make explicit that the variance parameters having a prior are standard deviations, as per [_Gelman, Prior distributions for variance parameters for hierarchical models_](http://www.stat.columbia.edu/~gelman/research/published/taumain.pdf). Note that we use the half-normal special case of the half-t distribution used in that paper.


EpiAware/src/latentmodels/randomwalk.jl
```diff
@@ -10,9 +10,8 @@ end
 
 @model function generate_latent(latent_model::RandomWalk, n)
     ϵ_t ~ MvNormal(ones(n))
-    σ²_RW ~ latent_model.var_prior
+    σ_RW ~ latent_model.var_prior
```
seabbs, line 13: Does this now not need a name change?

EpiAware/src/inferencemethods/manypathfinder.jl
```diff
@@ -94,8 +94,8 @@ largest ELBO estimate.
 - `best_pfs::PathfinderResult`: Best pathfinder result by estimated ELBO.
 """
 function manypathfinder(mdl::DynamicPPL.Model, ndraws; nruns = 4,
-        maxiters = 50, max_tries = 100, kwargs...)
+        maxiters = 50, max_tries = 100, ndraws_elbo = 10, kwargs...)
```
seabbs, line 97: Unrelated changes here?

EpiAware/src/utils/dist.jl
```diff
@@ -150,8 +150,6 @@ function r_to_R(r, w::AbstractVector)
 end
 
 """
-    NegativeBinomialMeanClust(μ, α)
```
seabbs, line 153: More unrelated changes. This seems fine though

EpiAware/src/utils/dist.jl
```diff
@@ -167,3 +165,17 @@ function NegativeBinomialMeanClust(μ, α)
...
+
+# Arguments
+- `prior_mean::AbstractFloat`: The mean of the prior distribution.
+
+# Returns
+- `Truncated{Normal}`: The half-normal prior distribution.
+
+"""
+function _make_halfnormal_prior(prior_mean::AbstractFloat)
+    return truncated(Normal(0.0, prior_mean * sqrt(pi) / sqrt(2)), 0.0, Inf)::Distribution
```
seabbs, line 180: For interest adding the type (distribution) is optional here right? What informa your choice here? Clarity or an expectation of a performance difference?

seabbs (commented):
Nice. Just a few questions about various small changes and I think some checking of the use of var Vs std prior

EpiAware/src/latentmodels/randomwalk.jl
seabbs: Also note that #128 followed the lead of the random walk and used the variance vs the standard deviation. I would prefer the standard deviation formulation but think we should address in a single issue/PR change to try and maintain consistency.

SamuelBrand1: Yes that should have been an issue.

EpiAware/src/utils/dist.jl
SamuelBrand1: This is more a quick fingered typo.

EpiAware/src/inferencemethods/manypathfinder.jl
SamuelBrand1: Yep. I'm going to revert those.

EpiAware/src/utils/dist.jl
SamuelBrand1: Removed

EpiAware/src/latentmodels/randomwalk.jl
SamuelBrand1: But this stage I think increasing the scope of the PR makes more sense. I'll check for other places to switch.

seabbs (approved):
LGTM. #128 now needs to be updated to move from var to std.

PR #132, opened 2024-03-08
Move to using submodules for organisation

seabbs:
As the title this PR explores reorganising into submodules.

Done: 

- [x] Check that the getting started example works without changes in this PR.
- [x] Discuss how we feel about this approach
- [ ] Update based on discussion or close
- [x] Reogranise tests and update
- [x] Working out if the docstrings templates have to be duplicated or if there are other options.
- [x] Review if the changes to imports work or if there is a better way to control what is being imported.


SamuelBrand1:
This looks a neat move. My only requirement is that at the end-user level we have some one dep which a high level user can do `using ...` and do some basic models. I don't think you are breaking that (its more like `using OrdinaryDiffEq` which imports `SciMLBase` under the hood I think); but just laying it out.

seabbs:
The outstanding with this remains the inheritance of types from each other. i.e scan is based on AbstractEpiModel but can't find that method when called from another submodule

SamuelBrand1:
> @SamuelBrand1 we need to discuss this in detail ASAP. It either needs to be merged in and then PRs retargeted at its organisation or canned. If we keep making new feature PRs updating this will only become more painful.

Essentially, we are considering the upsides/downsides of: 
- Implementing submodules for namespace management in an increasingly complicated codebase. 
- Creating finer-grained control over method dispatch within "sub-module/sub-package" type code collections. The 

The canonical way of doing this in Julia is to create packages (e.g. how `SciMLBase` is used across `SciML` packages). 

For the situation of _fairly_ complicated code but for whatever reason we don't want to create more than one package/project then there is no consensus of the best way to do this, see for example this [post](https://discourse.julialang.org/t/ann-patmodules-jl-a-better-module-system-for-julia/52226/40) and general discussion. There is a related JuliaLang [PR](https://github.com/JuliaLang/julia/issues/4600) that also touches on this at the bottom of the thread.

I, personally, don't really like submodules because they still require you to organise the overarching module in a linear code fashion. I'm not against linear code organisation, but if we're doing that then submodules are only really useful to limit the namespace access of bits of the code. If that is the goal, I've just found [`FromFile.jl`](https://github.com/Roger-luo/FromFile.jl/tree/master) as an alternative.

seabbs:
> The canonical way of doing this in Julia is to create packages (e.g. how SciMLBase is used across SciML packages).

I think we agree that the sensible end state of this is that these proposed submodels are spun out into their own packages. I also think we agree that we don't want to target that end state now. 

I agree that `fromFile.jl` is nice and it sounds very clear it should be part of Julia (i.e the topic of that PR). I am not sure we want to use it here as it doesn't clearly provide a pathway to independent modules in the way that submodules do. 

The pushback to submodules seems to still be not wanting to handle the complexities of type inheritance across a modularised code base but it seems clear to me these problems will have to be face at somee point so why not now when so much of the leg work has been done here? If not now then it will just be later and require this work to be redone.

SamuelBrand1:
Ok, more concrete proposal:

We have `EpiAwareBase` and everything else in one submodule. 

My reason is that the modules aren't _really_ modules (IMO). For example, every example usage of `EpiModel` types needs an example `LatentModel` because it sort of forms one concept. I wouldn't want to modularise those abstractions apart from each other even though its very nice that we can create `Turing` models from each component.

seabbs:
> EpiModel types needs an example LatentModel because it sort of forms one concept

As currently implemented this isn't quite true. It happens that is how we are composing them but all they actually need is some vector.

>  I wouldn't want to modularise those abstractions apart from each other even though its very nice

Why?

> Any future extension package, lets imagine one that specialises in providing specific nowcasting tools, can import EpiAwareBase, do stuff and can be imported by EpiAware.

Yes I agree or not imported by `EpiAware` but use the other packages.

> Minimises code loading implied by splitting the code base up too much.

Could you touch more on this? 

> Lots of basic functionality lives together in EpiAware.

This seems like a big negative to me if thinking about modular resuse. If thinking about this as something that most people use from `using EpiAware` maybe it makes sense but I think the clarity and independence of declaring things that are not dependent on each as independent modules would still help as they can be separately developed and understood.

EpiAware/test/test_utilities.jl
```diff
@@ -201,13 +201,13 @@ end
     using Distributions, HypothesisTests
     @testset "Check distribution type" begin
         prior_mean = 10.0
-        prior_dist = EpiAware._make_halfnormal_prior(prior_mean)
+        prior_dist = EpiAware.EpiWareUtils._make_halfnormal_prior(prior_mean)
```
SamuelBrand1, line 204: EpiWareUtils -> EpiAwareUtils

EpiAware/test/test_utilities.jl
```diff
@@ -201,13 +201,13 @@ end
...
         @test typeof(prior_dist) <: Distribution
     end
 
     @testset "Check distribution properties" begin
         prior_mean = 2.0
-        prior_dist = EpiAware._make_halfnormal_prior(prior_mean)
+        prior_dist = EpiAware.EpiWareUtils._make_halfnormal_prior(prior_mean)
```
SamuelBrand1, line 210: EpiWareUtils -> EpiAwardUtils

SamuelBrand1:
> As currently implemented this isn't quite true. It happens that is how we are composing them but all they actually need is some vector.

But that is true of almost any function. They are kind of meaningless separated except to demonstrate how overworked you can make an RandomWalk definition.

EpiAware/docs/src/examples/getting_started.jl
```diff
@@ -96,7 +96,7 @@ Z_0 &\sim \mathcal{N}(0,1),\\
 
 # ╔═╡ 56ae496b-0094-460b-89cb-526627991717
 rwp = EpiAware.RandomWalk(Normal(),
-    EpiAware._make_halfnormal_prior(0.1))
+    EpiAware.EpiWareBase._make_halfnormal_prior(0.1))
```
SamuelBrand1, line 100: EpiAware.EpiObsModels

SamuelBrand1: Nope

EpiAware/src/EpiAwareUtils/EpiAwareUtils.jl
```diff
@@ -0,0 +1,22 @@
...
+
+"""
+Module for defining utility functions.
+"""
+
+using ..EpiAwareBase
+
+import DataFramesMeta: DataFrame, @rename!
+import Turing: Chains
+import Distributions: Distribution, cdf, Normal, truncated
```
SamuelBrand1, line 11: I think you need `Normal` here. I would just do using Distributions

EpiAware/src/EpiAwareUtils/prior-tools.jl
```diff
@@ -0,0 +1,13 @@
...
+# Arguments
+- `prior_mean::AbstractFloat`: The mean of the prior distribution.
+
+# Returns
+- `Truncated{Normal}`: The half-normal prior distribution.
+
+"""
+function _make_halfnormal_prior(prior_mean::AbstractFloat)
+    return truncated(Normal(0.0, prior_mean * sqrt(pi) / sqrt(2)), 0.0, Inf)
+end
```
SamuelBrand1, line 13: `Normal` and (I think) `truncated` are not available in EpiAwareUtils

seabbs:
> But that is true of almost any function. They are kind of meaningless separated except to demonstrate how overworked you can make an RandomWalk definition.

What if you use a SEIR model in `generate_latent_infs`?

seabbs:
> Where should `make_epi_aware` live? 

It could live in `EpiAware` or in `EpiAwareBase`. The advantage of keeping it out of base is that we wouldn't need to depend on Turing in the base module but the disadvantage is that if you want to compose a model that users it you need to also inherit everything from all the other submodules. Its a bit academic right now but worth thinking about I thik

seabbs:
@SamuelBrand1 this now works and is ready for a proper discussion.

SamuelBrand1:
> > But that is true of almost any function. They are kind of meaningless separated except to demonstrate how overworked you can make an RandomWalk definition.
> 
> What if you use a SEIR model in `generate_latent_infs`?

If we had an SEIR model with a random walk input, I wouldn't be using `EpiAware` because I'd want to import `InterpolatingAdjoint` from SciMlSensitivity and ODE solvers.

EpiAware/src/EpiInference/EpiInference.jl
```diff
@@ -0,0 +1,15 @@
+module EpiInference
+
+"""
+Module for defining inference methods.
+"""
+
+import Pathfinder: pathfinder, PathfinderResult
```
seabbs, line 7:
```suggestion
using Pathfinder: pathfinder, PathfinderResult
```

EpiAware/src/EpiInfModels/EpiInfModels.jl
```diff
@@ -0,0 +1,22 @@
+module EpiInfModels
+
+"""
+Module for defining epidemiological models.
+"""
+
+using ..EpiAwareBase
+
+import ..EpiAwareUtils: scan, create_discrete_pmf
```
seabbs, line 9:
```suggestion
using ..EpiAwareUtils: scan, create_discrete_pmf
```

EpiAware/src/EpiAwareUtils/EpiAwareUtils.jl
```diff
@@ -0,0 +1,22 @@
+module EpiAwareUtils
+
+"""
+Module for defining utility functions.
+"""
+
+using ..EpiAwareBase
+
+import DataFramesMeta: DataFrame, @rename!
```
seabbs, line 9:
```suggestion
using DataFramesMeta: DataFrame, @rename!
```

EpiAware/src/EpiAwareUtils/EpiAwareUtils.jl
```diff
@@ -0,0 +1,22 @@
...
+import Turing: Chains
```
seabbs, line 10:
```suggestion
using Turing: Chains
```

EpiAware/src/EpiAwareUtils/EpiAwareUtils.jl
seabbs, line 11:
```suggestion
using Distributions: Distribution, cdf, Normal, truncated
```

SamuelBrand1 (approved):
LGTM

seabbs:
? If we had an SEIR model with a random walk input, I wouldn't be using EpiAware because I'd want to import InterpolatingAdjoint from SciMlSensitivity and ODE solvers.

An you couldn't nest it?

PR #134, opened 2024-03-11
Add differenced AR models

seabbs:
As the title. This PR extends #128 to include differencing for any latent model. It is joint work from @SamuelBrand1 and me but I think @SamuelBrand1 is still objective enough to review.

Since we discussed this I have: 

- Standardised the implementation based on the current `AR()` implementation
- Added input checks
- Update tests to work with other changes. 

I think that covered everything we discussed.


SamuelBrand1:
Looking at this... I'm loving the self-referential definition of AR differencing function. That is neat.

seabbs:
If it works.. I am seeing it work and then not work in quite a strange way. Getting errors about thread safety.

seabbs:
@SamuelBrand1 FYI https://github.com/CDCgov/Rt-without-renewal/pull/134/commits/4a4f85e8dfec4b952137c66a42b409271190d13e this branch was pulled off of #issue53 and so should have been rebased or merged from that PR.

SamuelBrand1:
Why not just kill off branch `issue53`?

SamuelBrand1 (approved):
I really like this way of adding differenced models; really slick

Issue #137, opened 2024-03-11
Docstrings for all EpiModel types


SamuelBrand1:
Lets break this down:

- [ ] Finalize template, doc the `EpiAware` module and the abstract types and fall-back methods that dispatch on them as the most general case.
- [ ] doc the `utils`
- [ ] doc the `epimodels`
- [ ] doc the `latentmodels`
- [ ] doc the `observationmodels`
- [ ] doc the `inferencemethods`

_Originally posted by @SamuelBrand1 in https://github.com/CDCgov/Rt-without-renewal/issues/35#issuecomment-1985884776_

Issue #135, opened 2024-03-11
Make sure no doc strings have function/struct names.


SamuelBrand1:
More unrelated changes. This seems fine though

_Originally posted by @seabbs in https://github.com/CDCgov/Rt-without-renewal/pull/130#discussion_r1518567943_

SamuelBrand1:
These should be generated by `DocStringExtensions`

SamuelBrand1:
This looks done with last reorganistion

Issue #133, opened 2024-03-11
Make all constructors have named and not positional arguments.


seabbs:
even 

```julia
function AR(;damp_prior::Priors, var_prior::Distribution, init_prior::Priors)
```

here so the constructor is not based on positional args (I realise that this doesn't exist for all structs yet)

_Originally posted by @SamuelBrand1 in https://github.com/CDCgov/Rt-without-renewal/pull/128#discussion_r1517922322_

seabbs:
Looking through the code there seems to be a lot of variance on this. I think we should agree a clear approach, document it, implement it everywhere, and expect future methods to follow it.

SamuelBrand1:
Breaking this down:

A list of struct types with available constructors with named args:

- [ ] `EpiModel`s and `EpiData`
- [ ] `LatentModel`s
- [ ] `ObservationModel`s

Issue #131, opened 2024-03-08
Revisit handling of default priors


seabbs:
Default priors are currently handled by small helper functions that have custom names. This seems surprisingly clunky given the rest of the code base. 

Whilst I agree default priors in structs could be dangerous in the hands of some users I also think providing opinionated priors can be very helpful and also guide practice. 

Personally I would vote for default priors in the structs but open to other suggestions?

SamuelBrand1:
My view is that if we can close #133 in such a way that all priors have defaults for their constructor; then this is closed too.

seabbs:
I think this is now resolved? (checked and yes)

PR #138, opened 2024-03-11
Docstrings for all epimodel types and generate_latent_infs plus named arguments for constructors

SamuelBrand1:
This is a draft PR for on going work on documenting `AbstractEpiModel` subtypes and methods that dispatch on them.

## Current focus and decisions:

- As recommended [here](https://docs.sciml.ai/SciMLStyle/stable/#documentation) I am focusing documentation at the highest/most abstract level. This means that I'm going to give the example usages and most extensive documentation to `generate_latent_infs` dispatch on generic `AbstractEpiModel`.
- I don't think the METHODLIST was useful in our doc template and has been removed.
- Example usage of a `DirectInfections` goes beyond just constructing the struct, but also its main usage with `generate_latent_infs`.
- Example usage is not a doc test; it has a complicated output that works ok with a spread of `testitem`s. However, a next step could be to combine as per discussed in #48. _I've readvised this... simple doctests will be included_.

This is aimed at addressing #137 which is part of #35 and #97 . It sets my current expectation of docstrings #135 and #126 .

## Extra activity

- When it is absolutely straightforward to create a default constructor with named arguments using `@kwdef` I am doing it, because this also makes example usage easier. This moves towards addressing #133 , but won't close this issue.
- Advantage of above is that default priors are added directly aiming to address #131 .


SamuelBrand1:
I've done everything in the opening description _except_ for the example usage of the EpiModels and `generate_latent_infs` is **not a doctest**.

The reason for this is that the pattern of doc testing against an expected result didn't seem to meet what we want from the examples and might lead into complexity around different random seed implementation on different OS etc. An alternative approach which just checks the examples run didn't seem to do anything that our testing CI doesn't do anyway.

Happy to get pushed back on this.

EpiAware/src/epimodels/directinfections.jl
```diff
@@ -1,9 +1,115 @@
...
+```
+
+where ``g`` is a transformation function and the unconstrained initial infections
+    ``\hat{I}_0`` are sampled from a prior distribution, `initialisation_prior` which must
+be supplied to the `DirectInfections` constructor. The default `initialisation_prior` is
+the standard Normal `Distributions.Normal()`.
+
+## Constructor
+
+`DirectInfections` can be constructed by passing an `EpiData` object and and subtype of
```
seabbs, line 20: repeated and

seabbs, line 20:
```suggestion
`DirectInfections` can be constructed by passing an `EpiData` object and subtype of
```

seabbs (approved):
This is nice. It doesn't look like you have implemented field or return information for functions or made use of `DocStringExtensions`. I think we need at least a clear return field but perhaps in another issue.

Issue #97, opened 2024-02-29
Add examples to documentation


seabbs:
As started in #94 we should move the example code into the documentation.

PR #140, opened 2024-03-12
Docstrings for all latentmodel types and generate_latent plus named arguments for constructors

SamuelBrand1:
This is a PR for documenting `AbstractLatentModel` subtypes and methods that dispatch on them.

## Current focus and decisions:

Follows #138 in decisions.

This PR is aimed at addressing #139 which is part of #35 and #97 .

## Extra activity

- When it is absolutely straightforward to create a default constructor with named arguments using `@kwdef` I am doing it, because this also makes example usage easier. This moves towards addressing #133 , but won't close this issue.
- Advantage of above is that default priors are added directly aiming to address #131 .

### DIFF
--- a/EpiAware/docs/src/examples/getting_started.jl
+++ b/EpiAware/docs/src/examples/getting_started.jl
@@ -95,8 +95,9 @@ Z_0 &\sim \mathcal{N}(0,1),\\
 "
 
 # ╔═╡ 56ae496b-0094-460b-89cb-526627991717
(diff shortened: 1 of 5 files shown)


seabbs:
This PR has some major conflicts. Happy to resolve and then review.

seabbs (approved):
This looks good to me. @SamuelBrand1 can you check you are also happy?

PR #147, opened 2024-03-15
Change `pos_shift` to be a property of Observation model

### LINKED ISSUE #146: `pos_shift` should be a property of `ObservationModel`
At the moment, we still have the up-jitter quantity `pos_shift` as an argument to the whole inference process. This should be part of specifying `ObservationModel`.

SamuelBrand1:
As mentioned in #146 . I think this aligns much better with the modular structure.

Close #146


seabbs (approved):
LGTM

Issue #122, opened 2024-03-07
Overall structure of EpiAware: proposing SciML style


SamuelBrand1:
The goal of this issue is to define the overall structure of `EpiAware` API; and to suggest an opinionated approach.

I think we should try and match the `SciML` pattern of:

- [x] #145
- [x] #151
- [x] #156

### EpiAware problem object
Propose calling the type `EpiAwareProblem`.
This should contain:
  - A latent model `<: AbstractLatentModel`
  - An epimodel `<: AbstractEpiModel`
  - An observation model `<: AbstractObservationModel`. _Any actual data we have should be a random variable in the Observation model and we can use variable passing or `DynamicPPL` conditioning_. Discussed this briefly with @seabbs off-line.

### Solution method problem object
Propose calling the type `EpiAwareInferenceMethod`.
This should contain:
- Initialisation method
- sampler type e.g. NUTS

### Our `solve` function
I think we should call this `epi_aware` and have signature

```julia
epi_prob = EpiAwareProblem(...)
inference_method = EpiAwareInferenceMethod(...)
sol = epi_aware(epi_prob, inference_method; kwargs...)
```

seabbs:
> I think we should try and match the SciML pattern of:

Strongly agree.

>  we can use variable passing or DynamicPPL conditioning. 

Agree and vote for conditioning as it seems very nice. Only issue I see is knowing precisely what to condition on and that its working as expected.

> An observation model <: AbstractObservationModel. 

To generalise this do we want hypothetical support for n `AbstractObservationModel`s or would we instead create a nested type to deal with this?

SamuelBrand1:
> To generalise this do we want hypothetical support for n AbstractObservationModels or would we instead create a nested type to deal with this?

My vote would be for a nested type. 

It would be nice to have a Turing (sub)model for observation where you could pass in:
- A times series of true/latent incidence
- parameter priors

And then you could sample observations. Then everything from a trivial "You just observes incidence as it occurs" through to a more complicated delayed obs and reporting model designed for nowcasting could all be accepted.

SamuelBrand1:
Yeah I think that is fine, I don't think we strong type the data anywhere. If we have an observation model that has something like 

`y_t ~ LogNormal(...)` 

in there it _should_ just work

seabbs:
[...]

Issue #119, opened 2024-03-06
Observed data API


SamuelBrand1:
## Goal

The goal of this issue is to lockdown how we want `EpiAware` to ingest observed data, especially in light of #107 .

## Current API

The data `y_t` is an argument to `make_epi_aware` constructor. If its `y_t = missing` then `EpiAware` generates the data from the defined generative model, otherwise it is treated as conditioning the model.

## Possible other option: upside and downside

The other `Turing` approach would be to not have `y_t` as an argument. Conditioning on observed data is then done using `condition` or the syntax sugar `|`, 

eg

```julia
inference_mdl = condition(generative_mdl, (y_t = my_case_data,))
```

or

```julia
inference_mdl = generative_mdl | (y_t = my_case_data,)
```

#### Upsides of this approach
- Flexible
- Close to the mathematical conception of the model.

#### Downsides of this approach
- Not specifying a clear way of proving data could be:
  - Confusing to users
  - Relies on knowing variable names inside the model.

seabbs:
> Conditioning on observed data is then done using condition or the syntax sugar |,

Yes I like this approach. Is there anyway to make it easy and discoverable so that uses can know what they can condition and to make sure that condition errors if they pick something that doesn't exist (maybe both of these issues are on `dynamicppl` or `turing`.

seabbs:
With the introduction of `StackObservationModel` in #249 we have an opinionated choice to support observations called `y_t` in the simple case and as a `NamedTuple` in the complex case which then gives `name.y_t`. I think this is a clear enough pattern and it seems robust.

Do we think there is more to do here or shall we close in favour of a new issue focussing on how we pass metadata?

PR #148, opened 2024-03-15
EpiAware Problem object

### LINKED ISSUE #145: Problem object

SamuelBrand1:
As discussed in #122, we want an `EpiAwareProblem` struct which defines the inference and/or generative modelling problem implied by case data and the choices in the underlying epi modelling.

Where I'm lacking a bit of clarity is where we include observed case data. At the moment, we are passing case data to a Turing model which implements the observation model behaviour. If we want to go on with this then case data should be a field of `EpiAwareProblem`. 

However, the alternative as discussed in #119 is to have observed case data as an intrinsic part of the Observation model through conditioning. In that case, this PR is nearly done already (but we'll need a bit more engineering on the observation model).

TBD:

- [x] Some testing for `EpiAwareProblem`.
- [x] ~~Add observational data field if we want to go that way.~~

Closes #145


seabbs (commented):
Yes this is nice and I think I agree that we want to use conditioning and so don't need to carry around the observed data.

The only thing I might need clarity on is what we are conditioning on. Is it the output of `make_epi_aware` or of the observation submodule?

SamuelBrand1:
> The only thing I might need clarity on is what we are conditioning on. Is it the output of `make_epi_aware` or of the observation submodule?

Either works, but my leaning is towards conditioning on the observation part and supplying the conditioned observation model to `EpiAwareProblem`.

The reason I lean that way is because that can have utility in isolation:
- We can test/check the `logpdf` of the conditioned observation model for any supplied array of unobserved infections.
- We can test/check partially missing data support rather than in the the full `make_epi_aware` output.

seabbs:
> The reason I lean that way is because that can have utility in isolation:

Can't we always condition on the lower levels modules whenever we want and so either approach would allow you to do this? The thing you lose I think is that you can't updating the conditioning on `make_epi_aware` and so need to respecify the problem w/ new data? 

> I envision this as the next step; this PR is just adding the EpiAwareProblem struct which is compatible with either approach.

That sounds sensible

seabbs:
Is this ready for a review?

SamuelBrand1:
> Can't we always condition on the lower levels modules whenever we want and so either approach would allow you to do this? The thing you lose I think is that you can't updating the conditioning on `make_epi_aware` and so need to respecify the problem w/ new data?

Yeah, thats a reasonable point.

seabbs:
The test failure looks like a stochastic issue with pathfinder maybe indicating those tests are a bit loose. Will kick it off again

SamuelBrand1:
> The test failure looks like a stochastic issue with pathfinder maybe indicating those tests are a bit loose. Will kick it off again

Yes. I tracked it down to

```julia
@model function bad_model()
            x ~ Normal(0, 1)
            return sqrt(x) #<-fails
        end
```

It supposed to not be able to optimize for that model!

seabbs (approved):
This looks good to me.

SamuelBrand1:
Ok now the merged in test failed.

> Testing DiffLatentModel process: Test Failed at /Users/runner/work/Rt-without-renewal/Rt-without-renewal/test/test_difflatentmodel.jl:67
  Expression: ks_test_pval_day2 > 1.0e-6
   Evaluated: 4.604227577138768e-8 > 1.0e-6

seabbs:
I've kicked them off again. 

Regarding the test failure - shall we resolve in a new issue?

SamuelBrand1:
> I've kicked them off again.
> 
> Regarding the test failure - shall we resolve in a new issue?

The test failure is because the test is broken.

The problem here is that a two fold `cumsum` on the random vector 

```math
X = [X_1, X_2] \to_{\Delta^{(-2)}} [X_1,  X_2 + 2 X_1]
```

But our test is treating it as 

```math
[X_1,  X_2 +  X_1 + \tilde{X}_1]
```

Where $\tilde{X}_1$ is an independent copy of $X_1$. The distribution is wrong, so the test was failing as it should.

PR #149, opened 2024-03-15
Add Reexport.jl

seabbs:
As the title this PR adds `Reexport.jl` which allows us to use a macro to pull all exported functions from submodules into the main module as exports (i.e. we don't need to duplicate exporting them).


SamuelBrand1:
Good move.

SamuelBrand1 (approved):
Like this.

PR #155, opened 2024-03-15
Structs to define inference/generative methods

### LINKED ISSUE #151: Solution method object

SamuelBrand1:
This draft PR is about defining a structure for defining the inference/generative method.

The approach I've taken involves defining an `EpiAwareMethod` struct (as well as creating an abstract type hierarchy in `EpiAwareBase`).

The `EpiAwareMethod` defines a sequence of optimization methods (defined as either MAP or VI) followed by a single sampling method. The concept here is that the optimizations methods pass results to the next one sequentially before supplying initialization information to a sampler (e.g. NUTS).

I have added a `ManyPathfinderMethod` struct into `EpiInference` to give an example; however, I'm saying the scope of this PR is structure rather than the implementation.

Closes #151


SamuelBrand1:
I think this is a good structure to scaffold implementation methods onto.

Thoughts @seabbs ?

EpiAware/src/EpiAwareBase/EpiAwareBase.jl
```diff
@@ -6,9 +6,17 @@ Module for defining abstract epidemiological types.
...
-       AbstractObservationModel, AbstractEpiAwareProblem, generate_latent,
-       generate_latent_infs, generate_observations
+#Export models
+export AbstractModel, AbstractEpiModel, AbstractLatentModel, AbstractObservationModel
+
+#Export problems
+export AbstractEpiAwareProblem
+
+#Export inference methods
+export AbstractEpiAwareMethod, AbstractEpiAwareOptMethod, AbstractEpiAwareSamplingMethod
```
seabbs, line 16: Are we happy that he naming is consistent w/ these? The others don't have EpiAware nested within their name.

EpiAware/src/EpiInference/EpiInference.jl
```diff
@@ -1,16 +1,21 @@
-module EpiInference
 
 """
 Module for defining inference methods.
 """
+module EpiInference
 
+using ..EpiAwareBase: AbstractEpiAwareMethod, AbstractEpiAwareOptMethod,
+                      AbstractEpiAwareSamplingMethod
```
seabbs, line 8: new line needed

EpiAware/src/EpiInference/EpiInference.jl
```diff
@@ -1,16 +1,21 @@
...
 using Pathfinder: pathfinder, PathfinderResult
 
 using DynamicPPL, DocStringExtensions
 
+export AbstractNUTSMethod
```
seabbs, line 13: Logic for moving to multiple line vs single line?

seabbs, line 13: Logic for moving to multiple line vs single line?

EpiAware/src/EpiInference/manypathfinder.jl
```diff
@@ -1,3 +1,15 @@
+"""
+A variational inference method that runs `manypathfinder`.
+"""
+@kwdef struct ManyPathfinderMethod <: AbstractEpiAwareOptMethod
```
seabbs, line 4: Do we want to have method in the name or call it ManyPathfinder (I know I know)

EpiAware/src/epiawaremethods/epiawaremethod.jl
```diff
@@ -0,0 +1,11 @@
+"""
```
seabbs, line 1: not sure we need this in its own folder? I think everything at the root or of src or in its own submodule makes sense as a pattern? Or do you think there is a good case for more organisation within overarching EpiAware? I agree it makes sense for this functionality to in the sample place as make_epi_aware

seabbs (commented):
Actual contents looks good. Just some small comments about the naming structure and folders etc (so pretty exciting).

EpiAware/src/EpiAwareBase/EpiAwareBase.jl
SamuelBrand1: I'm not 100% happy, and open to a suggestion here.

Maybe, `AbstractEpiAwareProblem` but others are `AbstractMethod`, `AbstractEpiAwareSamplingMethod` etc etc?

EpiAware/src/EpiInference/EpiInference.jl
SamuelBrand1: I've got a vague preference for moving export to groups depending on the object being exported e.g. function, Struct, abstract type etc.

This would seem to make the code easier to read, but its not a style I've seen strongly advocated anywhere and we'd have to refactor other bits of the code base to do it. Thoughts?

I'm also ok with going to single line exports.

EpiAware/src/EpiInference/manypathfinder.jl
SamuelBrand1: I think either have `Method` postfix or have `Opt` postfix which I'm using to indicate optimization methods (including VI).

EpiAware/src/epiawaremethods/epiawaremethod.jl
SamuelBrand1: Yeah ok, happy to move into same place as `make_epi_aware`

EpiAware/src/EpiAwareBase/EpiAwareBase.jl
seabbs: Yeah so `AbstractEpiAwareMethod` and everything else follows the other naming scheme?

EpiAware/src/EpiInference/EpiInference.jl
seabbs: I like it. Lets use it here and add an issue to do elsewhere.

EpiAware/src/EpiInference/manypathfinder.jl
seabbs: And that isn't duplicating the type-information? I fine with doing this so more asking about good practice as in R that would be probably considered duplicated info.

EpiAware/src/epiawaremethods/epiawaremethod.jl
seabbs: and your definitely not seeing it as part of `EpiInference`?

EpiAware/src/EpiAwareBase/EpiAwareBase.jl
SamuelBrand1: E.g. 

`AbstractOptMethod <: AbstractEpiAwareMethod`

etc?

seabbs: or just do epi for all of them?

SamuelBrand1: Lets do that.

EpiAware/src/EpiInference/manypathfinder.jl
SamuelBrand1: Thats a good point.

EpiAware/src/epiawaremethods/epiawaremethod.jl
SamuelBrand1: I think having the overall infrastructure in `EpiAware` but the individual method structs and implementations in EpiInference make sense.

EpiInference is then stand alone, and `EpiAware` is a judgement on a good way to weave the various submodules togethers.

seabbs: > individual method structs and implementations in EpiInference make sense.

I agree this makes sense.

With the current implementation there is nothing `EpiAware` specific about `EpiMethod` is there its just a way of chaining inference approaches together and making sure they will work? 

Given that it seems like you could imagine using them on other inputs that aren't from `EpiAware` and wouldn't it make more sense for it to live in `EpiInference`? i.ee it isn't weaving multiple submodules together all the parts of `EpiMethod` are from `EpiInference`. I totally agree that when it comes time to actually use a model object build out of other parts that is something that lives at the top level.

SamuelBrand1: I've moved it into `EpiInference`

seabbs (approved):
LGTM

PR #159, opened 2024-03-19
add Suppressor to Test env and suppress warnings

### LINKED ISSUE #154: Silence pathfinder.jl in tests
We have near silent tests apart from the info statements in our pathfinder wrappers. It would be nice to silence these to make the test output easier to read. From a skim it seems like using `Logging.jl` might be the way to do this.

SamuelBrand1:
Does what it say in title to address #154 .

Closes #154


seabbs (approved):
Nice

Issue #161, opened 2024-03-20
Transformed data concept


seabbs:
Our current design goal is that the model specification for each module doesn't require information about the overall dimensions of the problem or observation data. 

For many problems this limitation is fine but for some things that could be computed just once and used as data this leads to inefficiency. I had hoped that `Turing` would have an abstraction to deal with this but it doesn't appear to be the case. 

A good example of this issue is `LatentDelay` which needs to construct a delay kernel. At the moment this is effectively data once the time index is known as we are assuming the latent delay is fixed. However, due to our current design this needs to sit within `generate_observations` and so is called many times (I believe) 

Any potential solution to this issue needs to take into account future extensions. For this specific example these are: Uncertain delay summary parameters, time-varying delays, hierarchy in uncertain delays.

As workarounds, I have tried using `Memoization.jl` which results in long fit times (I think due to lack of support for reversediff with multi-threading but unsure). Another option is to use a generator function for the model macro but though I haven't tried it I think this solution doesn't work with submodels. 

We could also introduce another layer of abstraction and have structs define new trusts using time and dimension index information when we combine the problem. This could be quite tricky to do potentially and would require more infrastructure but failing any other solution seems like it would definitely work with little loss of generality.

Before taking any concrete action we should also follow up to see if it is indeed the case that `Turing` isn't able to perform a simplification here or that if it currently isn't there isn't a feature planned in the near term that would do this.

SamuelBrand1:
> We could also introduce another layer of abstraction and have structs define new trusts using time and dimension index information when we combine the problem. This could be quite tricky to do potentially and would require more infrastructure but failing any other solution seems like it would definitely work with little loss of generality.

I think this is the way we'll have to proceed.

SamuelBrand1:
Thinking about this a bit more.

1. I think that time and dimension have to be considered separately, for the same meta reasons (essentially the Markov property in time but not necessarily other dimensions) they effectively are in (say) `ODEProblem`.
2. Leaving _time_ out of model components then using `tspan` from `EpiProblem` to fill this in is pretty straightforward.
3. Leaving _other dimensions_ out of model components is a bit too ambitious. My earlier thinking was that we could abstract this away by using rules to standardize broadcasting/reduce across different dimensions; but this could easily get more complicated than just defining the dimensions for 
  - Latent processes
  - Infection processes
  - Observation processes
4. Defaults above are obviously implied dimension 1

seabbs:
[...]

Issue #160, opened 2024-03-20
General missing data support to all observations models


seabbs:
When dealing with #80 I have split delays and the observation model into separate structs. This still leaves a lot of boilerplate in the Negative binomial model that is needed in order to support missing data (even more if wanting to vectorise when we can) and more will be needed to broadcast single cluster factor priors to vectors of the same length as the observation (so that we can naturally support more complex cluster factor models without changing the observation model. 

I think it makes sense to construct an internal struct and then dispatch on this to methods that are either vectorised or use a for loop and then call a supplied function.

I have done a bit of scaffold work on that here but paused to finish up #80. 

```julia

struct ObswithMissing{F <: Function, I <: AbstractVector{<:Union{Real, Missing}}, L <: AbstractVector{<:Real}} <: AbstractObservationModel
    model::F
    o_t::I
    O_t:L
    params::NamedTuple

    function ObswithMissing(model::Function, o_t::AbstractVector{<:Union{Real, Missing}}, params::AbstractVector)
        new_params = NamedTuple{(:params,), Tuple{AbstractVector}}((params,))
        for i in eachindex(params)
            if length(params[i]) == 1
                new_params[i] = fill(params[i], length(obs))
            end
        end
        return new{typeof(model), typeof(obs), typeof(new_params)}(model, obs, new_params)
    end

    # Custom constructor for entirely missing obs
    function ObswithMissing(model::Function, obs::AbstractVector, params::AbstractVector)
        obs = Vector{eltype(Y_t)}(undef, length(Y_t))
        return ObswithMissing(model, obs, params)
    end
end

@model function EpiAwareBase.generate_observations(obs_model::NegativeBinomialError,
    y_t,
    Y_t)
cluster_factor ~ obs_model.cluster_factor_prior

if ismissing(y_t)
y_t = Vector{Int}(undef, length(Y_t))
end

return generate_observations(obs_model, y_t, Y_t, cluster_factor)
end

# Vectorized implementation for when there are no missing values
@model function generate_observations(,
y_t::Vector{<:Integer},
Y_t::Vector{<:Real},
cluster_factor)
y_t .~ NegativeBinomialMeanClust.(Y_t .+ obs_model.pos_shift, cluster_factor^2)
return y_t, (; cluster_factor,)
end

# Loop-based implementation for when there are missing values
@model function generate_observations(obs_model::NegativeBinomialError,
y_t::Vector{Union{Missing, Integer}},
[...]

PR #162, opened 2024-03-20
Issue 80: Decompose observation model

### LINKED ISSUE #80: Consider renaming DelayObservations and family
As the title. This is currently hard coded to a negative binomial. We don't plan to change this out in our code but it would be nice to make it clear that we could if we wanted to and to make that process fairly straightforward. 

Options I see here are 1. A single name change or 2. A more complete update to either split delays from the obs model or introduce some further nesting.

https://github.com/CDCgov/Rt-without-renewal/blob/8b9edf95ae8a513f5af2f555b4895078122e5304/EpiAware/src/observation-processes.jl#L7

@SamuelBrand1 in particular thoughts on which of these to consider?

seabbs:
This PR closes #80 by decomposing the current observation model into two structs: `LatentDelay` and `NegativeBinomialError()`. It also enforces our design goal of not having structs know about the specific index of the problem and so moves the creation of the sparse delay kernel into the `generate_obserations` call. Finally it updates all tests and does a minor reorganise of them so that test are by submodule. 

I have made the following issues based on work in this PR: 

- #160 
- #161


EpiAware/docs/src/examples/getting_started.jl
SamuelBrand1, line 1: We keep swapping Pluto version, I think it because I'm using my 1.10 env version and calling Pluto from command line.

I think we should `up` the docs env and then stick with whatever `Pluto` version that is

seabbs: ah yeah makese sense. i have been going into `EpiAware/docs` and running it which is a bit boring. It might be nice to include a little shell script in the `docs` repo to spin up the right `Pluto` environment as this would totally prevent this going forward?

seabbs: Making this a new issue.

EpiAware/src/EpiObsModels/LatentDelay.jl
```diff
@@ -0,0 +1,60 @@
...
+- `y_t`: The current observations.
+- `I_t`: The current infection indicator.
+
+## Returns
+- `y_t`: The updated observations.
+- `obs_aux`: Additional observation-related variables.
+
+"
+@model function EpiAwareBase.generate_observations(obs_model::LatentDelay, y_t, Y_t)
+    kernel = generate_observation_kernel(obs_model.pmf, length(Y_t))
```
SamuelBrand1, line 54: We've discussed this a bit f2f, as per the top:

> It also enforces our design goal of not having structs know about the specific index of the problem and so moves the creation of the sparse delay kernel into the generate_obserations call.

I think its reasonable to have a pattern where `obs_model` is expected to know how to construct the kernel if its dispatching on a method that defines a behaviour where it doesn't change with different parameters.

e.g. 

```julia
generate_observation_kernel(obs_model::StaticDelay) #<-- just gets a pre-calculated kernel
```

This is probably a new issue?

EpiAware/docs/src/examples/getting_started.jl
seabbs: Seee #163

SamuelBrand1 (approved):
I like the split between latent cases and the observation, and looks good to me.

Most of the comments are towards further development.

EpiAware/docs/src/examples/getting_started.jl
```diff
@@ -98,7 +98,7 @@ Z_0 &\sim \mathcal{N}(0,1),\\
 # ╔═╡ 56ae496b-0094-460b-89cb-526627991717
 rwp = EpiAware.RandomWalk(
     init_prior = Normal(),
-    std_prior = EpiAware.EpiLatentModels._make_halfnormal_prior(0.1))
+    std_prior = EpiAware.EpiLatentModels.HalfNormal(0.1))
```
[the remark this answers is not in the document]
seabbs: Yeah I thought about addressing that here and agree it could make sense. Agree its a new issue.

EpiAware/src/EpiAwareUtils/priors.jl
```diff
@@ -0,0 +1,13 @@
...
+Create a half-normal prior distribution with the specified mean.
+
+# Arguments
+- `prior_mean::AbstractFloat`: The mean of the prior distribution.
+
+# Returns
+- `Truncated{Normal}`: The half-normal prior distribution.
+
+"""
+function HalfNormal(prior_mean::AbstractFloat)
```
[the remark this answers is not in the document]
seabbs: Resolving as I suggest we solve in the issue for making this a true Distribution.

EpiAware/src/EpiObsModels/LatentDelay.jl
seabbs: Yes good idea I think this might be a new issue related to #161 (i.e resolve that and then this is a specific instance?)

seabbs: Are you happy to make one as you might be best placed to flesh out above?

EpiAware/docs/src/examples/getting_started.jl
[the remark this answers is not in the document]
seabbs: Issue for this #164

EpiAware/src/EpiObsModels/LatentDelay.jl
seabbs: Once we have an issue for this and this convo is resolved this PR will merge if we are happy?

PR #166, opened 2024-03-21
standardise pluto workflow

### LINKED ISSUE #163: Standardise Pluto workflow
We keep swapping Pluto version, I think it because I'm using my 1.10 env version and calling Pluto from command line.

I think we should `up` the docs env and then stick with whatever `Pluto` version that is

_Originally posted by @SamuelBrand1 in https://github.com/CDCgov/Rt-without-renewal/pull/162#discussion_r1533645387_
            
My suggested solution to this is that we make a entry point shell script for pluto in the `docs` folder (or elsewhere) that developers are expected to use when editing the docs. 

As suggested by @SamuelBrand1 in this issue we should also make sure the `docs` environment is updated as part of this issue resolution.

SamuelBrand1:
Does what it says above following #163 .

Closes #163

### DIFF
--- a/EpiAware/README.md
+++ b/EpiAware/README.md
@@ -77,3 +77,7 @@ E-->|sample...NUTS...| G
 G-.->H
 H-.->I
 ```
+
+## Pluto scripts
+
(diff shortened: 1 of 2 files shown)


seabbs (approved):
❤️

PR #168, opened 2024-03-25
Issue 164: Promote HalfNormal to a true Distributions.jl method

### LINKED ISSUE #164: Make HalfNormal a true Distributions.jl distribution.
I think if we are doing this then we should extend Distributions and make a new distribution rather than use distribution style ontop of a function call?

Maybe spin this as a new issue.

_Originally posted by @SamuelBrand1 in https://github.com/CDCgov/Rt-without-renewal/pull/162#discussion_r1533647029_

seabbs:
This PR closes #164 by implementing the required methods for `HalfNormal` to be a Distributions.jl distribution. It also renames `create_discretised_pmf` to `censored_pmf` to be more in line with `Distributions.jl` naming.

_Note that it does not implement all the optional methods_


SamuelBrand1:
Looks really good

SamuelBrand1 (approved):
Very neat.

PR #169, opened 2024-03-25
156 solve like function which dispatches on epiawareproblem and epiawaremethod

### LINKED ISSUE #120: Passing initialisations to NUTS/inference approach
Can we have a helper function for this or make it part of the output of best_pf?  I don't love reading this or the thought of having to repeatedly write it down.

_Originally posted by @seabbs in https://github.com/CDCgov/Rt-without-renewal/pull/117#discussion_r1516011356_

### LINKED ISSUE #156: `solve`- like function which dispatches on `EpiAwareProblem` and `EpiAwareMethod`

SamuelBrand1:
A draft PR for an `apply_method` method.

## Methods
The basic idea is that each behaviour of a `method`, which means either a sampling or optimization technique, has an `_apply_method` dispatch which implements this.

An important method is `EpiMethod` implies doing a sequence of optimizations passing infomation to each next level, then finally to a sampling method `NUTSampler`.

## `apply_method`

`apply_method` has two methods:
1. ingests a `Turing` model object and applies the epi method.
2. Ingests an `EpiProblem` and a data object, as well parameters to fix and/or condition on, creates an appropriate model and then apply the epi method.

The output of `apply_method` is a `NamedTuple` which returns `samples` as well as `gens`, that is generated quantities if that makes sense for the returned object from `_apply_method`, and `model` which is the model (after any conditioning or fixing of parameters).

## TBD

- Naming?
- Return objects? 
- More unit tests


EpiAware/src/EpiInference/EpiInference.jl
```diff
@@ -1,17 +1,18 @@
...
 """
 module EpiInference
 
-using ..EpiAwareBase: AbstractEpiMethod, AbstractEpiOptMethod,
-                      AbstractEpiSamplingMethod
-using Pathfinder: pathfinder, PathfinderResult
-
-using DynamicPPL, DocStringExtensions
+using ..EpiAwareBase: AbstractEpiMethod, AbstractEpiOptMethod, AbstractEpiSamplingMethod
+import ..EpiAwareBase: _apply_method
```
seabbs, line 7: why are you using both here? I thought we decided only to use using?

seabbs (commented):
In general I think this all looks good. Can look through in more detail or happy to wait until it goes from draft to full.

EpiAware/src/EpiInference/EpiInference.jl
SamuelBrand1: `import`ing a function means that each method on `_apply_method` is associated with that function. `using` only does that if you prefix the method definition with the module. I'm happy to do that instead if you prefer.

Here I wanted `_apply_method` in `EpiAwareBase` because its used differently in the inference submodule and the infection submodule.

SamuelBrand1:
Its failing the 
> Testing NegativeBinomialError against theoretical properties` 
test on CI but not locally. Any thoughts?

seabbs:
I think this is ready for review/merging now. 

Whilst there are definitely some edge case issues and potentially testing gaps I think we would be best served in making new issues for those based on review and only fixing large scale issues before merging (as this is already very large).

seabbs:
@SamuelBrand1 this is ready for review I think. 

There is an outstanding issue with the manypathfinder into NUTs in the getting started example that I think is unrelated but I am really not clear on. 

We also need to open some new issues to improve areas of this but as long as we do that I think we should move to merge asap as this is now very large, blocking, and complicated (none of which is ideal).

SamuelBrand1:
OK. I'm looking at this and the getting_started example

SamuelBrand1 (commented):
First, quick takeaway. I like the restructure in the files.

EpiAware/src/EpiAwareBase/EpiAwareObservables.jl
```diff
@@ -0,0 +1,32 @@
...
+- `data`: The data used for the observables.
+- `samples`: Samples from the posterior distribution.
+- `generated`: The generated observables.
+"
+struct EpiAwareObservables
+    model::Any
+    data::Any
+    samples::Any
+    generated::Any
+end
```
SamuelBrand1, line 15: I like this.

EpiAware/src/EpiAwareBase/types.jl
```diff
@@ -1,17 +1,27 @@
 abstract type AbstractModel end
 
 """
-The abstract supertype for all structs that define a model for generating unobserved/latent
-infections.
+The abstract supertype for all structs that define a model for generating
+unobserved/latent infections.
 """
```
SamuelBrand1, line 6: Not really a problem, but what config are we going for on doc string cols allowed per line?

EpiAware/src/EpiInfModels/EpiData.jl
```diff
@@ -1,4 +1,4 @@
-"""
+@doc raw"
```
SamuelBrand1, line 1: Is there any use-case for _not_ using `@doc raw"...some docs..."` pattern?

SamuelBrand1 (commented):
Looking at `getting_started` now

EpiAware/src/EpiInfModels/EpiData.jl
seabbs: I don't think so?

EpiAware/docs/src/examples/getting_started.jl
```diff
@@ -174,39 +190,85 @@ y_t \sim \text{NegBinomial}\Big(\mu = \sum_{s\geq 0} K[t, t-s] I(s), r\Big). \\
...
+    tspan = (1, 100))
+
 # ╔═╡ e813d547-6100-4c43-b84c-8cebe306bda8
 md"
 We also set up the inference to occur over 100 days.
 "
 
 # ╔═╡ c7580ae6-0db5-448e-8b20-4dd6fcdb1ae0
-time_horizon = 100
+time_horizon = 30
```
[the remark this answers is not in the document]
SamuelBrand1: Yes. I think we have a few issues on the tutorial:

- What workflow is the best to see first.
- Most representative model to use.

EpiAware/src/EpiAwareBase/types.jl
SamuelBrand1: Shall we make that a standard. I was doing 92 char per line for doc strings.

EpiAware/docs/src/examples/getting_started.jl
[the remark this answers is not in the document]
seabbs: So those are issues with the tutorial on `main` IMO. Here we are just trying to get it to work as well as the one on main (which is working not that well but doesn't fail)

SamuelBrand1:
Want to make me a reviewer? I'm distant enough I think due to going on holiday in the middle of this PR?

seabbs:
I can't becaise you opened it

SamuelBrand1:
> I can't because you opened it. We need to just mutually agree we are okay with merging and then can do

I'm good. I approve.

seabbs (approved):

PR #171, opened 2024-03-25
Issue 51: First pass at broadcasting

### LINKED ISSUE #51: Inclusion of day of week observation processes
Should we include some simple day of week periodicity in order to better reflect real data in both the simulated scenarios and in the inference models (i.e. in the observation process). This issue is based on #34 

## Advantages

- Potentially better reflects real-world data
- Similarly, potentially better reflects real-world practice
- Will likely improve the performance of all models on real-world data

## Disadvantages

- We will need to choose what periodicity to simulate
- We will need to choose a modelling approach for inference (ideally from common practice - but there are still a few options to pick from).
- Make our simulated data more complex and so perhaps harder to cleanly visualise etc

## My thoughts

Overall the advantages outweigh the negatives so long as we pick something simple for the simulated data (a single scenario) and also stick with a simple method for interence. @SamuelBrand1 @zsusswein what do you both think?

seabbs:
This PR partly resolves #51 by introducing a broadcasting approach to `EpiLatentModels` which can be used both for day of the week models and for piecewise constant latent models. 

Depending on how we feel about this approach we may want to abstract some of this into its own module at some point to allow broadcasting in other parts of the code base.

It does not add the ascertainment `EpiObsModel` or `Random` `EpiLatentModel` which will be needed to full address #51.


seabbs:
Coverage is a bit low so I will address that but I think this is still good to review as is.

EpiAware/src/EpiLatentModels/broadcast/latentmodel.jl
```diff
@@ -0,0 +1,61 @@
...
+## Arguments
+- `model::BroadcastLatentModel`: The broadcast latent model.
+- `n::Any`: The number of samples to generate.
+
+## Returns
+- `broadcasted_latent`: The generated broadcasted latent periods.
+- `latent_period_aux...`: Additional auxiliary information about the latent periods.
+
+"
+@model function EpiAwareBase.generate_latent(model::BroadcastLatentModel, n)
```
SamuelBrand1, line 55: Just to be clear, with `RepeatEach` with (say) `m=7` this generates 7 random variables. E.g. this is for specific day of week effects and has to be blended with other modelling to have day of week cyclic behaviour along with longer trend effects?

EpiAware/src/EpiLatentModels/broadcast/latentmodel.jl
```diff
@@ -0,0 +1,61 @@
+@doc raw"
+The `BroadcastLatentModel` struct represents a latent model that supports broadcasting of latent periods.
+
+## Constructors
+- `BroadcastLatentModel(model::M; period::Int, broadcast_rule::B)`: Constructs a `BroadcastLatentModel` with the given `model`, `period`, and `broadcast_rule`.
```
SamuelBrand1, line 5: These constructors look the same?

EpiAware/src/EpiLatentModels/EpiLatentModels.jl
```diff
@@ -3,19 +3,29 @@ Module for defining latent models.
 """
 module EpiLatentModels
 
+using Base: AbstractBroadcasted
```
SamuelBrand1, line 6: We aren't doing [custom broadcasting methods](https://docs.julialang.org/en/v1/manual/interfaces/#man-interfaces-broadcasting) so I'm not sure about having this as part of the `AbstractBroadcasted` type tree; unless thats a future goal. Its probably harmless though.

SamuelBrand1 (approved):
I like this approach, nice and flexible.

EpiAware/src/EpiLatentModels/broadcast/latentmodel.jl
seabbs: Yes exactly that would be the idea. I plan to make a new PR with a `CombineLatentModels` struct and method and another one for `ascertainment` (in `EpiObsModels`).

EpiAware/src/EpiLatentModels/broadcast/latentmodel.jl
seabbs: The first has keywords and the latter doesn't? That does make them different, right?

SamuelBrand1, line 5:
```suggestion
- `BroadcastLatentModel(;model::M; period::Int, broadcast_rule::B)`: Constructs a `BroadcastLatentModel` with the given `model`, `period`, and `broadcast_rule`.
```

SamuelBrand1: Yeah right made a suggestion

seabbs:
I think this addresses the coverage gaps and includes your suggestions.

PR #173, opened 2024-03-26
Overflow safety for negative-binomial

### LINKED ISSUE #172: Stress test `NegativeBinomial`
Occasionally, I hit errors on using `Distributions.NegativeBinomial` when I think that `logpdf = -Inf` would be more stable.

This issues is about determining more precisely the failure mode here, and negative interactions with compiled reverse passes (e.g. using `AutoReverseDiff(true)`).

SamuelBrand1:
It turns out that inside the Negative binomial distribution lives a call to `trunc` which throws an `Inexact` error for very large values of mean (`>1e17`)... this is because of a large mean method for Poisson. This occasionally can get triggered by using a random walk for ``R_t`` during warm-up and was causing silent occasional fails in conjunction with compiled reverse AD.

I've put in an overflow safety that seems to fix this problem, but the mean will be wrong if we ever want to do inference on a planet with a much larger population than current earth...

Closes #172

### DIFF
--- a/EpiAware/docs/src/examples/getting_started.jl
+++ b/EpiAware/docs/src/examples/getting_started.jl
@@ -29,6 +29,8 @@ begin
     using LinearAlgebra
     using Transducers
     using ReverseDiff
+
+    Random.seed!(1)
 end
 
 # ╔═╡ 3ebc8384-f73d-4597-83a7-07a3744fed61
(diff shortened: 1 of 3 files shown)


EpiAware/docs/src/examples/getting_started.jl
```diff
@@ -157,9 +159,11 @@ The observation model is a negative binomial distribution parameterised with mea
 ```math
 \sigma_{\text{rel}} =(1/\sqrt{\mu}) + (1 / \sqrt{r}).
 ```
-It is standard to use a half-t distribution for standard deviation priors (e.g. as argued in this [paper](http://www.stat.columbia.edu/~gelman/research/published/taumain.pdf)); we specialise this to a Half-Normal prior and use an _a priori_ assumption that a typical observation fluctuation around the mean (when the mean is $\sim\mathcal{O}(10^2)$) would be 10%. This implies a standard deviation prior,
+It is standard to use a half-t distribution for standard deviation priors (e.g. as argued in this [paper](http://www.stat.columbia.edu/~gelman/research/published/taumain.pdf)); we specialise this to a Half-Normal prior and use an _a priori_ assumption that a typical observation fluctuation around the mean (when the mean is $\sim\mathcal{O}(10^2)$) would be 1%, which is close to Poisson noise.
+
+This implies a standard deviation prior,
 ```math
-1 / \sqrt{r} \sim \text{HalfNormal}\Big(0.1 ~\sqrt{{\pi \over 2}}\Big).
+1 / \sqrt{r} \sim \text{HalfNormal}\Big(0.01 ~\sqrt{{\pi \over 2}}\Big).
```
seabbs, line 166: Why not update the default?

seabbs (approved):
Seems like a reasonable fix. Worth opening an issue with Distributions.Jl?

EpiAware/docs/src/examples/getting_started.jl
SamuelBrand1: I was just trying out a lower variance neg bin sampling... I think deciding the appropriate defaults is an issue.

PR #175, opened 2024-03-27
Issue 51: Adds Ascertainment

seabbs:
This PR does further work on #51 by adding an ascertainment model to `EpiObsModels`. This just leaves specifying an appropriate latent process and link function which will be addressed in an open PR. one open question is whether or not we want a `dayofweek` helper function as a latent process or/and we want a observation model that specifies a link etc.


SamuelBrand1 (approved):
I like this. Makes sense.

Issue #165, opened 2024-03-21
Standard surveillance output


seabbs:
I think we want to define some standard surveillance output for `_make_epi_aware` as this will help with postprocessing and be of general utility (given the aim of the tooling). 

I think the way to we want to do this is to create a new abstract function in `EpiAwareBase` that is called in the return of `_make_epi_aware` and dispatches on the `EpiInfModel` type used. This is so that we can have a generic method that just uses the expected infections for its metrics and more specific methods that use the already computed reproduction number estimates and growth rate calculations. 

Alternatively, we could consider this as a post-processing step for the model output (i.e it works with a for object) but I think its nice to make use of the struct information we have to provide custom solutions which I think favours doing this in model (albeit only as part of the generated quantities).

SamuelBrand1:
I think this is a really good idea. Should be reasonable to achieve, we want or already have the functions to fill this in.

We can also add a pretty print to the struct so you get a lot of information (Unicode plots even?)

seabbs:
> We can also add a pretty print to the struct so you get a lot of information 

I think we want more pretty printing of structs in general and wondered if there was a package out there. Can't find anything right now though

seabbs:
Aside from structs are there any other data structures (and links please) we could be thinking about here to use to control dispatch?

SamuelBrand1:
> > We can also add a pretty print to the struct so you get a lot of information
> 
> I think we want more pretty printing of structs in general and wondered if there was a package out there. Can't find anything right now though

I think the canonical way of doing this is overloading [`Base.show`](https://docs.julialang.org/en/v1/base/io-network/#Base.show-Tuple{IO,%20Any,%20Any}) with a custom method

seabbs:
Yeah I saw this and I was wondering if anyone had already done work on that for this kind of use case

seabbs:
This kind of thing: https://github.com/MechanicalRabbit/PrettyPrinting.jl

Issue #116, opened 2024-03-06
Function to backcalculate Rt and the growth rate


seabbs:
As shown here: 

https://github.com/CDCgov/Rt-without-renewal/blob/0c03621be16aa66c4ada8ec7248dafeeee68cf0f/EpiAware/docs/src/examples/getting_started.jl#L331-L332

We need a function to backcalculate Rt and the growth rate from infections generated in an arbitrary model

SamuelBrand1:
This is a good idea, but not really implementable in an _arbitrary model_. Do you mean one with a single infected type as per how we have been doing it?

seabbs:
yes and I was thinking we would use it in the generated quantities return and make sure it only got called via multiple dispatch when it would work (would need to have some kind of function that it nested in though not sure what to call that given that generated_quantities is taken...) -> maybe `situtational_awareness_metrics()`?

seabbs:
backcalculating Rt is handled in #184

seabbs:
Is this complete? I don't think we have a function for the growth rate explicitly (though as it can wrap the Rt one perhaps we do)

SamuelBrand1:
I'm doing this for the pipeline... which means avoiding dealing with multi-groups. Would that close this issue?

seabbs:
Shall we just make a `expected_growth` or similar to match the current `expected_Rt` in EpiAware and use that?

PR #184, opened 2024-04-21
Issue 116: Add function to back calculate Rt

seabbs:
This started as a PR for #165 (hence the branch name) but is just addressing (and closing #116) as #165 is turning out to be a little more complex than expected. 

Here we add two new functions

- `expected_Rt` Dispatching on a `EpiData` object or taking arguments directly.
- `expected_rt` dispatching on a vector and taking the log differences.


seabbs:
Closing as we. are now planing on doing this within the inf submodel

seabbs:
Okay looking at this again I vote to review and merge as a partial PR towards #116 leaving the expected growth rate to work out

EpiAware/src/EpiInfModels/EpiData.jl
```diff
@@ -61,3 +61,36 @@ struct EpiData{T <: Real, F <: Function}
...
+- `exp_Rt::Vector{Float64}`: A vector of expected Rt values.
+
+## Examples
+
+```julia
+using EpiAware
+
+data = EpiData([0.2, 0.3, 0.5], exp)
+infections = [100, 200, 300, 400, 500]
+expected_Rt(data, infections)
```
SamuelBrand1, line 86: close the example block

EpiAware/src/EpiInfModels/EpiData.jl
```diff
@@ -61,3 +61,36 @@ struct EpiData{T <: Real, F <: Function}
...
+# Arguments
+- `data::EpiData`: An instance of the EpiData type containing generation interval data.
+- `infections::Vector{<:Real}`: A vector of infection data.
+
+# Returns
+- `exp_Rt::Vector{Float64}`: A vector of expected Rt values.
+
+## Examples
+
+```julia
```
SamuelBrand1, line 81:
```suggestion
```julia
using EpiAware
data = EpiData([0.2, 0.3, 0.5], exp)
infections = [100, 200, 300, 400, 500]
expected_Rt(data, infections)
```
```

File: EpiAware/src/EpiInfModels/EpiData.jl [hunk 3]
```diff
@@ -61,3 +61,36 @@ struct EpiData{T <: Real, F <: Function}
...
+using EpiAware
+
+data = EpiData([0.2, 0.3, 0.5], exp)
+infections = [100, 200, 300, 400, 500]
+expected_Rt(data, infections)
+"
+function expected_Rt(data::EpiData, infections::Vector{<:Real})
+    n = data.len_gen_int
+
+    denom_Rt = [dot(reverse(data.gen_int),
```
SamuelBrand1, line 91: Might be worth having an `@assert` on length of infections vs `n`

SamuelBrand1 (changes requested):
LGTM. One docstring small fix.

Worth noting that this will give the _effective_ $R_t$ as opposed to unadjusted $R_t$; this will diverge as we add population effects e.g. #245 . This is a good thing but maybe also mention in docstring.

seabbs:
> Worth noting that this will give the effective  as opposed to unadjusted 

I think we should put this in the paper as a discussion point

SamuelBrand1:
> > Worth noting that this will give the effective  as opposed to unadjusted
> 
> I think we should put this in the paper as a discussion point
> 
> > One docstring small fix.
> 
> Do we have anything in CI etc that tests docstring examples at least run / docstrings are valid? It seems like this is quite important so it would be odd if the julia community didn't have tools for this

https://documenter.juliadocs.org/stable/man/doctests/#Doctesting-as-Part-of-Testing

IIRC we have talked about this but not done it yet.

SamuelBrand1 (approved):
LGTM

PR #185, opened 2024-04-23
Improved getting started tutorial and defaults

### LINKED ISSUE #174: Appropriate getting started tutorial and defaults
I was just trying out a lower variance neg bin sampling... I think deciding the appropriate defaults is an issue.

_Originally posted by @SamuelBrand1 in https://github.com/CDCgov/Rt-without-renewal/pull/173#discussion_r1540834169_

I think we need to reconsider the getting started example. The default model to demonstrate should be a renewal model, in my opinion.

Also, the direct infections model probably shouldn't have negative binomial sampling? It rather obviously difficult to identify compared to latent process variation in a way that (say) a renewal model wouldn't find problematic.

SamuelBrand1:
This PR demonstrates (mainly) recreating an early COVID modelling study in South Korea as a way of demonstrating the basic functionality of `EpiAware`.

A particular focus is on demonstrating that each modular component of the overall epidemiological model is itself a sampleable `Turing.Model` object.

Closes #174


seabbs (approved):
This is a really nice improvement. It would be great to add a follow up that better maps to realise world practice with delays from infection to report, and potentially ascertainment.

The latter may flag name collisions which we will need to solve for (likely using prefixes). 

We also may need to switch back to a weekly process when we had morebfeature in to maintain the aim of having a fast model (with a note about how this could likely be improved)

PR #190, opened 2024-04-29
188 using prefix to separate latent random variables used in different contexts

### LINKED ISSUE #188: Using `prefix` to separate latent random variables used in different contexts
At the moment, a user might want to use a constructor for `<: AbstractLatentModel` more than once in a nested model. For example, having a latent process for $R_t$ _and_ a time-varying ascertainment probability. 

Functionality for prefixing variable names exists, see [here](https://turinglang.org/DynamicPPL.jl/stable/api/#DynamicPPL.@submodel). This issue is about determining and implementing the best pattern to deal with naming combinations

SamuelBrand1:
This PR fixes the problem mentioned in #188 : That is that different components of a full epi model could both call a latent model leading to naming collision.

The solution is that all latent processes generated to be used in an epi model (for example, $\log R_t$) get sampled with naming prefix form `latent.varname`, whereas latent processes sampled as part of an overall observation model get prefix form `obs.varname`. This is unit tested with a time varying log-ascertainment scale [here](https://github.com/CDCgov/Rt-without-renewal/blob/5aa10e5e06d80fb05cfee578f0a67e677351815f/EpiAware/test/EpiAwareUtils/turing-methods.jl#L70C1-L78C64).

For me, the only open question is whether we want to custom generate prefixes, leave the version in this PR or plan a future issue about custom prefixes (e.g. as part of model struct).

Closes #188

### DIFF
--- a/EpiAware/src/EpiAwareUtils/turing-methods.jl
+++ b/EpiAware/src/EpiAwareUtils/turing-methods.jl
@@ -15,13 +15,14 @@ A `DynamicPPPL.Model` object.
         y_t, time_steps, epi_model::AbstractTuringEpiModel;
         latent_model::AbstractTuringLatentModel, observation_model::AbstractTuringObservationModel)
     # Latent process
-    @submodel Z_t, latent_model_aux = generate_latent(latent_model, time_steps)
+    @submodel prefix="latent" Z_t, latent_model_aux=generate_latent(
+        latent_model, time_steps)
 
     # Transform into infections
     @submodel I_t = generate_latent_infs(epi_model, Z_t)
 
     # Predictive distribution of ascertained cases
-    @submodel generated_y_t, generated_y_t_aux = generate_observations(
+    @submodel prefix="obs" generated_y_t, generated_y_t_aux=generate_observations(
         observation_model, y_t, I_t)
 
     # Generate quantities
--- a/EpiAware/docs/src/examples/getting_started.jl
+++ b/EpiAware/docs/src/examples/getting_started.jl
@@ -1,5 +1,5 @@
 ### A Pluto.jl notebook ###
-# v0.19.39
+# v0.19.41
 
 using Markdown
 using InteractiveUtils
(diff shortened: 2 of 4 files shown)


EpiAware/src/EpiAwareUtils/turing-methods.jl
```diff
@@ -15,13 +15,14 @@ A `DynamicPPPL.Model` object.
...
-    @submodel Z_t, latent_model_aux = generate_latent(latent_model, time_steps)
+    @submodel prefix="latent" Z_t, latent_model_aux=generate_latent(
+        latent_model, time_steps)
 
     # Transform into infections
     @submodel I_t = generate_latent_infs(epi_model, Z_t)
 
     # Predictive distribution of ascertained cases
-    @submodel generated_y_t, generated_y_t_aux = generate_observations(
+    @submodel prefix="obs" generated_y_t, generated_y_t_aux=generate_observations(
```
seabbs, line 25: why have we changed to having no spaces around `=`?

seabbs (approved):
Nice. One comment about styling. 

I think we should go for this now and make another issue to think about a nicer custom approach. Something to flag is we also have this problem when we have stacked ascertainment models potentially (if they share a latent model for example) or multiple observations processes.

EpiAware/src/EpiAwareUtils/turing-methods.jl
SamuelBrand1: I've gone with the form of argument input described [here](https://turinglang.org/DynamicPPL.jl/stable/api/#DynamicPPL.@submodel). I haven't tested alternatives, but I expect that the `@submodel` macro is looking for a pattern of string input `prefix=...` with no spaces.

Another alternative is `@submodel(string_expr, expr)` rather than using spaces but thats also a style deviation from Julia/our norm.

SamuelBrand1:
>  Something to flag is we also have this problem when we have stacked ascertainment models potentially (if they share a latent model for example) or multiple observations processes.

Potentially, changing every `@submodel` to have `prefix=true` could work here... but I'm not sure how readable the varnames would be.

EpiAware/src/EpiAwareUtils/turing-methods.jl
SamuelBrand1: PS I'm not sure why JuliaFormatter got rid of the second set of whitespaces around `=`

seabbs: Okay no worries happy to listen to the formatting gods on this one

seabbs:
(merge was automated but agree it was good to merge to fix the issue and then we can iterate in another feature for something more magical with a bit less urgency)

PR #191, opened 2024-04-29
update installation instructions and opinionated list of useful globally installed packages

### LINKED ISSUE #187: Installation instructions and suggested julia environment settings
This issue is about adding installation instruction for `EpiAware` aimed at installing from this repo rather than local cloning.

Within scope are adding some opinions about useful packages to include in the default Julia env e.g. `@v.10`.

SamuelBrand1:
Does as per title 

Closes #187


EpiAware/README.md
```diff
@@ -7,6 +7,14 @@
 
 A `Julia` package for flexible and composable modeling and inference of the effective reproduction number (Rt) and other situational awareness signals in the presence of different latent generative processes and observation models.
 
+## Installation instruction
+
+Eventually, `EpiAware` is likely to added to the Julia registry. Until then, you can install it from the `/EpiAware` sub-directory of this repository by running the following command in the Julia REPL:
```
seabbs, line 12:
```suggestion
Eventually, `EpiAware` is likely to be added to the Julia registry. Until then, you can install it from the `/EpiAware` sub-directory of this repository by running the following command in the Julia REPL:
```

EpiAware/README.md
```diff
@@ -81,3 +89,13 @@ H-.->I
 ## Pluto scripts
 
 We use [`Pluto.jl`](https://plutojl.org/) scripts as part of our documentation and testing. The scripts are located in `docs/src/examples` and can be run using the `Pluto.jl` package. We recommend using the version of `Pluto` that is pinned in the `Project.toml` file defining the documentation environment. An entry point to running or developing this documentation is the `docs/pluto-scripts.sh` bash shell script. Run this from the root directory of this repository.
```
seabbs, line 91: Will this activate the docs environment for people?

EpiAware/README.md
```diff
@@ -81,3 +89,13 @@ H-.->I
...
+
+## Useful Julia packages for global environment (optional and opinionated)
+In our view these packages are useful for a global environment available to other environments (e.g. add to the `@v1.10` environment), but are not required for the `EpiAware` package to work.
+
+- `Revise`: For modifying package code and using the changes without restarting Julia session.
```
seabbs, line 96: Do we want to go further and suggest anything for the startup file? (i.e `Term` and `Revise`?)

README.md
```diff
@@ -8,6 +8,8 @@ To evaluate the role of the latent generative process in the estimation of the e
 
 In `pipeline`, we will use `EpiAware` to explore whether the inclusion of a renewal process in the latent generative process improves the estimation of Rt and other situational awareness signals. This exploration is then written as a manuscript for publication (see `manuscript` for work supporting this).
 
+To install `EpiAware`, see installation instructions in the `EpiAware` directory.
```
seabbs, line 11: add a relative link to them (i.e [installation instructions](EpiAware/README.md))

seabbs (approved):
This is really nice. A few clarifying comments and potential suggestions for future improvements.

EpiAware/README.md
seabbs: This also now suggests a clone followed by running a bash script. Can pluto do anything smart about pointing at a URL?

SamuelBrand1: > Will this activate the docs environment for people?
Yes, the bash script does that.

SamuelBrand1: > This also now suggests a clone followed by running a bash script. Can pluto do anything smart about pointing at a URL?

Good spot. There is a tension here because on one hand we're using Pluto as a document generating platform (inspired by how [`TuringGLM.jl`](https://github.com/TuringLang/TuringGLM.jl) does docs), in which case then this is fine but otoh suggesting a script.

Not sure what the right balance is here.

SamuelBrand1: In other packages generally you run tutorials locally and the onus is on the user to choose how to do that (e.g. C&P into REPL, use a Jupyter nb, use a Pluto nb etc etc), and the dev of the documentation is not added.

On the other extreme, we can be fancy and use something like binder for Pluto. There might be some pain there, because Pluto as default is self-contained (e.g. installs its own package version as part of a nb specific env) whereas here it isn't. For example, we use the Pluto nbs for dev as well as docs and therefore we prefer a pattern of using `Pkg.develop` on `EpiAware` rather than `Pkg.add` (which I think is easier for doing binder on but I might be mistaken).

EpiAware/README.md
SamuelBrand1: Nice idea, added a bit on this.

README.md
SamuelBrand1: Another good move

EpiAware/README.md
seabbs: Yeah this makes this makes sense. My worry is essentially we are expecting a lot of people who are new to Julia to look at this and we want to make it as easy as possible.. 

For now let's leave as is and come back to it if we think of anything that makes working with it easier.

seabbs (approved):

PR #193, opened 2024-04-30
Add analysis pipeline information and defining a set of doable tasks to become issues

SamuelBrand1:
This is a draft PR for making the tasks in the pipeline analysis concrete. The reason its draft is that I'm not sure I have the full plan here and welcome input.

The proposed task list from the pov of issues:

- [ ] Create a `DrWatson` project for the pipeline. This requires a bit of thought about what features we want.
- [ ] Within the project setup an appropriate environment, including loading `EpiAware`
- [ ] Design generation of "truth" data script as described in `README`, along with saving data.
- [ ] Define the full sequence of different scenario options, e.g. as [here](https://juliadynamics.github.io/DrWatson.jl/dev/workflow/#3.-Prepare-simulations).
- [ ] Construct and test a function for going from scenario specification to running inference. At this point defining the output to save is important.
- [ ] Blend `Dagger.jl` checkpointing into the scenario inference loop process, e.g. like [here](https://juliaparallel.org/Dagger.jl/dev/checkpointing/).
- [ ] Evaluation method on inference output. I'm assuming we outsource this to `scoringutils`.

### DIFF
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -5,7 +5,7 @@ An overview of the purpose of this pipeline is:
 1. Generate cases from a defined gold standard model to act as synthetic "truth" data.
 2. Build a matrix set of scenario models, defined using `EpiAware`.
 3. Run inference on each scenario model using `EpiAware`. This will be done using the No U-Turns Sampler (NUTS) algorithm.
-4. Compare inferred time-varying reproductive number and forecasts to the truth data to evaluate the performance of the inference, and/or forecast. 
+4. Compare inferred time-varying reproductive number and forecasts to the truth data to evaluate the performance of the inference, and/or forecast.
 
 ## Synthetic truth data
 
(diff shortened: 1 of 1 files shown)


seabbs:
Yes that is a good summary. I think we want to tackle these as a meta-issue and individual PRs though vs one mega PR? 

I think we originally planned to have the post processing in R but due to staffing changes on this project I'm not sure that is really needed? That being said unless we find anything suitable it seems like using scoringutils from within Julia is likely to be the least painful option for the scoring?

SamuelBrand1:
> I think we want to tackle these as a meta-issue and individual PRs though vs one mega PR?

100% agree. This PR is a small change to `pipeline/README.md` and setting the individual issues to put in the task list of a meta-issue.

SamuelBrand1:
> That being said unless we find anything suitable it seems like using scoringutils from within Julia is likely to be the least painful option for the scoring?

I agree here too; it depends how much we want to save on disk though? I'm fairly agnostic to either outputting in a `scoringutils` friendly way to disk or direct to an R process. The steps are almost the same in either case.

SamuelBrand1:
Just to be double clear, if you guys are ok with this PR then I'll add a meta-issue and start ticking the tasks off.

seabbs:
> I agree here too; it depends how much we want to save on disk though?

I think it doesn't matter hugely either way. If we could do something fancy with parquet that might be interesting but I don't think we need to over think this in the first and second instancees

seabbs (approved):

PR #194, opened 2024-04-30
Issue 192: add Identity matrix specification for MvNormal

### LINKED ISSUE #192: Move away from depreciated multivariate normal call
We use multivariate normal calls based on the distribution `Distributions.MvNormal` alot in the code base, since we construct latent processes as transformations on samples from this dist.

However, the way we do this is deprecated and we should move to address this warning:

> Warning: `MvNormal(σ::AbstractVector{<:Real})` is deprecated, use `MvNormal(LinearAlgebra.Diagonal(map(abs2, σ)))` instead.

seabbs:
This PR closes #192 by switching to use `LinearAlgebra.I` rather than `ones` in the specification of the `MvNormal`.


SamuelBrand1 (approved):
LGTM.

PR #199, opened 2024-05-01
Pin to LERC-3

### LINKED ISSUE #198: `LERC-4.0` incompatibility
The raster image compression dep [LERC](https://github.com/Esri/lerc) has upgraded to `4.0` in a way that has caused downstream problems with `Plots.jl`; which causes problems for us with document generation.

Fix is to pin to `LERC-3.0` as described [here](https://github.com/JuliaPackaging/Yggdrasil/issues/8580#issuecomment-2088330577) as at least a temporary solution.

SamuelBrand1:
Does as says to fix #198 .

_NB:_ This might be duplicating the Julia registry pin [here](https://github.com/JuliaRegistries/General/pull/105958).

Closes #198


seabbs (approved):

seabbs:
It looks like this doesn't fix it? Shall we see if the central change does the job?

SamuelBrand1:
Hmmm I think this issue is already fixed by `giordano-patch-1` merging into the Julia registry... still want to add a pinned version?

SamuelBrand1:
I'm closing this because I think the central change is a better way to go

PR #203, opened 2024-05-02
New latent models

seabbs:
This PR adds a range of new latent models (WIP): 

- [x] Intercept
- [x] FixedIntercept
- [x] CombineLatentModels
- [x] TransformLatentModel
- [x] HierarchicalNormal

it also renames the broadcasting helpers to `broadcast_{helper}` name for consistency.


SamuelBrand1:
At a glance of your WIP this is very useful functionality, e.g. models with constant (but unknown) $R_0$ with time-varying ascertainment come in scope easily.

Looking forward to full review when its ready.

SamuelBrand1:
> Do you think we want to call combining latent models:
> 
> * `CombineLatentModel`
> * `NestLatentModel`
> * other
> 
> In the first instance I was going to write a struct that took 2 latent models and combined one with the other (with an optional link). In the future could have a plural version that does this iteratively?

Do we need a special name for combined models really?

seabbs:
How do we otherwise dispatch to multiple latent models at once?

SamuelBrand1:
> How do we otherwise dispatch to multiple latent models at once?
> 
> i.e
> 
> ```julia
> latent_model = CombineLatentModels(
> 	[AR(), 
>     Intercept(),
>     PiecewiseLinear(7)]
> )
> ```

Right, I was more thinking of the way we combine models already.

SamuelBrand1:
Update: For clarity, @seabbs and I discussed use-cases of `CombineLatentModels` f2f.

For example, one might want a linear model

```math
y_t = \beta_0 + \beta X_t + \epsilon_t
```
Where $\beta_0$, $\beta$ had different priors, and the residual process $(\epsilon_t)_{t\geq 0}$ was (say) an AR process. 

This PR exposes sampling for this kind of modelling functionality.

seabbs:
I think this in a reasonable place now for a review @SamuelBrand1. Note we do definitely do better with some of these and working on this in general made me think we need more tools to standardise how latent models etc are defined but I think those are for doing later.

EpiAware/src/EpiLatentModels/CombineLatentModels.jl
```diff
@@ -0,0 +1,62 @@
...
+"
+@model function EpiAwareBase.generate_latent(latent_models::CombineLatentModels, n)
+    @submodel final_latent, latent_aux = _accumulate_latents(
+        latent_models.models, 1, fill(0.0, n), [], n, length(latent_models.models))
+
+    return final_latent, (; latent_aux...)
+end
+
+@model function _accumulate_latents(models, index, acc_latent, acc_aux, n, n_models)
+    if index > n_models
```
SamuelBrand1, line 53: This adds to the number of functions being generated with conditionals; which means that if we use `ReverseDiff` as our AD method we'll have to restrict to uncompiled/re-taped calculations.

I'm not 100% against that, this is how stan at least used to go (e.g. [here](https://arxiv.org/pdf/1509.07164) but it might well have moved forward), but I think its worth reflecting upon whether we can mitigate with performance annotations (c.f. [here](https://juliadiff.org/ReverseDiff.jl/dev/api/#Optimization-Annotations)).

seabbs: I thought the conditional issue was only for dynamic quantities?

SamuelBrand1: Gotcha. This is ok.

seabbs: I'm not 100% on this FYI!

SamuelBrand1: I'm thinking!

seabbs:
Sorry just noticed I failed to push a commit containing the new `HierarchicalNormal` tooling.

seabbs:
@SamuelBrand1 where are we on this?

EpiAware/src/EpiLatentModels/CombineLatentModels.jl
```diff
@@ -0,0 +1,62 @@
...
+
+# Returns
+- `combined_latents`: The combined latent variables generated from all the models.
+- `latent_aux`: A tuple containing the auxiliary latent variables generated from each individual model.
+
+# Example
+"
+@model function EpiAwareBase.generate_latent(latent_models::CombineLatentModels, n)
+    @submodel final_latent, latent_aux = _accumulate_latents(
+        latent_models.models, 1, fill(0.0, n), [], n, length(latent_models.models))
```
SamuelBrand1, line 47: Have you tested this for AD/would you like me to? 

```julia
fill(0.0, n)
```
Would default to a Vector{Float64} which might be a problem for using `DualNumbers`.

EpiAware/src/EpiLatentModels/Intercept.jl
```diff
@@ -0,0 +1,78 @@
...
+
+```julia
+using Distributions, Turing, EpiAware
+int = Intercept(Normal(0, 1))
+int_model = generate_latent(int, 10)
+rand(int_model)
+int_model()
+```
+"
+@kwdef struct Intercept{D <: Sampleable} <: AbstractTuringLatentModel
```
SamuelBrand1, line 20: We might want to reason on models that have an intercept, so is it worth defining an abstract type something like

```julia
AbstractTuringIntercept <: AbstractTuringLatentModel
```

then making both the intercepts here a sub-type of `AbstractTuringIntercept`?

seabbs: ah yeah that is a good idea

EpiAware/src/EpiLatentModels/CombineLatentModels.jl
seabbs: I haven't that is a good point. Can we instead just define it differently?

SamuelBrand1: Usually, I go with the `eltype` of a generated variable from `~` e.g. like [here](https://github.com/CDCgov/Rt-without-renewal/blob/4ab6cc2e8ac4a0e89f0973324544e0b1953a195b/EpiAware/src/EpiLatentModels/RandomWalk.jl#L91C5-L91C39). I'll have a think about how to do this case.

EpiAware/test/EpiLatentModels/HierarchicalNormal.jl
```diff
@@ -0,0 +1,45 @@
...
+    n_samples = 100
+    samples = sample(fixed_model, Prior(), n_samples; progress = false) |>
+              chn -> mapreduce(vcat, generated_quantities(fixed_model, chn)) do gen
+        gen[1]
+    end
+
+    theoretical_mean = 0.2
+    theoretical_var = 0.1^2
+
+    @test isapprox(mean(samples), theoretical_mean, atol = 0.1)
```
SamuelBrand1, line 39: I don't think you need this and the KS test

EpiAware/test/EpiObsModels/Ascertainment.jl
```diff
@@ -0,0 +1,23 @@
+@testitem "Test Ascertainment constructor" begin
+    using Turing
+    function natural(x)
```
SamuelBrand1, line 3: Why not `x -> x`?

SamuelBrand1: Ah is it to check `==` at the bottom?

SamuelBrand1 (changes requested):
This looks very good. I'd like to have an inference check on Combine latent models (e.g. generate some data and then reinfer the parameters) then I'm super happy.

EpiAware/test/EpiLatentModels/HierarchicalNormal.jl
seabbs: This is a direct copy of what we have in the AR tests so perhaps we move to an issue or leave as its just over covering?

EpiAware/test/EpiObsModels/Ascertainment.jl
seabbs: because I am basic!

EpiAware/src/EpiLatentModels/Intercept.jl
seabbs: Done [c7e42fb](https://github.com/CDCgov/Rt-without-renewal/pull/203/commits/c7e42fb0f98a5f4c657827617e35aaf1f799df31)

EpiAware/test/EpiObsModels/Ascertainment.jl
seabbs: THis being said we don't want to address this here right as the change here was just to take out the use of scale?

seabbs:
> This looks very good. I'd like to have an inference check on Combine latent models (e.g. generate some data and then reinfer the parameters) then I'm super happy.

Shall I do this or are you already looking into it? I was unclear

SamuelBrand1:
> > This looks very good. I'd like to have an inference check on Combine latent models (e.g. generate some data and then reinfer the parameters) then I'm super happy.
> 
> Shall I do this or are you already looking into it? I was unclear

I've started, happy to catch this.

EpiAware/test/EpiLatentModels/HierarchicalNormal.jl
SamuelBrand1: Leave then is fine

seabbs (commented):
LGTM (self-review). This combined model test is great.

PR #204, opened 2024-05-03
Design generation of truth data script as described in readme along with saving data

### LINKED ISSUE #202: Design generation of "truth" data script as described in `README`, along with saving data.

SamuelBrand1:
This PR is about adding a "truth" data generating script, with the side benefit of adding `DrWatson.jl` style template for computation.

Implemented is three different mean generation intervals (2, 10, 20 days). These output daily case and latent infections to be saved in `.jld2` format (a version of `HDF5` data structure) along with unknown parameters (e.g. cluster factor of negative binomial sampling).

Pipeline code has been collected into a module `AnalysisPipeline`:
- Added is a `TruthSimulationConfig` struct which has the necessary settings/priors
- `simulate_or_infer` method dispatches on the configuration type. 

The next step is obviously the inferences, but I intend to reuse this pattern by defining an inference struct to dispatch upon.

Closes #202


pipeline/README.md
```diff
@@ -64,3 +78,11 @@ We score each scenario model in two ways:
 ## Running the pipeline
 
 The pipeline structure is built using [DrWatson.jl](https://github.com/JuliaDynamics/DrWatson.jl) for project management of simulation parameters/settings, saved results, and figures. Compute is done with [Dagger.jl](https://github.com/JuliaParallel/Dagger.jl) for organizing parallel computation and checkpointing of results.
+
+## Running pipeline tests
+
+To run the tests for the pipeline, run the following command in the Julia REPL with the root directory as the working directory:
+
+```julia
+include("pipeline/test/runtests.jl")
```
seabbs, line 87: can't we make them be in the pipeline directory for all of these and use test?

SamuelBrand1: Not precisely sure what you mean?

SamuelBrand1: I agree that this way of testing is fairly clunky. Maybe a clever use of `@testitem`? I am conscious that these will ultimately live in different places so I don't want to intermingle the pipe tests with the `EpiAware` tests too much

pipeline/src/TruthSimulationConfig.jl
```diff
@@ -0,0 +1,74 @@
...
+Simulates or infers the truth process and observations based on the given configuration.
+
+# Arguments
+- `config::TruthSimulationConfig`: The configuration object containing the parameters for the simulation.
+
+# Returns
+A dictionary containing the sampled infections and observations, along with other relevant information.
+
+"""
+function simulate_or_infer(config::TruthSimulationConfig)
```
seabbs, line 43: Can you clarify why we want this to be simulate or infer vs having simulate and infer functions

pipeline/test/test_SimulationConfig.jl
```diff
@@ -0,0 +1,15 @@
+
+# Test the TruthSimulationConfig struct constructor
+@testset "TruthSimulationConfig" begin
```
seabbs, line 3: why are we not using testitem?

pipeline/README.md
seabbs: I'm just confused why you don't tell people to activate the pipeline environment and run test?

pipeline/src/TruthSimulationConfig.jl
SamuelBrand1: Only one reason: makes the overall code pattern even more similar on the next PR. But happy to change this. I was clearly being influenced by reading about `DrWarson.produce_or_load`.

pipeline/test/test_SimulationConfig.jl
SamuelBrand1: Basically, I was keeping the testing very separate from `EpiAware`. Might be worth just being a bit careful and seeing what I can do here?

pipeline/README.md
SamuelBrand1: That doesn't work the way this is set up... I can change it so that could work though.

pipeline/test/test_SimulationConfig.jl
seabbs: does it not only show up if you are in the environment where those tests lives? Hmm I wonder if there is a good way to filter. Seems very annoying to not be able to use it?

seabbs: Shall wel make this an issue and leave for now?

pipeline/README.md
seabbs: as `pipeline` has no environment? I don't feel strongly if there is a strong reason for the way its set up at the mo?

pipeline/src/TruthSimulationConfig.jl
seabbs: Ha fair enough. Shall we pin it as an issue to discuss once we have the next PR?

pipeline/README.md
seabbs: once we have clarity on this its all good to go from my perspectivee

seabbs (approved):

pipeline/README.md
SamuelBrand1: It does have an env but I've not set up the name to have an associated `test`, just left as manual. So if you activate the env and just do 

`> test`

it chucks

> `trying to test unnamed project `

SamuelBrand1: Basically, if I'm not adding it to the test CI or the nice `TestItem` interface then I'm acting relaxed about having the tests as a manual `include(path/to/tests)`.

Obviously, mileage can vary on how relaxed I should be.

seabbs: why wouldn't we have another CI job for this that just triggers when the pipeline folder is updated?

PR #207, opened 2024-05-03
Define the full sequence of different scenario options 

SamuelBrand1:
This is a draft PR for the scenario list spin out.

As per `DrWatson` defaults I define some inference configs as a vector and unspool the combinations using `DrWatson.dict_list`.

I'd love a review on the inference list.


pipeline/scripts/generate_truth_data.jl
```diff
@@ -0,0 +1,49 @@
+using DrWatson
+@quickactivate "Analysis pipeline"
```
seabbs, line 2: Sad branching problems I see! Note to ourselves for the future is a branch is based on another branch we can't really use squash merging (or we can but its a bit more painful)

pipeline/scripts/inference_pipeline.jl
```diff
@@ -0,0 +1,41 @@
...
+      """)
+
+## Set up the truth Rt and save a plot of it
+using .AnalysisPipeline, Plots, JLD2, EpiAware, Distributions
+
+wkly_ar = AR(
+    damp_priors = [truncated(Normal(0.8, 0.05), 0, 1)],
+    std_prior = HalfNormal(1.0),
+    init_priors = [Normal(0.0, 0.25)]
+) |> ar -> BroadcastLatentModel(ar, 7, RepeatBlock())
```
seabbs, line 22: can we be flash and not repeat the broadcast (so have a tuple of models and then broadcast them all using map)

pipeline/scripts/inference_pipeline.jl
```diff
@@ -0,0 +1,41 @@
...
+      ---------------------------------------------
+      Currently active project is: $(projectname())
+      Path of active project: $(projectdir())
+      """)
+
+## Set up the truth Rt and save a plot of it
+using .AnalysisPipeline, Plots, JLD2, EpiAware, Distributions
+
+wkly_ar = AR(
+    damp_priors = [truncated(Normal(0.8, 0.05), 0, 1)],
```
seabbs, line 19: Is this from the method spec? It seems like its quite high - usually these tend to assume fairly small amounts of autoregressive behaviour)

pipeline/scripts/inference_pipeline.jl
```diff
@@ -0,0 +1,41 @@
...
+    std_prior = HalfNormal(1.0),
+    init_priors = [Normal(0.0, 0.25)]
+) |> ar -> BroadcastLatentModel(ar, 7, RepeatBlock())
+
+wkly_rw = RandomWalk(
+    std_prior = HalfNormal(1.0),
+    init_prior = Normal(0.0, 0.25)
+) |> rw -> BroadcastLatentModel(rw, 7, RepeatBlock())
+
+wkly_diff_ar = AR(
```
seabbs, line 29: again on dry could we be flash and reuse the other Ar ojbect here.

pipeline/scripts/inference_pipeline.jl
```diff
@@ -0,0 +1,41 @@
...
+    std_prior = HalfNormal(1.0),
+    init_priors = [Normal(0.0, 0.25)]
+) |>
+               ar -> DiffLatentModel(; model = ar, init_priors = [Normal(0.0, 0.25)]) |>
+                     diff_ar -> BroadcastLatentModel(ar, 7, RepeatBlock())
+
+## Parameter settings
+sim_configs = Dict(:igp => [DirectInfections, ExpGrowthRate, Renewal],
+    :latent_model => [wkly_ar, wkly_rw, wkly_diff_ar],
+    :gi_mean => [2.0, 10.0, 20.0], :gi_std => 2.0) |>
```
seabbs, line 40: if we are reusing the ones from the simulation then they should be defined in one place and reused in both places vs redefined twice. This goes for anything we define imo

seabbs (commented):
Detail of the models looks good here. I have a few comments about DRY and making sure we are keeping only one source of truth for the various bits of model we use.

pipeline/scripts/inference_pipeline.jl
SamuelBrand1: I'm all in favour of flash!

pipeline/scripts/inference_pipeline.jl
SamuelBrand1: Good spot. This is a C&P from the getting started notebook.

Maybe indicates that we need a chat about priors... obviously random walk is AR(1) with the damp variable fixed at 1, so I think of this as a prior on the scale of "White noise" <-> "random walk". This is IMO the justification for truncating the distribution onto [0,1]. Having said that, it would seem to be more natural in that case to use a `Beta` distribution.

pipeline/scripts/inference_pipeline.jl
SamuelBrand1: Think I've managed to do this in latest commit

pipeline/scripts/inference_pipeline.jl
SamuelBrand1: I've gone with a Jeffrey Prior for the `[0, 1]` valued processes e.g `Beta(0.5, 0.5)`.

pipeline/scripts/inference_pipeline.jl
```diff
@@ -0,0 +1,41 @@
...
+using .AnalysisPipeline, Plots, JLD2, EpiAware, Distributions
+
+wkly_ar = AR(
+    damp_priors = [truncated(Normal(0.8, 0.05), 0, 1)],
+    std_prior = HalfNormal(1.0),
+    init_priors = [Normal(0.0, 0.25)]
+) |> ar -> BroadcastLatentModel(ar, 7, RepeatBlock())
+
+wkly_rw = RandomWalk(
+    std_prior = HalfNormal(1.0),
```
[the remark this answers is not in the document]
SamuelBrand1: Done that, but in this script.

seabbs (approved):
Love it

PR #210, opened 2024-05-07
Make sure `DiffLatentModel` outputs a `AbstractTuringLatentModel`

SamuelBrand1:
Found a small problem in using `Broadcast` where `DiffLatentModel` wasn't outputting a struct in a type that was expected (e.g. a subtype of `AbstractTuringLatentModel`).

Fixed this and slightly extended the unit tests.


seabbs (approved):
Great catch.

Issue #208, opened 2024-05-03
simulate_or_infer or simulate and infer


seabbs:
In #204 a new function `simulate_or_infer` was introduced that dispatches to simulation or inference code depending on the type of struct passed (in the latter case it will once this is implemented in a future PR). Its currently unclear if this abstraction is helpful or just provides another layer of complexity. 

Once the `simulate_or_infer` inference method is implemented we should circle back to this issue and discuss the relative merits of both approaches in order to reduce our technical debt going forward. It is hard to have this conversation first as for me at least its hard to picture the relative merits in abstract.

SamuelBrand1:
In light of #217 is this still an issue? Genuine question... I think we still want to split this because we're not really using dispatch in a particularly useful way?

seabbs:
Yes exactly

PR #213, opened 2024-05-09
Loop over inference configurations script

SamuelBrand1:
This PR does a few things around running the inference scenario loop over possible configurations. 

Specifically, we have generated "truth data" with three different mean generation intervals (see #202 ). The proposal is that for each of these data set we run all the inference scenarios defined in the [`common_scenarios.jl`](https://github.com/CDCgov/Rt-without-renewal/blob/211-construct-and-test-a-function-for-going-from-scenario-specification-to-running-inference-at-this-point-defining-the-output-to-save-is-important/pipeline/scripts/common_scenarios.jl) script.

Contributions:
- A new type `InferenceConfig` which dispatched a method of `simulate_or_infer` that handles the `EpiAware` inference.
- A `DrWatson` style script to run the inference after defining various common values.
- A constructor unit test for `InferenceConfig` and a full inference run test on a short bit of fake data using a randomly selected inference scenario.

### DIFF
--- a/pipeline/scripts/common_param_values.jl
+++ b/pipeline/scripts/common_param_values.jl
@@ -6,6 +6,7 @@
 
       """)
 
+using EpiAware
 # Common generation interval values
 gi_means = [2.0, 10.0, 20.0];
 gi_stds = [2.0];
@@ -13,8 +14,16 @@ gi_stds = [2.0];
 #Common thread count
 num_threads = min(10, Threads.nthreads());
 
-#True Rt
+#Common inference method
+inference_method = EpiMethod(
+    pre_sampler_steps = [ManyPathfinder(nruns = 4, maxiters = 100)],
+    sampler = NUTSampler(adtype = AutoForwardDiff(),
+        ndraws = 2000,
+        nchains = num_threads,
+        mcmc_parallel = MCMCThreads())
+)
 
+#True Rt
 A = 0.3
 P = 30.0
 ϕ = asin(-0.1 / 0.3) * P / (2 * π)
@@ -23,3 +32,6 @@ true_Rt = vcat(fill(1.1, 2 * 7), fill(2.0, 2 * 7), fill(0.5, 2 * 7),
     fill(1.5, 2 * 7), fill(0.75, 2 * 7), fill(1.1, 6 * 7)) |>
           Rt -> [Rt;
                  [1.0 + A * sin.(2 * π * (t - ϕ) / P) for t in 1:(N - length(Rt))]]
+
+# Common inference tspan: Length of true Rt less three weeks
+inference_tspan = (1, N - 21)
--- a/pipeline/scripts/common_param_values.jl
+++ b/pipeline/scripts/common_param_values.jl
@@ -1,3 +1,11 @@
+@info("""
+      Setting common parameter in `Main` module.
+      ---------------------------------------------
+      Currently active project is: $(projectname())
+      Path of active project: $(projectdir())
+
+      """)
+
 # Common generation interval values
 gi_means = [2.0, 10.0, 20.0];
 gi_stds = [2.0];
--- a/pipeline/scripts/common_param_values.jl
+++ b/pipeline/scripts/common_param_values.jl
@@ -9,3 +9,17 @@
 # Common generation interval values
 gi_means = [2.0, 10.0, 20.0];
 gi_stds = [2.0];
+
+#Common thread count
+num_threads = min(10, Threads.nthreads());
+
+#True Rt
+
+A = 0.3
+P = 30.0
+ϕ = asin(-0.1 / 0.3) * P / (2 * π)
+N = 160
+true_Rt = vcat(fill(1.1, 2 * 7), fill(2.0, 2 * 7), fill(0.5, 2 * 7),
+    fill(1.5, 2 * 7), fill(0.75, 2 * 7), fill(1.1, 6 * 7)) |>
+          Rt -> [Rt;
+                 [1.0 + A * sin.(2 * π * (t - ϕ) / P) for t in 1:(N - length(Rt))]]
--- a/pipeline/scripts/common_param_values.jl
+++ b/pipeline/scripts/common_param_values.jl
@@ -6,22 +6,9 @@
 
       """)
 
-using EpiAware
 # Common generation interval values
 gi_means = [2.0, 10.0, 20.0];
 gi_stds = [2.0];
-
-#Common thread count
-num_threads = min(10, Threads.nthreads());
-
-#Common inference method
-inference_method = EpiMethod(
-    pre_sampler_steps = [ManyPathfinder(nruns = 4, maxiters = 100)],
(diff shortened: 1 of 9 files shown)


pipeline/scripts/common_param_values.jl
```diff
@@ -1,3 +1,37 @@
+@info("""
+      Setting common parameter in `Main` module.
+      ---------------------------------------------
+      Currently active project is: $(projectname())
+      Path of active project: $(projectdir())
+
```
seabbs, line 6: This script now feels a bit poorly named and I am a little uncomfortable with the fact it is rerun whenever it is used vs being run once and then used? I think we can address this in another overall issue for how we construct these pipelines

pipeline/scripts/common_scenarios.jl
```diff
@@ -0,0 +1,41 @@
+using DrWatson
+@quickactivate "Analysis pipeline"
+
+include(srcdir("AnalysisPipeline.jl"))
+include(scriptsdir("common_param_values.jl"))
```
seabbs: i.e here this means we have random stuff floating in our global environment some of which is used and some of which is not which is not great practice?

pipeline/scripts/common_scenarios.jl
```diff
@@ -0,0 +1,41 @@
...
+
+rw = RandomWalk(
+    std_prior = underlying_std_prior, init_prior = transformed_process_init_prior)
+
+diff_ar = DiffLatentModel(; model = ar, init_priors = [transformed_process_init_prior])
+
+wkly_ar, wkly_rw, wkly_diff_ar = [ar, rw, diff_ar] .|>
+                                 model -> BroadcastLatentModel(model, 7, RepeatBlock())
+
+naming_scheme = Dict(
```
seabbs, line 34: can't you use the. tooling from `DrWatson` to do this a bit more slickly?

pipeline/scripts/generate_truth_data.jl
```diff
@@ -16,14 +16,6 @@ include(scriptsdir("common_param_values.jl"));
 ## Set up the truth Rt and save a plot of it
 
 using .AnalysisPipeline, Plots, JLD2
```
seabbs, line 18: why does this still exist or converly why did it get changed? It seems like having a script that generates and saves this is a good idea? 

I am getting concerned about the lack of check pointing in this workflow FYI.

pipeline/scripts/inference_pipeline.jl
```diff
@@ -1,38 +1,46 @@
 using DrWatson
 @quickactivate "Analysis pipeline"
 
+# Include the AnalysisPipeline module
 include(srcdir("AnalysisPipeline.jl"))
+# Include the common parameter values
 include(scriptsdir("common_param_values.jl"))
+# Include the common inference scenarios
+include(scriptsdir("common_scenarios.jl"))
```
seabbs, line 8: As above I think we need an issue to discuss other pipeline approaches that are not script based sourcing as this is probably the least stable/ good practice option of those we could choose imo

pipeline/scripts/common_param_values.jl
```diff
@@ -1,3 +1,37 @@
+@info("""
```
seabbs, line 1: A small nitpick but if two things are called common why aren't they in a directory called common?

pipeline/scripts/inference_pipeline.jl
```diff
@@ -1,38 +1,46 @@
...
 
-## Parameter settings
-# Rolled out to a vector of inference configurations using `dict_list`.
-sim_configs = Dict(:igp => [DirectInfections, ExpGrowthRate, Renewal],
-    :latent_model => [wkly_ar, wkly_rw, wkly_diff_ar],
-    :gi_mean => gi_means, :gi_std => gi_stds) |> dict_list
+## Inference methods
+using Distributions, .AnalysisPipeline, EpiAware, JLD2, ADTypes, AbstractMCMC
+
+for filename in readdir(datadir("truth_data"))
```
seabbs, line 21: This is where dagger comes in? Or we would do a dagger based overhaul of the whole pipeline?

pipeline/scripts/inference_pipeline.jl
```diff
@@ -1,38 +1,46 @@
...
+            case_data = D["y_t"],
+            tspan = inference_tspan,
+            epimethod = inference_method
+        )
+        prfx = "observables" * "_igp_" * string(d[:igp]) * "_latentmodel_" *
+               naming_scheme[d[:latent_model]] * "_truth_gi_mean_" *
+               string(truth_data_gi_mean)
+
+        data, file = produce_or_load(
+            simulate_or_infer, config, datadir("epiaware_observables"); prefix = prfx)
```
seabbs, line 43: as we now have completely different tooling for simulation and inference I am not clear what the role of a joint function (as per #208). I guess if there is a single generating script it makes sense but does that actually help us?

pipeline/src/InferenceConfig.jl
```diff
@@ -0,0 +1,97 @@
...
+- `config::InferenceConfig`: The configuration object containing the case data
+to make inference on and model configuration.
+
+# Returns
+- `inference_results`: The results of the simulation or inference.
+
+"""
+function simulate_or_infer(config::InferenceConfig)
+    #Define infection-generating model
+    shape = (config.gi_mean / config.gi_std)^2
```
seabbs, line 69: why doesn't this happen just once per GI scenario and make a EpiData struct that we pass around?

pipeline/src/InferenceConfig.jl
```diff
@@ -0,0 +1,97 @@
...
+    gen_distribution = Gamma(shape, scale)
+
+    #Define the infection-generating process
+    model_data = EpiData(gen_distribution = gen_distribution, D_gen = config.D_gen,
+        transformation = config.transformation)
+
+    epi = config.igp(model_data, config.log_I0_prior)
+
+    #Define the infection conditional observation distribution
+    obs = LatentDelay(
```
seabbs, line 80: Why aren't we defining this externally as a joining model specification?

pipeline/src/InferenceConfig.jl
```diff
@@ -0,0 +1,97 @@
...
+        NegativeBinomialError(cluster_factor_prior = config.cluster_factor_prior),
+        config.delay_distribution; D = config.D_obs)
+
+    #Define the EpiProblem
+    epi_prob = EpiProblem(epi_model = epi,
```
seabbs, line 85: I think I was expecting to see a bunch of EpiProblems being defined and then apply_method dispatched on those vs doing it like this? It feels like rehashing the EpiAware infra?

seabbs (approved):
This is good for a first pass and in line with the rest of the implementation. 

I am a bit uncomfortable with all the script sourcing we are depending on, the amount of apparently duplicate infra, and the lack of use of checking pointing or pipeline dependency tools which I was sort of expecting to see. 

I think all of above can be addressed as new issues and may need some discussion

PR #217, opened 2024-05-10
refactor pipeline smaller scripts and functions

### LINKED ISSUE #215: Refactor pipeline: smaller scripts and functions
After a f2f chat, we're merging this PR since it does do the job as exists and generating issues to make it more robustly extensible in future refactors (as well as completing other tasks from the meta-issue #195 ).

_Originally posted by @SamuelBrand1 in https://github.com/CDCgov/Rt-without-renewal/issues/213#issuecomment-2102733434_
            
The easiest solution to questions about the robustness of the current analysis pipeline is to refactor the current block of scripts into smaller single job scripts: this will be easier to reason on with a task scheduler (e.g. Dagger.jl).

SamuelBrand1:
This PR aims to refactor, but not extend the functionality, of the analysis pipeline in such a way that:

1. It is more readable.
2. Its closer to DRY principles.
3. Its written in such a way that a next step for job scheduling is as easy as possible.

Concretely,

- Default variables are generated by calling a default function from the `AnalysisPipeline` module.
- Vectors of scenarios combinations are generated by calling `make_...` functions from the `AnalysisPipeline` module.
- Plotting functions are declared separately; I don't feel they belong inside the analysis module but open to changing this.
- The pipeline is now one script, which essentially follows an analogous sequence of steps to the opening example of `targets.R`.
- Inference scenarios are an inner loop of the the "truth data" generation process.
- Each function has a unit test, as does the whole pipeline for a randomly selected inference scenario on a short flat time series of fake data.

Closes #215 .


pipeline/plots/plot_functions.jl
```diff
@@ -0,0 +1,49 @@
+"""
```
seabbs, line 1: I think I would put these in the analysis module as this is part of the analysis and we don't plan to ship it? Alternatively I suppose we could have them as part of a less generic paper module or something (and so the pipeline would aim to be generic and the paper would aim to have the specific implementations for this work in it?) that might be a bit extra but also might not be that much work to get some quite clean code?

pipeline/plots/plot_functions.jl
```diff
@@ -0,0 +1,49 @@
...
+Plot and save the plot of the true Rt values over time.
+
+# Arguments
+- `true_Rt`: An array of true Rt values.
+
+# Returns
+- `plt_Rt`: The plot object.
+
+"""
+function plot_Rt(true_Rt)
```
seabbs, line 39: something I don't love about this is that it gives no indication which packages are needed to run these functions (above module might help with this?)

pipeline/scripts/analysis_pipeline.jl
```diff
@@ -0,0 +1,85 @@
...
+include(plotsdir("plot_functions.jl"));
+
+@info("""
+      Running the analysis pipeline.
+      ---------------------------------------------
+      Currently active project is: $(projectname())
+      Path of active project: $(projectdir())
+      """)
+
+using .AnalysisPipeline
```
seabbs, line 21: i.e under the pipeline is generic this would become `EpiAwarePipeline` and we would also have `RtwithoutRenewalAnalysis` with the specific scenario grid and plots etc?

seabbs: looking at below though I like it as a "default" so perhaps those would always stay in the pipeline module?

pipeline/scripts/analysis_pipeline.jl
```diff
@@ -0,0 +1,85 @@
...
+latent_models_names = Dict(value => key for (key, value) in latent_models_dict)
+tspan = default_tspan()
+inference_method = default_inference_method()
+
+# truth data configurations (e.g. different GI means) and inference configurations
+# (e.g. different infection generation processes and latent models etc.)
+truth_data_configs = make_truth_data_configs(
+    gi_means = default_gi_param_dict["gi_means"], gi_stds = default_gi_param_dict["gi_stds"])
+
+inference_configs = make_inference_configs(
```
seabbs, line 46: In my head this should comee after all the truth data stuff so its a clear task block?

pipeline/scripts/analysis_pipeline.jl
seabbs: also if it didn't it would be very full as its mostly defaults :)

pipeline/scripts/analysis_pipeline.jl
```diff
@@ -0,0 +1,85 @@
...
+    gi_means = default_gi_param_dict["gi_means"],
+    gi_stds = default_gi_param_dict["gi_stds"])
+
+truth_data_config_structs = map(truth_data_configs) do d
+    TruthSimulationConfig(
+        truth_process = true_Rt, gi_mean = d["gi_mean"], gi_std = d["gi_std"])
+end
+
+# Produce and save the truth data
+truth_data_scenarios = map(truth_data_config_structs) do truth_data_config_struct
```
seabbs, line 57: its not a lot of code but in my head we would start pushing this stuff into functions in the analysis pipeline?

pipeline/scripts/analysis_pipeline.jl
```diff
@@ -0,0 +1,85 @@
...
+# Analogy: tar_option_set(...) in targets
+using JLD2, Plots
+
+## Run the pipeline steps
+# Analogy: list(tar_targets...) followed by tar_make(...) in targets
+# This is an intermediate commit that runs but will be updated with a job
+# scheduler for these tasks in the next commit. Probably Dagger.jl
+
+# Default parameter values and plot true Rt
+default_gi_param_dict = default_gi_params()
```
seabbs, line 33: could we be really fancy and have a pipeline type and then dispatch on standard function names to make it really easy to overload parts of the pipeline? Then to change the analysis you would just change the target type and have a write your own new methods where you wanted?

pipeline/src/default_Rt.jl
```diff
@@ -0,0 +1,20 @@
...
+
+## keyword Arguments
+- `A`: Amplitude of the sinusoidal variation in Rt. Default is 0.3.
+- `P`: Period of the sinusoidal variation in Rt. Default is 30 time steps.
+
+## Returns
+- `true_Rt`: Array of default Rt values over time.
+
+"""
+function default_Rt(; A = 0.3, P = 30.0)
```
seabbs, line 12: under the type dispatch idea this would be come i.e `generate_Rt(<type>; stuff)`.

pipeline/src/default_priors.jl
```diff
@@ -0,0 +1,20 @@
+"""
+Constructs a dictionary of default prior distributions for the parameters used
+    in `EpiAware` models.
+
+# Returns
+- `Dict{String, Distribution}`: A dictionary containing the default prior
+    distributions.
+
+"""
+function default_priors()
```
seabbs, line 10: can we call this something a little clearer. its just shared latent model priors right?

seabbs (approved):
I really like this. I had a few suggestions to make it more complicated/have more infra but not essential. We can discuss and port what we want to do into new issues and merge this I think

pipeline/scripts/analysis_pipeline.jl
SamuelBrand1: I agree. Done this with obvious tests (and change to end-to-end test)

pipeline/scripts/analysis_pipeline.jl
SamuelBrand1: Yep

SamuelBrand1:
I've made a few minor changes:

- More small(ish) functions with tests
- plotting functions to inside `AnalysisPipeline`
- moved some struct construction to inside a generate function to make the pipeline script a bit cleaner.

#218 Should cover some of the more exciting ideas we could do with pipeline analysis going forwards! But I think this pipeline (which runs) is currently a good template to layer on the rest the tasks in #201

seabbs (approved):

PR #220, opened 2024-05-14
Implementing `Dagger.@spawn` for pipeline jobs

### LINKED ISSUE #219: Blend `Dagger.jl` checkpointing into the scenario inference loop process, e.g. like here.

SamuelBrand1:
This PR rewrites the main pipeline script to use [`Dagger.jl`](https://github.com/JuliaParallel/Dagger.jl) to execute the analysis as a [DAG of spawned tasks](https://en.wikipedia.org/wiki/Directed_acyclic_graph#Scheduling).

Notable changes/decisions:
- The string names of each model now have a default constructor function.
- Nearly all function calls are first created as a task using `Dagger.@spawn`, with results fetched as required or passing `EagerThunk` objects to downstream tasks to get executed in a schedule determined by `Dagger`. Not all of this is necessary (could be replaced by e.g. `@everywhere ...some code...`), but maintains the pattern and should be more extensible.
- Construction of the list of scenarios is not a `Dagger` generated task; I'm viewing this a being created by an orchestrating main node which then implicitly defines the DAG topology. Happy to get pushed back on this.
- The default inference is now serial with 2 sequentially executed MCMC chains. I'm assuming that we'll aim for deploying resources to parallelise across tasks rather than within tasks.
- **Checkpointing.** I've checked that doing the inference as a set of tasks is compatible with the checkpointing provided by `DrWatson.produce_or_load`. `Dagger` can provide more methods of checkpointing, but at the moment saving to disk and recognising that a task has been completed by the existence of a saved file works. However, we should have a think about vulnerability to an edge case causing an error?

Closes #219 .

### DIFF
--- a/pipeline/scripts/analysis_pipeline.jl
+++ b/pipeline/scripts/analysis_pipeline.jl
@@ -3,10 +3,8 @@ using DrWatson
 
 # Activate the project environment
 # Analogy: source(functions.R) in targets
-@quickactivate "Analysis pipeline"
-
-# Include the AnalysisPipeline module
-include(srcdir("AnalysisPipeline.jl"));
+quickactivate(@__DIR__(), "Analysis pipeline")
+using Dagger
 
 @info("""
       Running the analysis pipeline.
@@ -17,7 +15,18 @@ include(srcdir("AnalysisPipeline.jl"));
 
 ## Other dependencies
 # Analogy: tar_option_set(...) in targets
-using .AnalysisPipeline
+# Add processes for parallel computing and ensure all have same
+# dependencies/environment
+using Distributed
+addprocs(1)
+
+@everywhere begin
+    using DrWatson
+    quickactivate(@__DIR__(), "Analysis pipeline")
+    include(srcdir("AnalysisPipeline.jl"))
+end
+
+@everywhere using .AnalysisPipeline
 
 ## Run the pipeline steps
 # Analogy: list(tar_targets...) followed by tar_make(...) in targets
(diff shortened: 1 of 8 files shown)


pipeline/scripts/analysis_pipeline.jl
```diff
@@ -17,50 +15,70 @@ include(srcdir("AnalysisPipeline.jl"));
...
+#     default_gi_param_dict = default_gi_params()
+#     true_Rt = default_Rt()
+#     plt_Rt = plot_Rt(true_Rt)
+#     latent_models_dict = default_epiaware_models()
+#     latent_models_names = Dict(value => key for (key, value) in latent_models_dict)
+#     tspan = default_tspan()
+#     inference_method = default_inference_method()
+# end
+
+default_gi_param_dict_thunk = Dagger.@spawn default_gi_params()
```
seabbs, line 47: I assume they have exported `@spawn` and so we need to write `Dagger.@spawn`? Otherwise it is a bit clunky?

seabbs (approved):
The implementation looks good. Just need to discuss your points and good to go

pipeline/scripts/analysis_pipeline.jl
```diff
@@ -17,50 +15,70 @@ include(srcdir("AnalysisPipeline.jl"));
...
 
 # Default parameter values and plot true Rt
-default_gi_param_dict = default_gi_params()
-true_Rt = default_Rt()
-plt_Rt = plot_Rt(true_Rt)
-latent_models_dict = default_epiaware_models()
-latent_models_names = Dict(value => key for (key, value) in latent_models_dict)
-tspan = default_tspan()
-inference_method = default_inference_method()
+# @everywhere begin
```
seabbs, line 37: Holding comment to stop auto-merge whilst we discuss

seabbs:
> ), but maintains the pattern and should be more extensible.

I think taking this approach makes a lot of sense as we get a complete task DAG of the analysis.

>  I'm viewing this a being created by an orchestrating main node which then implicitly defines the DAG topology. Happy to get pushed back on this.

I see this argument. In `targets` I would still have this as part of the workflow potentially but I would make it happen on the local node (that might be hard here). I am happy sticking with the current approach I think.

pipeline/scripts/analysis_pipeline.jl
```diff
@@ -17,50 +15,70 @@ include(srcdir("AnalysisPipeline.jl"));
...
+#     plt_Rt = plot_Rt(true_Rt)
+#     latent_models_dict = default_epiaware_models()
+#     latent_models_names = Dict(value => key for (key, value) in latent_models_dict)
+#     tspan = default_tspan()
+#     inference_method = default_inference_method()
+# end
+
+default_gi_param_dict_thunk = Dagger.@spawn default_gi_params()
+true_Rt_thunk = Dagger.@spawn default_Rt()
+plt_Rt_thunk = Dagger.@spawn plot_Rt(true_Rt_thunk)
```
seabbs, line 49: do you not want to be fetching the thunk here?

pipeline/scripts/analysis_pipeline.jl
SamuelBrand1: Yeah `@spawn` is used by `Threads` e.g. `Threads.@spawn` which is exposed from `Base.Threads`, i.e. it is available from `Main` for every Julia . 

I'm assuming that the style of naming `Dagger.@spawn` throughout the `Dagger` docs, which I've propagated here is due to this.

pipeline/scripts/analysis_pipeline.jl
```diff
@@ -17,50 +15,70 @@ include(srcdir("AnalysisPipeline.jl"));
...
+latent_models_dict_thunk = Dagger.@spawn default_epiaware_models()
+latent_models_names_thunk = Dagger.@spawn default_latent_models_names()
```
seabbs, line 51: Do we really need to crowd the naming with `_thunk` if its just going to be used as is? Maybe we assume everything is a funk and have a different name for actualised ones (is the dicts below)

PR #222, opened 2024-05-14
splits simulate_or_infer into simulate and infer plus updated unit tests

### LINKED ISSUE #208: simulate_or_infer or simulate and infer

SamuelBrand1:
As per title, and after discussion #208 

Closes #208 .


seabbs (approved):
LGTM

PR #223, opened 2024-05-15
Rolling out a pipeline type system for dispatching on analysis details

### LINKED ISSUE #218: Rolling out a Pipeline type system for dispatching on analysis details
could we be really fancy and have a pipeline type and then dispatch on standard function names to make it really easy to overload parts of the pipeline? Then to change the analysis you would just change the target type and have a write your own new methods where you wanted?

_Originally posted by @seabbs in https://github.com/CDCgov/Rt-without-renewal/pull/217#discussion_r1598153887_
            
I think this is a really neat idea. I'm not implementing this in PR #217 but I think we should keep this idea going.

SamuelBrand1:
This PR creates a simple type system for dispatching on with `make_pipeline`, although for this PR only one default is supplied.

The two component functions to the pipeline are `make_truthdata` which spawns data generating tasks with option `cache=true` (i.e. thunk caches its results) and `make_inference` which spawns inference tasks. These wrap into `make_pipeline`.

A nice feature is that `Dagger` uses the available resources to complete the spawned tasks in the pipeline. In `scripts/run_pipeline.jl` I set up for distributed, but if you neglect this and run with threads it will do threaded (it can also do both distributed and threaded).

Closes #218


SamuelBrand1:
We are closing this PR to rewrite using the lazy API of `Dagger.jl`

PR #228, opened 2024-05-20
prior predictive modelling for each inference model

### LINKED ISSUE #227: Prior predictive modelling for each inference model

SamuelBrand1:
This PR adds a pipeline for generating prior predictive samples from the list of inference model types as discussed in #227 . The approach is to define a new `<: AbstractEpiAwarePipeline` type `RtwithoutRenewalPriorPipeline` which dispatches a method for:
1. Only passing `missing` data to the inference problem.
2. Uses `DirectSample` to sample from the model.

Also adds some tests for this and a pipeline script.

Closes #227 .

### DIFF
--- a/pipeline/data/epiaware_observables/placeholder.md
+++ b/pipeline/data/epiaware_observables/placeholder.md
@@ -1,0 +1,1 @@
+# placeholder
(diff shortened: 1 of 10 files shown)


seabbs (approved):
Really nice. Once we are clear on all the sub workflows we may want to make a meta one that does everything (which suggests that the current `RtwithoutRenewal` may need to be renamed to `RtwithoutRenewalInference` etc.

Issue #209, opened 2024-05-03
Setup CI for analysis pipeline


seabbs:
As discussed in #202 we should set up CI in `pipeline` and make sure that `EpiAware` CI only runs when `EpiAware` code is updated? It seems like for the most part we would want all the CI we have for `EpiAware` to be implemented as well for `pipeline` (I think its fine to duplicate this vs doing clever things to avoid that). 

We would also need to properly setup testing in `pipeline` for this (so this may related to #209).

SamuelBrand1:
And vice versa CI for `EpiAware` doesn't run when we have only changed the pipeline.

Issue #206, opened 2024-05-03
Work out how to use @testitem in pipeline


seabbs:
We have been using @testitem to better integrate our testing suite in vscode. However, in #204 there was some concern that this would make it hard to keep our analysis code (i.e in `pipeline`) tests apart from our tooling tests (in `EpiAware`) so in that code base we have fallen back to using `@testitem`.

We ideally don't want to do this so should investigate ways of using `@testitem` but still being able to keep the testing suites apart.

SamuelBrand1:
I've been looking at this. This is my current understanding:

Its pretty easy to have a test suite for the analysis pipeline provided by the base Julia package manager `Pkg`, the steps to do this are:
- Rename the pipeline environment `AnalysisPipeline` to match the module name
- provide an `uuid` for the env project file.

Then the pipeline tests can be run as 

```julia
(AnalysisPipeline) pkg> test
```

as per usual.

It seems less easy to include in the VS code live tests because that wants a single link to a `runtests.jl` and (I think) will execute in the `Test` env of the `EpiAware` project as we've set it up now.

seabbs:
I'm not totally clear about this as I would have thought that it worked on a folder level

seabbs:
I have been working at the root of the `Rt-without-renewal` dir but I guess if you are in the working directory of `EpiAware` and `pipleine` that should be treated as its own project? If that isn't the case you couldn't have a folder of Julia packages that all used @testitem?

PR #229, opened 2024-05-20
Change AnalysisPipeline into "package"-like project

SamuelBrand1:
This is a small PR for upgrading `AnalysisPipeline` into a "package"-like project, e.g. a Julia environment with some package-like qualities.

The main advantages are:
- no need to `include` the `src` code of the pipeline from scripts in addition to setting the env.
- `using AnalysisPipeline` rather than the clunkier `using .AnalysisPipeline`
- can use
> `(AnalysisPipeline) pkg> test`

to run unit tests.

This contributes on issues #206 and #209 but doesn't close them.

PS for the pipeline I don't think we need a separate test environment for this PR but might need one in the future (e.g. `TestItemRunner` dep etc)


seabbs (approved):
Great idea

PR #235, opened 2024-05-22
add `get_param_array` function

### LINKED ISSUE #234: Utility for translating a `MCMCChains.Chain` object into a vector of `NamedTuples`
I don't think this utility exists (see https://github.com/TuringLang/MCMCChains.jl/issues/456#issue-2310168856 ), but it would be useful for addressing #231 .

This issue is about adding this utility to `EpiAware`, but with knowledge that we will deprecate this in favour of a `MCMCChains` utility if that comes about (or already exists and we haven't noticed it).

SamuelBrand1:
This PR adds a function to `EpiAwareUtils` which maps a `MCMCChains.Chains` type object with `n` samples per `m` chains (i.e. `n * m` samples) to a `(n, m)` sized array of `NamedTuple`s representation of each sample.

Closes #234 .


SamuelBrand1:
I've changed this following this suggestion https://github.com/TuringLang/MCMCChains.jl/issues/456#issuecomment-2124634019 

Its a really simple utility now.

seabbs (approved):
Really nice.

PR #236, opened 2024-05-28
Add `RtwithoutRenewalTestPipeline` type with updated pipeline tests

SamuelBrand1:
This small PR follows on from f2f discussion and adds a test mode pipeline type to differentiate between sampling done running tests vs full pipeline runs.


pipeline/src/EpiAwarePipeline.jl
```diff
@@ -15,7 +15,7 @@ using CSV, Dagger, DataFramesMeta, Dates, Distributions, DocStringExtensions, Dr
 
 # Exported pipeline types
 export AbstractEpiAwarePipeline, EpiAwarePipeline, RtwithoutRenewalPipeline,
-       RtwithoutRenewalPriorPipeline
+       RtwithoutRenewalPriorPipeline, RtwithoutRenewalTestPipeline
```
seabbs, line 18: I thought the plan was to move towards EpiAwarePipeline. Based on that idea shouldn't this be `EpiAwareTestPipeline` or `EpiAwareTodyPipeline`?

seabbs (approved):
Only comment is about the naming

pipeline/src/EpiAwarePipeline.jl
SamuelBrand1: Yes, good point!

seabbs:
Stochastic failure is negative binomial which is an epiaware issue. Aside from updating the name this is good to go.

Issue #244, opened 2024-05-30
Bounded population size as default


SamuelBrand1:
Another more radical possibility is that we accept that having models that _can_ sample > `1e17` infections (which is easier to do in unbounded pops than one might expect due to RW $\log R_t$ and the magic of exponential growth) are bad and enforce a max population size. Then these warm up phase problems should disappear in a more principled way.

_Originally posted by @SamuelBrand1 in https://github.com/CDCgov/Rt-without-renewal/issues/237#issuecomment-2139307623_
           
We haven't done this because we are much more focused on other factors in the analysis, but it would help the sampling during edge cases in warm up.

seabbs:
I don't think we should have this as a default but I do think we should have a clear path to making this a composable feature and other similar ideas like feedbacking Rt based on incidence etc that proxy this.

SamuelBrand1:
> I don't think we should have this as a default but I do think we should have a clear path to making this a composable feature and other similar ideas like feedbacking Rt based on incidence etc that proxy this.
> 
> In my head this would either be a generalisation of the current `Renewal` infection model to multiple more complex types or it would be making this somehow specifiable in the way we are handling latent models and observation models. The tricky bit here is that you need to loop over infections to do this in most (nearly all?) cases and so nesting models based on vector passing is tricky

My vote is for generalisation of the infection generating process (igp) model. I think we set the behaviour of the igp by the type, then it has its own callable for generating the infections. The trick here is to create an abstract `Renewal` type for `generate_latent_infs` to dispatch on with the specific renewal models (e.g. no population, infection reduced by fraction of remaining pop, recent feedback) behaviour in the callable.

seabbs:
After this PR my concerns remain about stacking multiple processes

PR #245, opened 2024-05-30
add `RenewalWithPopulation` infection generating process model

SamuelBrand1:
This draft PR is to demonstrate a possible way of doing models with population size. This is aimed at informing problems found in #237 and discussion #244 . If this is a good approach then its easy to extend.

Features:

- Adds an `AbstractRenewal <: AbstractTuringEpiModel` type. This allows a generic form of `generate_latent_infs` which dispatches on any renewal model.
- Adds a `RenewalWithPopulation <: AbstractRenewal` struct with a pop-size argument.
- A callable on this struct type gives the discrete time dynamics and passes along available number of susceptibles.
- The effective Rt is scaled by $\max(S_{t-1} / N, 0)$, to avoid negs. There are slicker ways to do this but it is a first pass.
- Trials converting the `getting_started` example into using this model along with `adtype = AutoReverseDiff(true)`... This seems to work very well and the sampling is about 2x faster (always a good sign!).

**Doesn't have extended unit tests! But no existing code is broken.**


EpiAware/src/EpiInfModels/EpiInfModels.jl
```diff
@@ -10,7 +10,7 @@ using ..EpiAwareUtils: scan, censored_pmf
 using Turing, Distributions, DocStringExtensions, LinearAlgebra
 
 #Export models
-export EpiData, DirectInfections, ExpGrowthRate, Renewal
+export EpiData, DirectInfections, ExpGrowthRate, Renewal, RenewalWithPopulation
```
seabbs, line 13: Maybe it should be `PopulationBoundedRenewal`?

EpiAware/src/EpiInfModels/Renewal.jl
```diff
@@ -111,6 +116,166 @@ function (epi_model::Renewal)(recent_incidence, Rt)
...
+- `epi_model::RenewalWithPopulation`: The renewal model with population.
+- `I₀`: The initial number of infected individuals.
+- `Rt`: The time-varying reproduction number.
+
+# Returns
+- A tuple containing the initial number of infected individuals at each generation
+interval and the population size of susceptible/available people.
+
+"""
+function make_renewal_init(epi_model::RenewalWithPopulation, I₀, Rt)
```
seabbs, line 273: This duplication is what worries me. Shouldn't this just be based on AbstractRenewal?

EpiAware/src/EpiInfModels/Renewal.jl
```diff
@@ -111,6 +116,166 @@ function (epi_model::Renewal)(recent_incidence, Rt)
...
+- `recent_incidence_and_available_sus`: A tuple with an array of recent incidence
+values and the remaining susceptible/available individuals.
+- `Rt`: Reproduction number.
+
+# Returns
+- Tuple containing the updated incidence array and the new `recent_incidence_and_available_sus`
+value.
+"""
+function (epi_model::RenewalWithPopulation)(recent_incidence_and_available_sus, Rt)
+    recent_incidence, S = recent_incidence_and_available_sus
```
seabbs, line 250: am I missing something here?

seabbs: where is this used?

EpiAware/src/EpiInfModels/Renewal.jl
```diff
@@ -77,7 +82,7 @@ unobserved infections.
 I_t = generated_quantities(latent_inf, θ)
 ```
 "
-@kwdef struct Renewal{S <: Sampleable} <: EpiAwareBase.AbstractTuringEpiModel
+@kwdef struct Renewal{S <: Sampleable} <: AbstractRenewal
```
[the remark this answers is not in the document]
SamuelBrand1: Do we want this type hierarchy?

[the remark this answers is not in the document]
seabbs: I mean I think so? Is there a reason not to?

EpiAware/src/EpiInfModels/Renewal.jl
seabbs: oh I see in scan

EpiAware/src/EpiInfModels/Renewal.jl
SamuelBrand1: No, the different models are based on different states so get different `init`s

This model has a state $((I_{t-1}, I_{t-2}, ..., I_{t-m}), S_{t-1})$

EpiAware/src/EpiInfModels/Renewal.jl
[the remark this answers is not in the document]
SamuelBrand1: Cool, then yes in base.

seabbs (approved):
Okay I was a bit lost as the work is being done in scene right but this now seems really neat. My only push back is that make_renewal_init seems like it should be a abstract method.

EpiAware/src/EpiInfModels/Renewal.jl
seabbs: I see... can we not nest this better?

SamuelBrand1: Obviously, we can get rid of this by looking at whole infection time series but I don't like that (very inefficient) but I can be persuaded

SamuelBrand1:
From this and f2f chat I think we want to move this towards something that is more similar to the observation models e.g. recursive dispatch downwards, not using callables etc

EpiAware/src/EpiInfModels/Renewal.jl
```diff
@@ -111,6 +116,166 @@ function (epi_model::Renewal)(recent_incidence, Rt)
...
+values and the remaining susceptible/available individuals.
+- `Rt`: Reproduction number.
+
+# Returns
+- Tuple containing the updated incidence array and the new `recent_incidence_and_available_sus`
+value.
+"""
+function (epi_model::RenewalWithPopulation)(recent_incidence_and_available_sus, Rt)
+    recent_incidence, S = recent_incidence_and_available_sus
+    new_incidence = max(S / epi_model.pop_size, 0.0) * Rt *
```
seabbs, line 251: do we not want to use the correction from here: https://academic.oup.com/jrsssa/article/186/4/601/7060037

EpiAware/src/EpiInfModels/Renewal.jl
seabbs: yeah don't love that either. Looking at this its just that one has I_0 at the end right? So why can't the internals be otherwise shared?

seabbs:
I mean maybe but I really like this all depending on scan which is otherwise called from an abstract method. So this would imply scan should dispatch on a stuct and use a standard function? Is that really preferable?

EpiAware/src/EpiInfModels/Renewal.jl
SamuelBrand1: Thats an option, there are other constructions, but yeah we don't want 

```julia
max(S / epi_model.pop_size, 0.0)
```

as the main model!

seabbs: cool shall we have this for now and upgrade using an issue?

SamuelBrand1: Yeah, the way I've got this model a (large) population actually gets depleted. I don't think the linked paper actually does depletion ($S_0$ seems fixed), which might well be fine in many contexts (e.g. infer $S_0$) and other models have history dependent feedback. The idea is that we can easily implement all of those (and indeed naturally extend to other igps).

SamuelBrand1:
> From this and f2f chat I think we want to move this towards something that is more similar to the observation models e.g. recursive dispatch downwards, not using callables etc

I feel we are moving towards committing this structure; I'm feeling a love for callables is growing!

I'm going to extend test coverage to new functions (and maybe make and then we can have issues for:

- How to make as many functions as possible maximally generic.
- Range of models we want for population adjustment.
- Extension to other igps: e.g. exponential growth

SamuelBrand1 (commented):
> Maybe it should be `PopulationBoundedRenewal`? 

Maybe that should be a supertype? Reasoning being that there are several `PopulationBoundedRenewal` approaches.

SamuelBrand1 (commented):
> turing specific again? 

I think we are going to go this way, but atm thats not really true retrospectively.

EpiAware/src/EpiInfModels/EpiInfModels.jl
SamuelBrand1: I feel thats a super-type if/when we start adding more models.

seabbs:
Yes agree. I think we should use this for now but plan for a breaking change that generalises this. 

Good point that it isn't currently turing specific. Happy to see this made into a full pr and to review.

SamuelBrand1:
> Good point that it isn't currently turing specific. 

I've gone this way in anticipation that it will be soon (e.g. using `@submodel` to sample pop susceptibility params).

seabbs:
Why no review :(

EpiAware/src/EpiInfModels/Renewal.jl
```diff
@@ -111,6 +111,166 @@ function (epi_model::Renewal)(recent_incidence, Rt)
...
+- `epi_model::Renewal`: The renewal model.
+- `I₀`: The initial number of infected individuals.
+- `Rt`: The time-varying reproduction number.
+
+# Returns
+The initial vector of infected individuals.
+
+"""
+function make_renewal_init(epi_model::Renewal, I₀, Rt)
+    r_approx = R_to_r(Rt[1], epi_model)
```
seabbs, line 127: Do we have an issue to solve this issue?

EpiAware/src/EpiInfModels/Renewal.jl
```diff
@@ -111,6 +111,166 @@ function (epi_model::Renewal)(recent_incidence, Rt)
...
+```julia
+# Sample from the unobserved infections model
+
+#Sample random parameters from prior
+θ = rand(latent_inf)
+#Get unobserved infections as a generated quantities from the model
+I_t = generated_quantities(latent_inf, θ)
+```
+"
+@kwdef struct RenewalWithPopulation{S <: Sampleable} <: EpiAwareBase.AbstractTuringRenewal
```
seabbs, line 212: My current thinking is we want one renewal struct that then has a field that takes a struct which dictates the scan function (and maybe the initialisation model?) We can then compose these all together? 

Alternatively we rethink this whole thing (the inf module) as the composition of some later seeding and scan process.

seabbs (commented):
Looks as discussed. Please don't merge in without review when changes have been made! Also as discussed this is strictly temporary and we expect breaking changes to this in the near future.

SamuelBrand1:
That was a mistake!I thought you had accepted it but the draft -> full PR meant auto merge was off.My bad.Sent from my iPhoneOn 31 May 2024, at 15:21, Sam Abbott ***@***.***> wrote:﻿
@seabbs commented on this pull request.

Looks as discussed. Please don't merge in without review when changes have been made! Also as discussed this is strictly temporary and we expect breaking changes to this in the near future.

seabbs:
No worries. We may need to tighten up PR blocks!

PR #249, opened 2024-06-03
Issue 176: Add support for multiple observation models

### LINKED ISSUE #176: Standard approach to multiple group/ multi-signal modelling
## Multi-group modelling

At the moment all our methodology is predicated on one dimensional time series, either infections, cases or driving latent processes. In general, we'll want all of the above to be potentially higher dimensional.

This brings up some issues:
- How to design intuitive API for linking methods for connecting processes of different dimension/size.
- What kind of Julia object are the multi-dimensional objects?

My view on the second issue here is that we should have each process sample paths as something like `Vector{Union{Array{T},T}} where T <: Number` parametric types. The reason is that time index is trivially associated with the index of the vector, but each time point can be an arbitrary Array. I think that this finds a sweet point between generality and convenience and also opens up using utilities like [RecursiveArrayTools.jl](https://github.com/SciML/RecursiveArrayTools.jl).

seabbs:
This PR closes #176 by adding a new `StackObservationModels` which supports arbitary stacks of observation models with both 1 to 1 and 1 to many infection -> obs mappings.

Note that this does not support the passing of unnamed `y_t` in the generated quantities (in the sense of it making no effort to track these) as long term this doesn't seem sustainable - especially due to the lack of integration with `prefix`.

This implies we are expecting posterior predictions to be done on data using the `y_t` (with prefix) sampled at the lowest level of the stack.

 if we did want to keep more in the returns this we would need manually prefix I think. To me this suggests a package wide rethink of the use of `return` is needed.

### DIFF
--- a/EpiAware/src/EpiObsModels/EpiObsModels.jl
+++ b/EpiAware/src/EpiObsModels/EpiObsModels.jl
@@ -9,11 +9,16 @@ using ..EpiAwareUtils: censored_pmf, HalfNormal
 
 using Turing, Distributions, DocStringExtensions, SparseArrays
 
-export PoissonError, NegativeBinomialError, LatentDelay, Ascertainment
+# Observation models
+export PoissonError, NegativeBinomialError
+
+# Observation model modifiers
+export LatentDelay, Ascertainment, StackObservationModels
 
 include("docstrings.jl")
 include("LatentDelay.jl")
 include("Ascertainment.jl")
+include("StackObservationModels.jl")
 include("PoissonError.jl")
 include("NegativeBinomialError.jl")
 include("utils.jl")
--- a/EpiAware/src/EpiObsModels/StackObservationModels.jl
+++ b/EpiAware/src/EpiObsModels/StackObservationModels.jl
@@ -28,15 +28,6 @@ obs_model = generate_observations(obs, y_t, fill(10, 10))
 rand(obs_model)
 samples = sample(obs_model, Prior(), 100; progress = false)
 
-# extract samples for cases.y_t and deaths_y_t
-# from the chain of samples (not using generated_quantities)
-function extract_obs(samples, obs_name)
-    obs = group(samples, obs_name) |>
-          DataFrame |>
-          x -> stack(x, Not(:iteration, :chain))
-    return obs
-end
-
 cases_y_t = group(samples, \"cases.y_t\")
 cases_y_t
 
--- a/EpiAware/src/EpiObsModels/StackObservationModels.jl
+++ b/EpiAware/src/EpiObsModels/StackObservationModels.jl
@@ -1,0 +1,30 @@
+@kwdef struct StackObservationModels{
+    M <: AbstractVector{<:AbstractTuringObservationModel},
+    N <: AbstractVector{<:AbstractString}} <:
+              AbstractTuringObservationModel
+    "A vector of observation models."
+    models::M
+    "A vector of observation model names"
+    model_names::N
+
+    function StackObservationModels(models::Vector{<:M},
+            model_names::Vector{<:N}) where {
+            M <: AbstractTuringObservationModel,
+            N <: AbstractString
+    }
+        @assert length(models)==length(model_names) "The number of models and model names must be equal."
+        new{typeof(models), typeof(model_names)}(models, model_names)
+    end
+
+    function StackObservationModels(models::Vector{<:M}) where {
+            M <: AbstractTuringObservationModel
+    }
+        model_names = [typeof(i) for i in 1:length(models)]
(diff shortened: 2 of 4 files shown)


seabbs:
Implementing this has also made me thing about hierarchical priors. Currently the only way I can think we might do this is similar to this stacking approach but passing in uninstantiated structs and a prior model function and then in the method instantiating the structs using the prior model. 

This feels a bit clunky? As far as I am aware we don't have an issue for this so shall we split out and discuss?

EpiAware/src/EpiObsModels/EpiObsModels.jl
```diff
@@ -9,11 +9,16 @@ using ..EpiAwareUtils: censored_pmf, HalfNormal
 
 using Turing, Distributions, DocStringExtensions, SparseArrays
 
-export PoissonError, NegativeBinomialError, LatentDelay, Ascertainment
+# Observation models
+export PoissonError, NegativeBinomialError
+
+# Observation model modifiers
```
SamuelBrand1, line 15: I like this distinction

EpiAware/src/EpiObsModels/StackObservationModels.jl
```diff
@@ -0,0 +1,112 @@
...
+
+obs = StackObservationModels(
+    (cases = PoissonError(), deaths = NegativeBinomialError())
+)
+y_t = (cases = missing, deaths = missing)
+obs_model = generate_observations(obs, y_t, fill(10, 10))
+rand(obs_model)
+samples = sample(obs_model, Prior(), 100; progress = false)
+
+cases_y_t = group(samples, \"cases.y_t\")
```
SamuelBrand1, line 31:
```suggestion
cases_y_t = group(samples, "cases.y_t")
```

EpiAware/src/EpiObsModels/StackObservationModels.jl
```diff
@@ -0,0 +1,112 @@
...
+cases_y_t
+
+deaths_y_t = group(samples, \"deaths.y_t\")
```
SamuelBrand1, line 34:
```suggestion
deaths_y_t = group(samples, "deaths.y_t")
```

EpiAware/src/EpiObsModels/StackObservationModels.jl
```diff
@@ -0,0 +1,112 @@
...
+
+## Constructors
+
+- `StackObservationModels(models::Vector{<:AbstractTuringObservationModel},
+    model_names::Vector{<:AbstractString})`: Construct a `StackObservationModels` object with a vector of observation models and a vector of model names.
+    - `StackObservationModels(; models::Vector{<:AbstractTuringObservationModel},
+    model_names::Vector{<:AbstractString})`: Construct a `StackObservationModels` object with a vector of observation models and a
+    vector of model names.
+- `StackObservationModels(models::Vector{<:AbstractTuringObservationModel})`:
+    Construct a `StackObservationModels` object with a vector of observation models. The model names are automatically generated from the observation model types.
```
SamuelBrand1, line 14: Do we really want to support this constructor? To get it to work you need data like `(PoissonError = missing, NegativeBinomialError = missing)`

EpiAware/src/EpiObsModels/StackObservationModels.jl
seabbs: Is this a valid docstring? It wasn't rendering for me due to the external "" wrap from @doc raw or was that a late night delusion?

EpiAware/src/EpiObsModels/StackObservationModels.jl
seabbs: Its not totally clear to me. I would be happy to take it out as we are unlikely to want to use it.

EpiAware/src/EpiObsModels/StackObservationModels.jl
SamuelBrand1: Ah right! No you're good, I forgot about the string escape

EpiAware/src/EpiObsModels/StackObservationModels.jl
SamuelBrand1: I would support leaving it out because just passing a vector of models _without_ names and then getting confused would be something I'd expect to do! I'd rather have an error about not having supplied names at the moment of construction

SamuelBrand1 (approved):
I like this! This is really neat.

EpiAware/src/EpiObsModels/StackObservationModels.jl
SamuelBrand1: Saw this done

PR #250, opened 2024-06-03
Issue 183: buffering of latent delays

### LINKED ISSUE #183: Past sequence of latent infections for initial case representation
At the moment, we use an initialisation procedure inside `Renewal` model:

https://github.com/CDCgov/Rt-without-renewal/blob/9a3a519548a9741988d859ab86f81cf26425ef91/EpiAware/src/EpiInfModels/Renewal.jl#L164-L166

This creates (temporarily) a vector of past latent infections based on the sampled initial value of the reproductive number (which is converted into an exponential growth rate $r$).

It might be a good idea to extend this concept to a modular plug-in; in particular this would allow generation of sensible cases early in a time series.

seabbs:
This PR closes #183 by adding support for buffering in all currently supported observation models and in `LatentDelay`.

I am quite unhappy about the current approach because: 

- Requires support in all observation models that will be used with `LatentDelay`
- Has different behaviour when completely missing a vector of missingness is passed in

All that being said I think we should review this as a fix and use any major review comments in a new issue to improve this. One option we could consider is  no longer support passing just `y_t = missing` and instead enforce that this must be a empty vector (and helping people achieve this). Just this change would streamline at least part of this.

### DIFF
--- a/EpiAware/src/EpiObsModels/LatentDelay.jl
+++ b/EpiAware/src/EpiObsModels/LatentDelay.jl
@@ -1,6 +1,9 @@
 @doc raw"
 The `LatentDelay` struct represents an observation model that introduces a latent delay in the observations. It is a subtype of `AbstractTuringObservationModel`.
 
+Note that the `LatentDelay` observation model shortens the observation vector by the length of the delay distribution and this is then passed to the underlying observation model. This is to prevent fitting to partially
+observed data.
+
 ## Fields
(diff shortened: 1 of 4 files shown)


SamuelBrand1 (approved):
This looks good. 

As a note: this PR implies redoing the underlying data generating for the pipeline (which is good and essentially what we are fixing here).

Issue #251, opened 2024-06-03
Aggregate expected obs


seabbs:
To support aggregating obs from daily to other timescales we likely want to have a struct called `Aggregate` that mirrors `Ascertainment` but allows for aggregating expected observations

This would allow for fitting to weekly data etc.

Another option here which would require less code would be to relax te constraint in `LatentDelay` that enforces the static PMF to be proper.

seabbs:
I think specific functionality would be better as for efficiency this would change the time scale of the data (like broadcasting in the latent model)

PR #253, opened 2024-06-03
Issue 51: Add ascertainment_dayofweek

### LINKED ISSUE #51: Inclusion of day of week observation processes

seabbs:
This PR closes #51 by adding a new helper function to `EpiObsModels` that gives semi-sensible defaults for specifying a day of the week ascertainment model (as a day of the week effect).

It does this by introducing a new dependence on `EpiLatentModels` and slightly modifying `Ascertainment` so that the transform has to be manually specified as vectorised (to allow for actually vectorised transforms to be supported).

_Note: We had discussed using a softmax but this doesn't make sense once a latent model has been broadcast. We could instead do so using a `TransformLatentModel` but to me (and as currently formulation is fairly common practice) what is here seems reasonable as a starting point._


SamuelBrand1 (approved):
This looks great!

Wrt using `softmax`, I think that would be cool after we clear #251 e.g. we might model a weekly process but then broadcast out the weekly observations to a day of the week in a way that respects the weekly average obs. 

This implementation hit a different form of modelling and looks nice! 

Note: this was part of analysis plan and should now be added to pipeline.

seabbs:
I made a new issue (#258 for the `softmax` feature)

PR #260, opened 2024-06-05
add softmax transformation to latent process in day of week effects

### LINKED ISSUE #258: Add softmax to broadcast_dayofweek
#253 added a new helper to construct day of the week ascertainment models. 

In the review of that PR (and elsewhere) it was noted that it might make sense to constrain the within week model to sum to 7 using a `LogExpFunctions.softmax`. This makes sense to provide as the default I think. 

To do this we would need to add a `TransformLatentModel` wrapper to the `latent_model` passed to `broadcast_dayofweek`.

SamuelBrand1:
This PR add to the helper function for day of the week (relative) ascertainment by introducing a `7 * softmax` transformation inbetween latent model generation and periodic broadcast.

As discussed in #253 and #258 , the reasoning here is that a null situation for day of the week relative ascertainment is that each day is equally likely to have an asserted cases (conditional on equal numbers of individuals available to be asserted on each day), which corresponds to broadcasting $x = [1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0]$ repeatedly. This PR allows this to vary, but in a way that the $\sum_i x_i = 7.0$ constraint is preserved.

Unit tests have been updated and `LogExpFunctions.jl` is now a dep of both `EpiAware` and the `EpiAware` test env.

Closes #258 .


EpiAware/src/EpiLatentModels/broadcast/helpers.jl
```diff
@@ -8,7 +8,8 @@ Constructs a `BroadcastLatentModel` appropriate for modelling the day of the wee
 - `BroadcastLatentModel`: The broadcast latent model.
 "
 function broadcast_dayofweek(model::AbstractTuringLatentModel)
-    return BroadcastLatentModel(model, 7, RepeatEach())
+    transformed_model = TransformLatentModel(model, x -> 7 * softmax(x))
```
seabbs, line 11: do we really want to hard code this vs provide it as the default link? I think there is an arg either way and given that might as well make it an arg?

EpiAware/src/EpiObsModels/ascertainment/helpers.jl
```diff
@@ -23,6 +23,6 @@ rand(gen_obs)
 "
 function ascertainment_dayofweek(model::AbstractTuringObservationModel;
         latent_model::AbstractTuringLatentModel = HierarchicalNormal(),
-        link = x -> exp.(x))
+        link = x -> x)
```
seabbs, line 26: Need to update the docs here to reflect changes to broadcast and the change in default.

seabbs (changes requested):
LGTM. Just need

- docs updated
- Clarity on if the softmax should be a default arg or if we want to hard code (as now)

EpiAware/src/EpiLatentModels/broadcast/helpers.jl
SamuelBrand1: I guess its sort of a debate about what helper functions are for: one definite thing or a reduction of options to a few sensible ones?

seabbs (approved):
Nice. LGTM

PR #262, opened 2024-06-06
update epiaware and make necessary changes

### LINKED ISSUE #261: Update `EpiAware` to depend on more recent version of `Turing.jl` and `>=` in `compat`
At the moment in main branch the precise version of deps being used in `EpiAware` are:

```bash
Status `~/GitHub/CFA/Rt-without-renewal/EpiAware/Project.toml`
⌅ [47edcb42] ADTypes v0.2.7
  [80f14c24] AbstractMCMC v5.2.0
  [0bf59076] AdvancedHMC v0.6.1
⌅ [1313f7d8] DataFramesMeta v0.14.1
  [31c24e10] Distributions v0.25.109
  [ffbed154] DocStringExtensions v0.9.3
⌅ [366bfd00] DynamicPPL v0.24.11
  [2ab3a3ac] LogExpFunctions v0.3.28
  [c7f686f2] MCMCChains v6.0.6
  [b1d3bc72] Pathfinder v0.8.7
  [1fd47b50] QuadGK v2.9.4
  [189a3867] Reexport v1.2.2
  [bd369af6] Tables v1.11.1
⌅ [fce5fe82] Turing v0.30.9
  [37e2e46d] LinearAlgebra
  [9a3f8284] Random
  [2f01184e] SparseArrays v1.10.0
  [10745b16] Statistics v1.10.0
```

`Turing` is moving forward and changing the interface to samplers (see #238 and links from that issue).

What I think is happening here is that there is some kind of `[compat]` mutual blocking going on because latest versions of some packages like `ADTypes.jl`, which has moved to `v1.2.1` but we have pinned to somewhere `[0.2.0 - 0.3.0)`.

A related question is should we have weaker compat, e.g. here https://pkgdocs.julialang.org/v1/compatibility/#Inequality-specifiers

However, I'm slightly worried about this causing longer term problems.

SamuelBrand1:
This PR updates the deps of `EpiAware` to latest version by updating `[compat]` first. I've made the decision here to pin on minor version again _except_ for packages that come with base Julia (including Julia itself) which I'm stating have to have `>= 1.9`.

### `MvNormal` constructor

Following from this it turns out that 
```julia
MvNormal(I(n))
```
is no longer a valid constructor for our purposes. The reason is that `I(n)` constructs a `Diagonal{Integer}` and now `Turing` is sensitive to the `eltype` of the constructor of `MvNormal`. This leads to `InexactError` chucks later on.

I've replaced with
```julia
MvNormal(n, one(eltype(...)))
```
although potentially other options are available.

Closes #261 .


seabbs (approved):
All makes sense. LGTM

PR #265, opened 2024-06-08
change default LatentDelay constructor to right truncation at 99th perc

SamuelBrand1:
I spotted that the default behaviour of the usual `LatentDelay` constructor was to have right truncation at 15 time steps irrespective of the underlying delay distribution. This is obviously a potential source of making silent errors.

This PR changes the constructor so that default of behaviour of this constructor is to calculate the 99th percentile of the observation distribution i.e.

```math
x_{99} = F^{(-1)}(0.99).
```
Where $F^{(-1)}$ is the inverse of the distribution function of the delay[^1]. And use the rounded value of $x_{99}$ as the right truncation value for the constructor.

[^1]: NB: the constructor is only defined for continuous delay distributions so this is always valid.

### DIFF
--- a/EpiAware/src/EpiObsModels/LatentDelay.jl
+++ b/EpiAware/src/EpiObsModels/LatentDelay.jl
@@ -9,7 +9,13 @@ observed data.
 - `pmf::T`: The probability mass function (PMF) representing the delay distribution.
 
 ## Constructors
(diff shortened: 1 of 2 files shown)


seabbs (approved):
Nice. LGTM

PR #267, opened 2024-06-09
Issue 266: Fix kernel bug

### LINKED ISSUE #226: Pipeline constructor to include save path specification
For the future we might want to have another module that defines paths to save results to and pass in as an arg (vs using the default) or alternative specify this via do_inference?

_Originally posted by @seabbs in https://github.com/CDCgov/Rt-without-renewal/pull/224#discussion_r1604949013_

seabbs:
Closes #226 but adding a `partial = false` option to the observation kernel generating funciton, adds tests for this, and uses this in `LatentDelay` method.

_Still not loving how this shifts complexity into the error model here. Holding this though as thinking about making a abstract type for observation error models where this could be offloaded and if we have that then this would seem to be less of a problem._


SamuelBrand1 (approved):
This seems to work... I'm a bit uncomfortable that of you generate data e.g. `y_t = missing` then you have to pre-pend something like `fill(missing, length(obs.pmf))` to then infer parameters.

But I think this kind of clunkiness is being improved by other PRs/issues.

seabbs:
> But I think this kind of clunkiness is being improved by other PRs/issues.

Yes, in #268 hopefully (and if not does seem like its own issue beyond the bug fix)

PR #268, opened 2024-06-09
Issue 160: Add ObservationErrorModel type and functions

seabbs:
This PR proposes an approach for resolving #160 by introducing a new `AbstractTuringObservationErrorModel` and associated generic methods.

Before making this a full PR I need to: 

- [x] Add documentation for new methods
- [x] Add tests for abstract methods
- [x] Decide if `pos_shift` can live in the generic as a hard-coded value.
- [x] Decide if we want a `EpiAwareBase` set of abstract types.

_Note that this approach also suggests a modifier wrapper (i.e `AbstractObservationModifierModel`) and also suggests this branch/leaf setup elsewhere (i.e latent models) but this can be discussed here and acted on via other issues. It may just be infra we don't currently need IMO._

**This now extends #267 and updates handling of latent delays so that missingness is instead concatenated into the beginning of the vector and the internal obs error model detects this and only starts fitting. To get this working I needed to stop using `undef` when `y_t = missing` which has required changing some of the tests. This fix is still not ideal but it at least preserves the input out of the model.**


EpiAware/src/EpiObsModels/ObservationErrorModels/AbstractTuringObservationModel.jl
```diff
@@ -0,0 +1,29 @@
...
+        y_t,
+        Y_t)
+    @submodel priors = generate_observation_error_priors(obs_model, y_t, Y_t)
+
+    if ismissing(y_t)
+        y_t = Vector{Union{Real, Missing}}(missing, length(Y_t))
+    end
+
+    first_Y_t = findfirst(!ismissing, Y_t)
+    for i in first_Y_t:length(Y_t)
```
seabbs, line 14: We could now also fairly easily add support for using a vectorised observation model in cases where we know there are no missing elements (or all are missing). I don't think we know if this has performance implications though.

EpiAware/src/EpiObsModels/LatentDelay.jl
```diff
@@ -64,13 +64,16 @@ Generates observations based on the `LatentDelay` observation model.
 
 "
 @model function EpiAwareBase.generate_observations(obs_model::LatentDelay, y_t, Y_t)
-    @assert length(obs_model.pmf)<=length(Y_t) "The delay PMF must be shorter than or equal to the observation vector"
+    first_Y_t = findfirst(!ismissing, Y_t)
```
SamuelBrand1, line 67: I'm abit confused that this is looking for the possibility of `missing` values in `Y_t`? What is the use case for this? Different length latent models?

EpiAware/src/EpiObsModels/EpiObsModels.jl
```diff
@@ -11,6 +11,10 @@ using ..EpiLatentModels: HierarchicalNormal, broadcast_dayofweek
 
 using Turing, Distributions, DocStringExtensions, SparseArrays
 
+# Abstract observation model
+export AbstractTuringObservationErrorModel, generate_observation_error_priors,
```
SamuelBrand1, line 15: I thought we were keeping a code pattern of types and methods being exported in separate lines for readability?

EpiAware/src/EpiObsModels/LatentDelay.jl
SamuelBrand1: OK I can see you mention this in the new docstrings.

seabbs: This is so you can stack `LatentDelay` (i.e it preserves composability). Maybe another solution here but not sure what it is

EpiAware/src/EpiObsModels/EpiObsModels.jl
seabbs: yes though this was just to group this discrete part. The main question I have here is should we move the abstract type to base as well as the non-Turing method?

SamuelBrand1:
> @SamuelBrand1 do you have thoughts about the outstanding questions ?

Re: Base types. The upside of having the abstract types in `EpiAwareBase` is that we are providing the type tree for people to create concrete subtypes for. Given that we have fairly specific abstract types there e.g. `AbstractTuringIntercept` then I think `AbstractTuringObservationErrorModel` can live there too?

SamuelBrand1:
> Decide if pos_shift can live in the generic as a hard-coded value.

I'm agnostic on this.

SamuelBrand1:
This is looking good so far; conditional on 

> **This fix is still not ideal but it at least preserves the input out of the model.**

seabbs:
> I'm agnostic on this. 

We need to make a decision. Is it important too be able to customise/turn off. If its not it can go in the abstract type. If we can implement it so that you can still do so it can also go in the abstract type.

SamuelBrand1:
> > This is looking good so far; conditional on
> 
> Conditional on which part?

That this is probably (? Idk I haven't thought of a better approach yet) not ideal but it fixes my problem with #267 .

SamuelBrand1:
> > I'm agnostic on this.
> 
> We need to make a decision. Is it important too be able to customise/turn off. If its not it can go in the abstract type. If we can implement it so that you can still do so it can also go in the abstract type.

This is not important to customise.

seabbs:
@SamuelBrand1 this is ready for review. I have added the positive shift to the abstract `generate_observations` (hard coded), removed it from the error models, and moved the abstract type into EpiAwareBase.

seabbs:
Note the reason you are being asked for a review is that I tighted branch controls to dismiss reviews when new commits are made.

SamuelBrand1 (approved):
LGTM. I like these changes; opens up other obvious observation error models (e.g. `LogNormal`, `Normal`, `TDist` etc with less repetitive code/time shift reasoning.

PR #270, opened 2024-06-10
Add `ConcatLatentModels`

seabbs:
This PR adds a new latent model helper `ConcatLatentModels`. This is similar to `CombineLatentmodels` but as the name suggests it concatenates rather than combines. This PR also reorganises the latent models module to be split into folders based on purpose.


SamuelBrand1 (dismissed):
I like this, both the new functionality and the refactor.

SamuelBrand1 (approved):
LGTM

Issue #269, opened 2024-06-10
Test EpiAware CI running very slowly


seabbs:
As can be seen here https://github.com/CDCgov/Rt-without-renewal/actions/workflows/test-EpiAware.yaml?query=branch:main

as of a recent update our test CI has started to run very slowly for no apparent reason. Do we have clarity on if this is on our side, or due to a dependency change. If it is due to a dependency change can we update our code (i.e to remove reliance on a depreciated feature) to speed the CI (and so potentially our package performance) back up

SamuelBrand1:
Did this start after #262 ?

SamuelBrand1:
Looking at current CI... The two Mac OS CI groups finished in 10 and 15 mins. And documenter (I guess built under Ubuntu) took 5 mins.

seabbs:
It looks like we have a bunch of depreciations from #262 that could be causing problems? https://github.com/CDCgov/Rt-without-renewal/actions/runs/9462089214/job/26064276243#step:7:448

SamuelBrand1:
> It looks like we have a bunch of depreciations from #262 that could be causing problems? https://github.com/CDCgov/Rt-without-renewal/actions/runs/9462089214/job/26064276243#step:7:448

I would think that this just creates some warnings; I don't think it explains the performance diff between different OS targets.

SamuelBrand1:
This was true back here too? https://github.com/CDCgov/Rt-without-renewal/actions/runs/8736028303/usage

SamuelBrand1:
```bash
test (1.10, ubuntu-latest) | 2h 11m 34s |  
test (1.10, windows-latest) | 45m 51s |  
test (1.10, macOS-latest) | 15m 45s

```
Latest complete CI... thats pretty wild.

seabbs:
This looks to have been resolved by #272. This makes the point very strongly to me that we need either to have continuous benchmarks as part of the CI or watch runtimes more closely.

PR #272, opened 2024-06-11
Update `MvNormal` constructor with `FillArrays.Fill`

### LINKED ISSUE #271: Fix `MvNormal` depreciation
> `MvNormal(d::Int, σ::Real)` is deprecated, use `MvNormal(LinearAlgebra.Diagonal(FillArrays.Fill(σ ^ 2, d)))` instead.

I am seeing this depreciation warning. If there is a performance problem due to depreciation as raised in #269 then this is a likely culprit.

SamuelBrand1:
Does as per title to close #271


seabbs (approved):
This looks like it fixes the performance issues seein in #269 which is great. The syntax really is getting horrible with these changes. Is using `MvNormal` still the recommendation in these simple independent cases?

Issue #256, opened 2024-06-04
Redo underlying "truth data" sampling


SamuelBrand1:
This looks good. 

As a note: this PR implies redoing the underlying data generating for the pipeline (which is good and essentially what we are fixing here).

_Originally posted by @SamuelBrand1 in https://github.com/CDCgov/Rt-without-renewal/pull/250#pullrequestreview-2096114906_

As per title in light of this PR.

seabbs:
Is this done or still in progress?

SamuelBrand1:
I'm updating this... it makes more sense to also have "truth" day of week effect

PR #278, opened 2024-06-12
Make the 99th perc default constructor a `censored_pmf` feature

SamuelBrand1:
This is a small PR that follows on from #265. Whilst working on #256 I realised that #265 change should apply to both the observation and the discretised generation interval construction (this is useful when considering scenarios with different mean generation intervals).

This PR does this, plus adds a unit test. The unit tests to `LatentDelay` are unchanged because this default is now inherited from `censored_pmf`.

Slight modification is that the inferred `D` (right truncation value) is now rounded to nearest multiple of $\Delta d$ rather than integer.


seabbs (dismissed):
Agree. LGTM

SamuelBrand1:
Just caught the `EpiData` constructor and set to this new default.

seabbs (approved):

PR #282, opened 2024-06-13
scoring tool for pipeline

### LINKED ISSUE #247: Evaluation method on inference output. I'm assuming we outsource this to `scoringutils`.

SamuelBrand1:
This PR adds scoring utility to the pipeline and closes #247.

The main contribution here is the `score_parameters` function, which collects inference/forecast samples into a Dataframe and sends to an `R` run time to use `scoringutils` and returns a dataframe of summary scores.

Example usage is in `pipeline/test/end-to-end/test_scoring.jl`, my main worry after doing a full run is that the function is abit low level given the boilerplate needed to look at _all_ the times points.


pipeline/src/infer/InferenceConfig.jl
```diff
@@ -19,7 +19,7 @@ struct InferenceConfig{T, F, I, L, E}
     "Latent model type."
     latent_model::L
     "Case data"
-    case_data::Union{Vector{Integer}, Missing}
+    case_data::Union{Vector{Union{Integer, Missing}}, Missing}
```
seabbs, line 22: are these actually related to getting the pipeline working with the latest `EpiAware` version?

pipeline/src/scoring/score_parameters.jl
```diff
@@ -0,0 +1,45 @@
...
+
+"""
+Internal function for scoring a DataFrame containing a prediction and truth value
+    for a parameter using the `scoringutils` package.
+"""
+function _score(df)
+    @rput df
+    R"""
+    library(scoringutils)
+    result = df |> as_forecast() |> score() |> summarize_scores(by = "parameter")
```
seabbs, line 19: it seems weird to be summarising by score and hard coding this -> given we know we want to be breaking this down by other strata (in the analysis plan). Why not just get rid of the summarise_scores call and summarise in julia as we need?

seabbs (commented):
This overall looks good - though there are a collection of random changes. 

My main question is do we want to summarise in R and if so we need a follow up to generalise this in order to allow us to do all the evaluations we need to do.

pipeline/test/end-to-end/test_scoring.jl
```diff
@@ -0,0 +1,104 @@
...
+        map(eachrow(forecast_y_t)) do row
+            if any(ismissing, row)
+                missing
+            else
+                quantile(row, q)
+            end
+        end
+    end
+    plt = scatter(truthdata, xlabel = "t", ylabel = "y_t", label = "truth data")
+    vline!(plt, [n_observation_steps + missing_padding + 0.5], label = "forecast start")
```
seabbs, line 59: why not reuse truth_gen here it seems like unnecessary room for error

pipeline/src/scoring/score_parameters.jl
```diff
@@ -0,0 +1,45 @@
...
+- `samples`: A `MCMCChains.Chains` object of samples.
+- `truths`: Truth values for each parameter.
+- `model`: (optional) The name of the model. Default is "EpiAware".
+
+## Returns
+- `result`: A DataFrame containing the summarized scores for the parameter.
+
+"""
+function score_parameters(param_names, samples, truths; model = "EpiAware")
+    df = mapreduce(vcat, param_names, truths) do param_name, truth
```
seabbs, line 41: I don't really get why you are doing construction of the required dataframe here? Surely our lives will be easier if we have a dataframe of posterior predictions more generally and then just pass to score with conversion of appropriate column names?

seabbs (commented):
Looking again I wonder at the decision to nest creating a dataframe with scoring.

pipeline/src/infer/InferenceConfig.jl
SamuelBrand1: Yes.

pipeline/test/end-to-end/test_scoring.jl
SamuelBrand1: Good spot

pipeline/src/scoring/score_parameters.jl
SamuelBrand1: Constructing dataframes is pretty fast, and get passed to `R`. This pattern is more like a julia function which happens to have R under the hood once you've got a `Chains` object from either inference or forecasting. Obviously, in a Julia workflow a `Chains` object is more convenient because it can act like a table-esque object and has other features.

SamuelBrand1:
I think the key comments from @seabbs here are:

> My main question is do we want to summarise in R and if so we need a follow up to generalise this in order to allow us to do all the evaluations we need to do.

> I don't really get why you are doing construction of the required dataframe here? Surely our lives will be easier if we have a dataframe of posterior predictions more generally and then just pass to score with conversion of appropriate column names?

This PR gives a function that get target any parameter in a `MCMCChains.Chains` object to score vs a truth data point. It might be better to do all the model results, combine into a single dataframe and do one pass to `scoringutils`?

pipeline/src/scoring/score_parameters.jl
seabbs: Okay fair enough.

seabbs:
I think what my questions are trying to get at is does this function give you the flexibility we need to do the analysis plan and I think the answer to that is no as it only supports summarising by parameter.

pipeline/src/scoring/score_parameters.jl
seabbs: or passing scores out to summarise_scores (its quite handy as it knows about scores to summarise).

SamuelBrand1:
> I think what my questions are trying to get at is does this function give you the flexibility we need to do the analysis plan and I think the answer to that is no as it only supports summarising by parameter.

It does processes by adding the `process_strings = [myprocess[$(i)] for i in 1:n]` pattern which then get scored for each time point. I demo that in the extra end-to-end run. But having to write that might mean this approach is too low level/inconvenient which is what I was worried about in the opening.

SamuelBrand1:
> This is really locked in by the `DataFrame` setup in the linking function and hard coding `parameter` in by.

Yes, I can see that. I guess that an argument for a post-processing step which makes a big dataframe out of _all_ the models we trial and we send selected cols to R/scoringutils?

seabbs:
yes or does julia -> score -> julia followed by julia -> summarise -> julia

SamuelBrand1:
> yes or julia -> score -> julia followed by julia -> summarise -> julia

I'm undecided, but I think the `julia -> score -> julia` function is handy functionality even if we end up going for 

> makes a big dataframe out of all the models we trial and we send selected cols to R/scoringutils

approach

pipeline/src/scoring/score_parameters.jl
```diff
@@ -0,0 +1,45 @@
...
```
seabbs, line 41: Why not just use the `get` function and pull multiple parameters at once?

seabbs (approved):
LGTM. I still have some reservations but I think we should merge and see how this works out when sticking together with other parts of the pipeline.

Issue #289, opened 2024-06-17
Refactor away unnecessary pipeline methods


SamuelBrand1:
Generally, our pattern has been to control pipeline behaviour with methods dispatching on pipeline type. This is fine IMO, but as we've developed its become clear that we can control things without as many methods. For example, `generate_inference_results`: with a small refactor we don't need this to dispatch directly on pipeline type because the underlying function chooses its behaviour on pipeline dispatch.

This is low priority but IMO worth flagging.

SamuelBrand1:
I think #295 has closed this.

PR #295, opened 2024-06-18
Example mode and unit tests for pipeline functions.

### LINKED ISSUE #293: Example/test mode for pipeline functions
Reviewing the pipeline, there is missing test/example mode testing for the actual pipeline functions.

SamuelBrand1:
Does as per title and closes #293

### DIFF
--- a/pipeline/src/constructors/make_inference_configs.jl
+++ b/pipeline/src/constructors/make_inference_configs.jl
@@ -20,31 +20,6 @@ function make_inference_configs(pipeline::AbstractEpiAwarePipeline)
         "gi_std" => gi_param_dict["gi_stds"], "log_I0_prior" => priors["log_I0_prior"]) |>
                         dict_list
 
-    return inference_configs
-end
-
-"""
-Create inference configurations for the given pipeline. This is the example method,
-which only returns a randomly selected inference configuration.
-
-# Arguments
-- `pipeline`: An instance of `EpiAwareExamplePipeline`.
-
-# Returns
-- An object representing the inference configurations.
-
-"""
-function make_inference_configs(pipeline::EpiAwareExamplePipeline)
-    gi_param_dict = make_gi_params(pipeline)
-    namemodel_vect = make_epiaware_name_latentmodel_pairs(pipeline)
-    igps = make_inf_generating_processes(pipeline)
-    obs = make_observation_model(pipeline)
-    priors = make_model_priors(pipeline)
-
-    inference_configs = Dict("igp" => igps, "latent_namemodels" => namemodel_vect,
-        "observation_model" => obs, "gi_mean" => gi_param_dict["gi_means"],
-        "gi_std" => gi_param_dict["gi_stds"], "log_I0_prior" => priors["log_I0_prior"]) |>
-                        dict_list
-
-    return [rand(inference_configs)]
+    selected_inference_configs = _selector(inference_configs, pipeline)
+    return selected_inference_configs
 end
--- a/pipeline/src/constructors/make_inference_configs.jl
+++ b/pipeline/src/constructors/make_inference_configs.jl
@@ -22,3 +22,29 @@ function make_inference_configs(pipeline::AbstractEpiAwarePipeline)
 
     return inference_configs
 end
+
+"""
+Create inference configurations for the given pipeline. This is the example method,
+which only returns a randomly selected inference configuration.
+
+# Arguments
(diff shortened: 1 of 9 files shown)


pipeline/src/pipeline/do_inference.jl
```diff
@@ -17,3 +17,24 @@ function do_inference(truthdata, pipeline::AbstractEpiAwarePipeline)
...
+
+# Arguments
+- `truthdata`: The truth data used for generating inference results.
+- `pipeline::EpiAwareExamplePipeline`: Sets the pipeline behavior.
+
+# Returns
+An array of inference results.
+
+"""
+function do_inference(truthdata, pipeline::EpiAwareExamplePipeline)
```
seabbs, line 33: Am I missing something? Why can't this rely on the abstract/default method?

pipeline/src/pipeline/do_truthdata.jl
```diff
@@ -16,3 +16,22 @@ function do_truthdata(pipeline::AbstractEpiAwarePipeline)
...
+Generate truth data for the EpiAwarePipeline. This is example mode.
+
+# Arguments
+- `pipeline::EpiAwareExamplePipeline`: The EpiAwarePipeline object.
+
+# Returns
+An array of truth data generated from the given pipeline.
+
+"""
+function do_truthdata(pipeline::EpiAwareExamplePipeline)
```
seabbs, line 30: Same point here about relying on the default

seabbs: Doesn't this really suggest a new method for make truth data configs?

pipeline/src/pipeline/do_inference.jl
SamuelBrand1: The thing this method does differently is pick an inference config randomly out of the list generated by `make_inference_configs`. An alternate approach would be to do the dispatch at that level though (connected to #289 )

pipeline/src/pipeline/do_truthdata.jl
SamuelBrand1: As per #289 ? That makes sense

SamuelBrand1:
I've moved the dispatch on being the example mode pipeline to the constructor for the list of configutations to loop over. Good shout @seabbs it is cleaner this way.

pipeline/src/pipeline/do_inference.jl
seabbs: This seems like it would reduce copy and paste across methods?

pipeline/src/constructors/make_inference_configs.jl
```diff
@@ -22,3 +22,29 @@ function make_inference_configs(pipeline::AbstractEpiAwarePipeline)
...
+
+# Arguments
+- `pipeline`: An instance of `EpiAwareExamplePipeline`.
+
+# Returns
+- An object representing the inference configurations.
+
+"""
+function make_inference_configs(pipeline::EpiAwareExamplePipeline)
+    gi_param_dict = make_gi_params(pipeline)
```
seabbs, line 38: do we not wnant to just call make_inference_configs(abstract method) here internally and then use rrand on it?

seabbs:
See my comment. I thought the solution was just calling the abstract method internally to the new method and then calling rand on that?

SamuelBrand1:
> See my comment. I thought the solution was just calling the abstract method internally to the new method and then calling rand on that?

That was the original pass which I didn't like

seabbs:
What about having a `DefaultEpiAwarePipeline` that is just the abstract type fall backs and calling that? It just seems like a weird pattern (in the sense of forcing a lot of duplication)?

seabbs (commented):
Nice. Seems like a great solution.

seabbs (approved):

SamuelBrand1:
> What about having a `DefaultEpiAwarePipeline` that is just the abstract type fall backs and calling that? It just seems like a weird pattern (in the sense of forcing a lot of duplication)?

I think a `DefaultEpiAwarePipeline` is tricky because we want to dispatch within the highest level method (highest level from p.o.v. of the type hierarchy).

What I've just pushed is the creation of an internal `_selector` method which does the selection from the list of possible scenarios (with default being to return all of them).

Going forwards I think the way we have done the pipe leans into trait-based programming (cf [`SimpleTraits.jl`](https://github.com/mauro3/SimpleTraits.jl)), but it was a bit of a leap to implement that in this PR.

PR #296, opened 2024-06-18
Issue 252: Composing complex models

### LINKED ISSUE #252: Support complex composed models by preventing name collisions
At the moment no matter what we do you can end up with situations where models can have name conclusions (i.e when they use a lot of latent models. 

What should we do about this? #249 introduces a partial solution for observation models (that uses a prefix based on the named assigned by the user - partial as the uniqueness of this name is not enforced).

We also see this problem more generally (for example if ascertainment models are stacked and if multiple latent models are combined)

We have a few options as I see it.

- Enforce prefix for all submodel calls so that a parameter name becomes dependent on its call stack
- Introduce the pattern that all calls to a latent model should have a clear (and documented prefix) - potentially as in the above solution based on the type of the dispatching struct. 

The latter seems preferable to me as variables names will be shorter and easier to reason on. There may also be other options.

seabbs:
This PR closes #252 by adding prefix constructors to all models that call a submodel in addition to passing down to an underlying submodel. It also adds helpers (`PrefixLatentModel` and `PrefixObservationModel`) to allow users to easily add prefixes themselves to problematic submodules. 

I made the choice to make make changes at the struct level vs within the turing models as this allows users to reason on their model based on struct nesting. The downside is it doesn't easily allow for dynamic prefix changes and creates a potential disconnect between the prefix in the top level struct and the one in the wrapped model (i.e if you used `Accessors.jl` to copy the struct with changes to one prefix and not the other).
 
I think there are some options here for additional helpers and maybe dropping support for "" in prefix_submodel (maybe has a small performance hit) but that these should be addressed in their own issues.

> [!NOTE]
>Following up on previous work I have also moved a few things around to make the organisation a bit more standard.


seabbs:
Any chance to review this this week @SamuelBrand1?

SamuelBrand1:
> Any chance to review this this week @SamuelBrand1?

I'll try for tomorrow, its contingent on how some other work comes together. This needs a bit of deep dive and thought.

seabbs:
@SamuelBrand1 bumping this

EpiAware/src/EpiAwareUtils/prefix_submodel.jl
```diff
@@ -0,0 +1,30 @@
...
+
+# Examples
+
+```julia
+using EpiAware
+submodel = prefix_submodel(CombineLatentModels([FixedIntercept(0.1), AR()]), generate_latent, \"Test\", 10)
+rand(submodel)
+```
+"
+@model function prefix_submodel(
```
SamuelBrand1, line 22:
```suggestion
@model function prefix_submodel(
        model::AbstractModel, fn::Function, prefix::String = ""; kwargs...)

        @submodel prefix=eval(prefix) submodel=fn(model, kwargs...)
        
    return submodel
end
```

Would seem to be more idomatic, does this cause problems with `eval`?

EpiAware/src/EpiObsModels/StackObservationModels.jl
```diff
@@ -77,7 +80,7 @@ Generate observations from a stack of observation models. Assumes a 1 to 1 mappi
 
     obs = ()
     for (model, model_name) in zip(obs_model.models, obs_model.model_names)
-        @submodel prefix=eval(model_name) obs_tmp=generate_observations(
+        @submodel obs_tmp = generate_observations(
             model, y_t[Symbol(model_name)], Y_t[Symbol(model_name)])
         obs = obs..., obs_tmp...
```
SamuelBrand1, line 85: This seems an unusual way to `reduce` over a collection? You're recursively generating an immutable `Tuple`. `mapreduce` is smart enough to produce a tuple give tuple input, I'm not sure why you've gone a different way here?

EpiAware/src/EpiObsModels/modifiers/PrefixObservationModel.jl
```diff
@@ -0,0 +1,29 @@
...
+
+    # Examples
+    ```julia
+    using EpiAware
+    observation_model = PrefixObservationModel(Poisson(), \"Test\")
+    obs = generate_observations(observation_model, 10)
+    rand(obs)
+    ```
+"
+@kwdef struct PrefixObservationModel{M <: AbstractTuringObservationModel, P <: String} <:
```
SamuelBrand1, line 16: Do we really need different classes and methods for latent and observation models? It seems pretty close to copy and paste? Is the hangup that `prefix_submodel` is taking more postional args?

SamuelBrand1 (commented):
This looks really good directionally, this is an important extra feature. I especially like the concat/combine model handling plus a really expressive doc string example of potential usage (e.g. cases _and_ deaths).

Slight queries on the underlying tuple generation and is there really not a way to have a `PrefixModel` struct where you end up dispatching on the type of the `model` field so we don't have more than one `PrefixModel` type? It seems a bit inelegant and makes the code a bit more complex.

SamuelBrand1:
> Do we really need different classes and methods for latent and observation models? It seems pretty close to copy and paste? Is the hangup that prefix_submodel is taking more postional args?

After f2f conversation, this is the reason for the design decision here... which makes sense.

SamuelBrand1 (approved):
After conversation about the design choice, I think this approach makes the most sense. I'm pro this.

EpiAware/src/EpiObsModels/StackObservationModels.jl
seabbs: I don't think I know what these words mean? Can we port to an issue?

EpiAware/src/EpiAwareUtils/prefix_submodel.jl
seabbs: This is because if you supply anything at all to `prefix` as part of submodule it changes in name. So `""` produces `.original-var` for example.

I think it is debatable if we want to support empty prefixing here as you should never not have a prefix (and I assume this takes some compute. Going to move to an issue.

seabbs: See #309

EpiAware/src/EpiObsModels/modifiers/PrefixObservationModel.jl
seabbs: Resolved face to face

PR #298, opened 2024-06-19
Issue 274: Add continuous benchmark structure

### LINKED ISSUE #274: Continuously benchmark performance
`epinowcast` uses `{touchstone}` for continuous benchmarking of PRs (see https://github.com/epinowcast/epinowcast/pull/464#issuecomment-2115829621). This is very useful when making proposed model "improvements" as it gives a very clear sanity check of if it is an improvement.

This would have been very useful to us for catching #269 faster and automating the identification of the cause. This suggests to me it is only going to be more useful going forward. 

In the Julia world we have `BenchmarksTools.jl` which is nice and it seems like there is an integration with this https://github.com/benchmark-action/github-action-benchmark action which seems like it would do nearly everything we might want apart from giving a clear indicator in PR (like `touchstone` does). 

Another option we could take would be to commit some output from `BenchmarkTools.jl` (which is what @kaitejohnson did in their wastewater work for scores) for runtimes this might not be ideal as you need to have timings from the server where CI is running. 

A final option would be to replicate `touchstone` in julia. i.e running both the target and proposed code on the same CI machine and comparing run times and then reporting.

seabbs:
This PR closes #274. 

It adds infrastructure for benchmarking using `BenchmarkTools`, `BenchmarkCI`, and `PkgBenchmark`. I have also added proposed structure for future benchmarks but this is not set in stone. When I originally looked at this I kind of wanted to integrate this with `TestItemRunner` (and so use out test suite for benchmarks) but I couldn't work out how to do this and also potentially it is better to have more targetted benchmarks?

The action CI for this should: 

- Run on all PRs
- Cancel currently running benchmarks
- Run benchmarks on the PR code and on main
- Post a comment comparing these two sets of benchmarks.
- It only runs on Ubuntu on the latest Julia version. `BenchmarkCI` doesn't appear to support a build matrix hence this decision (plus saving compute). This means we won't catch platform-specific performance regressions.

This could potentially get a bit noisy/slow so may need to watch how this is working out and go from there.

This can be tested locally in the `benchmark` environment using:

```julia
using PkgBenchmark
benchmarkpkg(Rtwithoutrenewal)
``` 

You should also be able to use `BenchmarkCI` locally to compare commits but I haven't tested this. 

> [!NOTE]
> This PR does not add an actual benchmark suite. I don't really have a great handle on `BenchmarkTools` yet and I think this should probably be done in a few issues once we have an idea of the sorts things we want to benchmark (to discuss here). Initially I would propose doing an issue per submodule and benchmarking each generative model + a few combinations and then also have a few integrated models that we benchmark in more detail (in another module/issue?).


EpiAware/benchmark/Project.toml
SamuelBrand1, line 1: Is this from doing `develop` on `EpiAware`?

SamuelBrand1:
This looks a really good move.

EpiAware/benchmark/Project.toml
seabbs: yes

.github/workflows/benchmark.yaml
```diff
@@ -0,0 +1,32 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: julia-actions/setup-julia@v1
+      - uses: julia-actions/cache@v1
+      - name: Build EpiAware
+        run: julia --project=EpiAware -e 'using Pkg; Pkg.build()'
+      - name: Build EpiAwarePipeline
+        run: julia --project=EpiAwarePipeline -e 'using Pkg; Pkg.build()'
+      - uses: julia-actions/julia-buildpkg@v1
```
seabbs, line 26: @SamuelBrand1 locally this all works now after moving to a repo wide benchmarking suite (as it can't handle subdirectories and no way to hack) but I am really not clear why the develop packages in `Rtwithoutrenewal` can't be found in CI?

seabbs: I've tried a bunch of things here with no succes.

seabbs:
This now works i think but needs testing from main. As it doesn't touch any other code going to merge based on previous partial review

PR #308, opened 2024-06-27
add forecasting to pipeline solve and save

### LINKED ISSUE #292: Add forecasting to pipeline "solve and save".

SamuelBrand1:
This PR adds a `generate_forecasts` function step inside the `infer` function; which gets either loaded or saved as part of the pipeline.

Note that I've created a new method for `generate_forecasts` to be used internally in `infer` along with the previous method which was aimed at being used _after_ inference by dispatching on a dictionary of inference results. The reason I've kept the post-inference method is because it makes it easy to vary the forecast lookahead window post-inference as desired.

Closes #292


seabbs (approved):
Nice good idea. Strong agree on keeping the post inference method around!

PR #314, opened 2024-06-28
Add calculate processes function to pipeline

### LINKED ISSUE #116: Function to backcalculate Rt and the growth rate

SamuelBrand1:
This PR adds a function that calculates the three process targets for inference:

- log incidence
- exponential growth rate
- (Instantaneous) reproduction number.

Adds a simple `rt` calculation, and wraps `EpiAware.expected_Rt`, but with left truncation effect dealt with backwards exponential growth according to first value of `rt`.

Closes #116


pipeline/src/utils/calculate_processes.jl
```diff
@@ -0,0 +1,67 @@
...
+- `I0`: Initial infections at time zero.
+- `init_rt`: Initial exponential growth rate.
+- `data::EpiData`: An instance of the `EpiData` type containing generation interval data.
+"""
+function _calc_Rt(I_t, I0, init_rt, data::EpiData)
+    @assert all(I_t .> 0) "Log infections must be positive definite."
+    @assert I0>0 "Initial infections must be positive definite."
+    n = length(data.gen_int)
+    rev_pmf = reverse(data.gen_int)
+    aug_I_t = [I0 * exp(-init_rt * (n - i)) for i in 1:n] |>
```
seabbs, line 37: I don't love encoding this seeding assumption everywhere (i.e not in `EpiAware` for sure and potentially not in a version of this pipeline that is meant to be more generic). 

I wonder if it would be better to just only return Rt from the start of when there is enough infection data? 

Design wise I also think we should try and have a single I_t vector as the input vs expecting people to treat I_t and I_0 differently? 

The problem with my suggestion is that you then get a different length Rt to everything else but perhaps the solution there is to pad with missingness?

This is all just a thought and happy to move discussion into a new issue

SamuelBrand1: > The problem with my suggestion is that you then get a different length Rt to everything else but perhaps the solution there is to pad with missingness?

Yes, this was my reason for going this way (but not changing the underlying `EpiAware` function with a strong decision).

SamuelBrand1:
I've kept the functionality the same here, but added an internal method for seeding the infections. The backwards exponential is now a default dispatching on the pipeline type. This should make it easier to address new functionality later as flagged by @seabbs .

seabbs (approved):
Good idea

PR #320, opened 2024-07-01
Reorganise documentation

### LINKED ISSUE #275: Make the docs correctly represent the submodules that make up EpiAware and surface the Public and private APIs
As the title.

seabbs:
This PR closes #275 and # by reorganising the documentation. Note that many of the files are placeholders and need content (and also issues to drive content creation).

> [!WARNING]
> I haven't tested this locally so it may ping off the CI a few times as I update it.


seabbs:
@SamuelBrand1 what do you think of this so far? 

Any thoughts as to why the `build` function is having problems? As far as I can tell nothing has changed here that should be causing the pathing error it is throwing and yet...

SamuelBrand1:
Looks like you need to move the underlying data? Missing dir/file is causing the documenter CI fail.

EpiAware/docs/src/getting-started/tutorials/intro.md
```diff
@@ -0,0 +1,66 @@
+The diagram below shows the relationship between the modules in the package for a typical workflow.
```
SamuelBrand1, line 1: I think this mermaid diagram needs updated? Maybe make an issue for that.

SamuelBrand1 (dismissed):
I really like the new structure for the docs, and it seems a lot more user readable.

EpiAware/docs/src/getting-started/tutorials/intro.md
seabbs: yup it does. Once this is merged going to make a wave of issues for all the doc gaps

seabbs:
I think this should now build. Note I did some reorganisation vs before. The issue was in the @contents call. It looks like using the side bar panels has some confusing bugs that are hard to pin down. The main issue was that even commenting them out didn't do away with the. bug

SamuelBrand1 (approved):
LGTM. I like the rename to explainers

SamuelBrand1 (approved):

Issue #232, opened 2024-05-21
Set up pipeline as a separate package


SamuelBrand1:
At the moment `AnalysisPipeline` pakage/env (soon to be renamed `EpiAwarePipeline as per #230 ) has a combination of generic functions and abstract types and some specific behaviours (e.g. dispatching on prior predictive modelling with `RtwithoutRenewalPriorPipeline`).

It would be neater to abstract the pipeline infrastructure into a full package which can be used by the specific pipeline `src` code.

## Tasks

- [ ] Move generic `EpiAwarePipeline` tools into a package hosts at `~/EpiAwarePipeline`.
- [ ] Do #206 
- [ ] Do #209

PR #329, opened 2024-07-03
Prefix method

### LINKED ISSUE #328: A prefix method to simplify dispatch on pipeline type for saved data.

SamuelBrand1:
Simple PR that closes #328


SamuelBrand1:
Added a prefix for `simulate` as well.

pipeline/src/infer/inference_prefix.jl
```diff
@@ -0,0 +1,54 @@
+"""
+This is an internal method that generates the prefix for the inference results file name.
+"""
+function _inference_prefix(truthdata, inference_config, pipeline::AbstractEpiAwarePipeline)
```
seabbs, line 4: There is a huge amount of copy and pasting here?

pipeline/src/pipeline/pipelinetypes.jl
```diff
@@ -22,3 +20,43 @@ The pipeline type for the Rt pipeline without renewal in test mode.
...
+using EpiAwarePipeline, Plots
+pipeline = SmoothOutbreakPipeline()
+Rt = make_Rt(pipeline) |> Rt -> plot(Rt,
+    xlabel = "Time",
+    ylabel = "Rt",
+    lab = "",
+    title = "Smooth outbreak scenario")
+```
+"""
+struct SmoothOutbreakPipeline <: AbstractRtwithoutRenewalPipeline
```
seabbs, line 40: the fact we keep needing to make things abstract makes mee thing that everything should be abstract and then have a specific instance so extensions can be more easily made?

pipeline/src/pipeline/pipelinetypes.jl
```diff
@@ -22,3 +20,43 @@ The pipeline type for the Rt pipeline without renewal in test mode.
...
+end
+
+"""
+The pipeline type for the Rt pipeline for an outbreak scenario where Rt has
+    discontinuous changes over time due to implementation of measures.
+"""
+struct MeasuresOutbreakPipeline <: AbstractRtwithoutRenewalPipeline
```
seabbs, line 47: ideally both the outbreaks would share a meta type (with the susceptible depletion and then measures would be a special case? Same in the endemic setting?

pipeline/src/simulate/simulate_prefix.jl
```diff
@@ -0,0 +1,12 @@
+"""
+Internal method for setting the prefix for the truth data file name.
+"""
+_simulate_prefix(pipeline::AbstractEpiAwarePipeline) = "truth_data"
```
seabbs, line 4: It seems like you could do this via being a bit clever with the type name in the default method and then not need all this boiler plate?

seabbs (commented):
Looks like it will work. Quite a bit of boiler plate here and its unclear to me right now how we will put these different scenarios in our overall dagger framework to compute efficiently?

pipeline/src/pipeline/pipelinetypes.jl
SamuelBrand1: Ultimately, the analysis is coming down to about 4/5 examples + some actual data.

In terms of readability with regard to an eventual reader of a paper, I think having more verbose types is worth it even if there are more lines of code.

On the other hand, having a powerful traits based system would be very slick for letting people define their own analysis... this is connected to #232 because a traits based system would be a natural feature of a `EpiAwarePipeline` package.

The main question is what do we prioritise first?

SamuelBrand1:
> Looks like it will work. Quite a bit of boiler plate here and its unclear to me right now how we will put these different scenarios in our overall dagger framework to compute efficiently?

Mapping across scenarios activates the `@Dagger.spawn` macro for making `EagerThunks`, its essentially adding one layer of mapping to what we have now.

pipeline/src/pipeline/pipelinetypes.jl
seabbs: I aagree we should gt it working and then think about reusability at a later date

seabbs (dismissed):
Agree with the push back. This is good to go for now. I see its pointing at main but really is built on another PR so will need to be careful when merging in

seabbs (approved):
LGTM

Issue #48, opened 2024-02-19
Doctests with @testitem


seabbs:
See this discussion for a work around: https://github.com/julia-vscode/TestItemRunner.jl/issues/49#issuecomment-1656704296

This would be a nice feature to have as it would allow us to bring examples into the testing suite and link these with our interactive local testing UI

PR #332, opened 2024-07-04
Add Doctest infra and add doctests to EpiAwareUtils

seabbs:
As the title this PR makes a start on resolving #48 by adding doctests to @testitem and adding some to code base for EpiAwareUtils

After looking into it it doesn't look like there is a way to make sure example code is being run in docstrings unless we use doctests and definitely not if we want the output in Documenter (which I think we do to guide our users).


SamuelBrand1 (approved):
This is good stuff. Noting #333 issue, which is very irritating but worth moving forwards here and trying to add more complicated doctests later.

Issue #318, opened 2024-06-29
Robustness against stochastic failure in test CI


SamuelBrand1:
I note that #314 got bumped from the merge queue due to an occasional stochastic failure... despite this PR not being concerned with `EpiAware`.

what is the best way to build resilience here without causing false pass checks? Maybe fixing the rand seed?

seabbs:
Yes I think fixing the seed here and generally makes a lot of sense

seabbs:
Good point. Is there really no ability to run across test setup code?

SamuelBrand1:
Looks like this is coming https://github.com/julia-vscode/TestItemRunner.jl/pull/57

AFAIK its blocked by integration into VS-code test suite.

SamuelBrand1:
Things moving here

https://github.com/julia-vscode/TestItemRunner.jl/pull/57#issuecomment-2212396718

SamuelBrand1:
Alternate suggestion from f2f discussion is an `EpiAware` submodule with test utility functions.

SamuelBrand1:
This is covered by #419 .

PR #334, opened 2024-07-05
Fix random seed for _apply_method test

SamuelBrand1:
As per title.

This mini PR contributes to #318 but doesn't solve that issue. Full solution approach depends on `TestItemRunner` capability as discussed in issue #318.


seabbs (approved):
LGTM

PR #335, opened 2024-07-05
Rolling window for inference

### LINKED ISSUE #330: Overload for `make_tspan` to dispatch on a rolling inference window.

SamuelBrand1:
This PR makes the inference problem into a sequence of inference problems with a rolling window `lookback` (default 35 time steps), `start` (Default 21 ts), and `stride` (default 7 ts) indicating the start and strides of the "current" time along with the size of the rolling window.

It also modifies methods that expected a fixed `tspan` to create a tspan from information in the inference config.

Closes #330


seabbs (approved):
LGTM

PR #337, opened 2024-07-07
Add TuringBenchmarks

### LINKED ISSUE #300: Implement skeleton benchmarks
This should look like a few key benchmarks per submodule and a few benchmarks doing integration testing across modules.

@SamuelBrand1 maybe your combination work can be used here vs implementing as tests?

This issue is only for skeleton/partial benchmarks and any PR resolving it should make new issues to implement a wider set of benchmarks.

seabbs:
This PR closes #300 by using `TuringBenchmarks` to add benchmarks for all Turing models defined in `EpiInfModels`, `EpiObsModels`, and `EpiLatentModels`. This pattern seems sensible to adopt everywhere (i.e all other submodules and in utility code) as it allows for testing multiple auto-diff backends and the sampler function with little effort. 

In doing this work I have identified a issue with `scan` that is unresolved (and so `Renewal` benchmarks are not currently being run and that in several places gradients from different backends differ indicating issues. More generally there are a range of Zygote issues we could address if we want better support this backend.


seabbs:
I've added new issues (#339, #340,  #338, #341) to resolve issues identified here. I plan to tackle #341 on Tuesday.

SamuelBrand1:
>  More generally there are a range of Zygote issues we could address if we want better support this backend.

Quick opinionated view. Since `Zygote` is a source-to-source AD it greatly benefits from custom `rrule`s (https://github.com/JuliaDiff/ChainRulesCore.jl), my prior without a deep analysis is that to get good performance with Zygote for `EpiAware` we'd need to be committed to creating a system of custom rules.

seabbs:
We could just stop benchmarking against it for now then

EpiAware/src/EpiLatentModels/manipulators/ConcatLatentModels.jl
```diff
@@ -113,13 +113,13 @@ Generate latent variables by concatenating multiple latent models.
     @assert sum(dims)==n "Sum of dimensions must be equal to the dimension of the latent variables"
 
     @submodel final_latent, latent_aux = _concat_latents(
-        latent_models.models, 1, [], [], dims, latent_models.no_models)
+        latent_models.models, 1, Real[], [], dims, latent_models.no_models)
```
SamuelBrand1, line 116: Is forcing the array to have an abstract eltype 100% necessary here? Given the performance implication from [here](https://turinglang.org/docs/tutorials/docs-13-using-turing-performance-tips/#ensure-that-types-in-your-model-can-be-inferred)

EpiAware/src/EpiLatentModels/manipulators/ConcatLatentModels.jl
```diff
@@ -113,13 +113,13 @@ Generate latent variables by concatenating multiple latent models.
...
 
     return final_latent, (; latent_aux...)
 end
 
 @model function _concat_latents(
-        models, index::Int, acc_latent, acc_aux,
+        models, index::Int, acc_latent::AbstractVector{<:Real}, acc_aux,
```
SamuelBrand1, line 122: Similar to above, does this capture concrete eltypes (answer might be yes?)

EpiAware/src/EpiLatentModels/manipulators/ConcatLatentModels.jl
seabbs: I don't know but I think this should be its own issue. This was just done to get benchmarks to run without error.

benchmark/Project.toml
```diff
@@ -2,6 +2,9 @@
 BenchmarkCI = "20533458-34a3-403d-a444-e18f38190b5b"
 BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
+DynamicPPL = "366bfd00-2699-11ea-058f-f148b4cae6d8"
 EpiAware = "b2eeebe4-5992-4301-9193-7ebc9f62c855"
 EpiAwarePipeline = "66314692-3678-4cbb-b3ac-6fcb54505317"
 PkgBenchmark = "32113eaa-f34f-5b0d-bd6c-c81e245fc73d"
+Turing = "fce5fe82-541a-59a6-adf8-730c64b5f9a0"
+TuringBenchmarking = "0db1332d-5c25-4deb-809f-459bc696f94f"
```
SamuelBrand1, line 10: I had forgotten about `TuringBenchmarking`! Great spot.

EpiAware/src/EpiLatentModels/manipulators/ConcatLatentModels.jl
seabbs: Closing this as a repeat. I suggest we deal with in its own issue

benchmark/bench/EpiInfModels/Renewal.jl
```diff
@@ -0,0 +1,34 @@
+# Error:  ArgumentError: Converting an instance of ReverseDiff.TrackedReal{Float64, Float64, Nothing} to Float64 is not defined. Please use `ReverseDiff.value` instead.
```
SamuelBrand1, line 1: Are these auto-generated error messages added to the file by benchmarking or personal notes?

SamuelBrand1: Ah, these are noted in the opening message.

EpiAware/src/EpiLatentModels/manipulators/ConcatLatentModels.jl
seabbs: See #345. We can use it as a test case as if it improves performance the benchmarks should flag that.

benchmark/bench/EpiInfModels/Renewal.jl
seabbs: There is an issue here as well #338 where we could also store these?

SamuelBrand1 (approved):
This is really good. I've got some queries about use of abstract eltypes that should become issues, but a before and after picture on that would be handy.

PR #349, opened 2024-07-08
new do_pipeline method for multiple scenarios

### LINKED ISSUE #348: New layer function for `pipeline` to go over multiple scenarios.

SamuelBrand1:
This PR does the following things:

- Adds a `do_pipeline` method for ingesting multiple scenario pipelines and dispatching them sequentially.
- Reduces the number of `ndraws` in `EpiAwareExamplePipeline` dispatch on `make_inference_method` to just 20.
- Includes the pipeline function testing (which is now much faster).

The obvious caveat here is that `EpiAwareExamplePipeline` type is now aimed at checking code ability to run rather than inference correctness. This affects the purpose of the end-to-end tests and should be a new issue.

Closes #348

### DIFF
--- a/pipeline/src/pipeline/do_pipeline.jl
+++ b/pipeline/src/pipeline/do_pipeline.jl
@@ -21,6 +21,8 @@ Perform the pipeline for each `AbstractEpiAwarePipeline` in the given vector `pi
 
 """
 function do_pipeline(pipelines::Vector{<:AbstractEpiAwarePipeline})
-    map(do_pipeline, pipelines)
+    map(pipelines) do pipeline
+        Dagger.@spawn do_pipeline(pipeline)
+    end
     return nothing
 end
--- a/pipeline/src/pipeline/do_pipeline.jl
+++ b/pipeline/src/pipeline/do_pipeline.jl
@@ -2,7 +2,7 @@
 Create a pipeline by generating truth data and making inferences.
 
 # Arguments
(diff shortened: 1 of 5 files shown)


pipeline/src/pipeline/do_pipeline.jl
```diff
@@ -12,3 +12,15 @@ function do_pipeline(pipeline::AbstractEpiAwarePipeline)
...
+
+"""
+Perform the pipeline for each `AbstractEpiAwarePipeline` in the given vector `pipelines`.
+
+# Arguments
+- `pipelines`: A vector of `AbstractEpiAwarePipeline` objects.
+
+"""
+function do_pipeline(pipelines::Vector{<:AbstractEpiAwarePipeline})
+    map(do_pipeline, pipelines)
```
seabbs, line 24: shouldn't this really be a dagger loop to take advantage of parallisation?

seabbs (approved):

pipeline/src/pipeline/do_pipeline.jl
SamuelBrand1: > shouldn't this really be a dagger loop to take advantage of parallisation?

Actually I meant to make a comment on this. This still exposes parallelism, because of the `@Dagger.spawn` in the loop. But if we had more than ~120 cores I'm not sure anymore.

I guess its zero cost so I can change this. It might be a bit smarter at transition from one scenario to another.

seabbs (approved):

PR #355, opened 2024-07-09
fix concatlatentmodel fail to construct vector of non-prefix models

### LINKED ISSUE #353: ConcatLatentModels can't handle two HierarchicalNormal
As the title. This appears to be an issue in the `ConcatLatentModels` constructor when `PrefixLatentModel` is called.

```julia
ConcatLatentModels(
                [HierarchicalNormal(-0.1, truncated(Normal(0, 0.1), 0, Inf)),
                HierarchicalNormal(1, truncated(Normal(0, 0.1), 0, Inf))]
)
ERROR: MethodError: Cannot `convert` an object of type 
  PrefixLatentModel{HierarchicalNormal{Float64, Truncated{Normal{Float64}, Continuous, Float64, Float64, Float64}}, String} to an object of type 
  HierarchicalNormal{R, Truncated{Normal{Float64}, Continuous, Float64, Float64, Float64}} where R<:Real

Closest candidates are:
  convert(::Type{T}, ::T) where T
   @ Base Base.jl:84
```

SamuelBrand1:
This fixes the bug identified #353 . The origin of the bug was a vector of models with defined eltype trying to write back into itself with a different eltype.

I've made the bug a unit test.

Closes #353

### DIFF
--- a/EpiAware/src/EpiLatentModels/manipulators/ConcatLatentModels.jl
+++ b/EpiAware/src/EpiLatentModels/manipulators/ConcatLatentModels.jl
@@ -44,15 +44,14 @@ struct ConcatLatentModels{
         @assert typeof(check_dim)<:AbstractVector{Int} "Output of dimension_adaptor must be a vector of integers"
         @assert length(check_dim)==no_models "The vector of dimensions must have the same length as the number of models"
         @assert length(prefixes)==no_models "The number of models and prefixes must be equal"
-        for i in eachindex(models)
-            if (prefixes[i] != "")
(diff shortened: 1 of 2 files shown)


seabbs (commented):
Nice. Sorry we have worked slightly at cross purposes here! #356 also does this but generalises to CombineLatentModels which had a similar problem. Do you see any tradeoffs/advantages between the two approaches?

SamuelBrand1:
Only thing here is using `map` vs array comprehension but I really don't think that makes much difference!

PR #356, opened 2024-07-09
Issue 353: Use map to assign prefixes in modifier models

### LINKED ISSUE #353: ConcatLatentModels can't handle two HierarchicalNormal

seabbs:
This PR closes #353 by using `map` vs a for loop to create a new vector of `prefix_models`. I have updated `CombineLatentModels`, `ConcatLatentModels`, and `Ascertainment` and added new tests for `CombineLatentModels` and `ConcatLatentModels` that demonstrate the fix working as expected.

### DIFF
--- a/EpiAware/src/EpiLatentModels/manipulators/CombineLatentModels.jl
+++ b/EpiAware/src/EpiLatentModels/manipulators/CombineLatentModels.jl
@@ -34,6 +34,8 @@ latent_model()
         prefix_models = map(eachindex(models)) do i
             if prefixes[i] != ""
                 PrefixLatentModel(models[i], prefixes[i])
+            else
+                models[i]
             end
         end
         return new{AbstractVector{<:AbstractTuringLatentModel}, AbstractVector{<:String}}(
--- a/EpiAware/src/EpiLatentModels/manipulators/CombineLatentModels.jl
+++ b/EpiAware/src/EpiLatentModels/manipulators/CombineLatentModels.jl
@@ -31,13 +31,9 @@ latent_model()
             P <: AbstractVector{<:String}}
         @assert length(models)>1 "At least two models are required"
         @assert length(models)==length(prefixes) "The number of models and prefixes must be equal"
-        prefix_models = map(eachindex(models)) do i
-            if prefixes[i] != ""
-                PrefixLatentModel(models[i], prefixes[i])
-            else
-                models[i]
-            end
-        end
+        prefix_models = [prefixes[i] == "" ? models[i] :
+                         PrefixLatentModel(models[i], prefixes[i])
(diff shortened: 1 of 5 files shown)


EpiAware/src/EpiLatentModels/manipulators/CombineLatentModels.jl
```diff
@@ -31,13 +31,11 @@ latent_model()
             P <: AbstractVector{<:String}}
         @assert length(models)>1 "At least two models are required"
         @assert length(models)==length(prefixes) "The number of models and prefixes must be equal"
-        for i in eachindex(models)
-            if (prefixes[i] != "")
-                models[i] = PrefixLatentModel(models[i], prefixes[i])
-            end
-        end
+        prefix_models = [prefixes[i] == "" ? models[i] :
```
SamuelBrand1, line 34: Pedantic point, but this is array comprehension rather than `map`. There isn't a difference in output but array comprehension guarantees a sequence of operations in array construction (this is used in e.g. `Flux` recurrent NNs), whereas `map` doesn't. 

As of Julia > `0.5`, `map` is a sometimes a (small) bit faster than array comprehension (e.g [this discussion](https://discourse.julialang.org/t/map-vs-loops-array-comprehensions-in-julia-1-0/16137)). However, in this case its probably a judgement call on code readability which one to use.

SamuelBrand1 (approved):
LGTM.

EpiAware/src/EpiLatentModels/manipulators/CombineLatentModels.jl
seabbs: Yeah fair enough. I started with map and ported over as I felt like this was easier to read. I don't feel strongly.

seabbs: Going to resolve but could maybe have an issue to use a standard iterator pattern everywhere

PR #358, opened 2024-07-10
Issue 246: Standarise outputs

### LINKED ISSUE #246: Remove passing large amounts of `NamedTuple`s in `return` value of `Turing` models
At the moment, we are passing a lot of `NamedTuple`s with "auxiliary" variables as generated quantities / return object from constructed `Turing` models.

This is mainly used internally via `@submodel` macro usage on constructing `Turing` models that represent various components of the epidemiological model we want to model generatively.

I don't think passing all these variables is necessary, cause code bloat and _could_ be making type inference more difficult (I'm not so sure about this though). I think we can eliminate to only pass samples of the latent models.

### LINKED ISSUE #345: Update _concat_latent to avoid enforcing a Real
As the title. This may have performance implications. Thanks to #337 we should also be able to track if updating this improves performance.

seabbs:
This PR closes #246 

It standardises all submodels currently in use to have a single return unless otherwise required (currently only the observation error prior model).

It looks like this gives a nice little performance boost and definitely cleans up the interface. 

Note that in testing I found having a `ExpectObs` observation model that adds deterministic recording - perhaps we want this? or alternatively we need a better pattern for reconstructing those parts of the model.

Something else that I realised when testing is we have zero parameter recover tests of any kind. Having these for a picked ad backend would probably catch a lot of issues when developing models. Not sure where these sit in the world where you can have gradient tests (and those would overlap heavily with our benchmarks).

I in general also still find chain objects really clunky and really just want them in a long format data frame with variables by index and name (i.e expected obs in one column and index in another) - I think I've just been spoiled by `tidybayes` here.

This PR also cheekily closes #345 and updates observation stacking to avoid mutating a tuple.


SamuelBrand1:
> It standardises all submodels currently in use to have a single return unless otherwise required (currently only the observation error prior model).
> 
> It looks like this gives a nice little performance boost and definitely cleans up the interface.

Love it.

SamuelBrand1:
> Something else that I realised when testing is we have zero parameter recover tests of any kind. Having these for a picked ad backend would probably catch a lot of issues when developing models. Not sure where these sit in the world where you can have gradient tests (and those would overlap heavily with our benchmarks).

Its not quite zero e.g 

https://github.com/CDCgov/Rt-without-renewal/blob/929deb47e7523f5aac64a0ac5c1add828192a3cf/EpiAware/test/EpiLatentModels/manipulators/CombineLatentModels.jl#L91-L106

Where we test against ability to recover samples from a known posterior distribution using data generated randomly from the test model.

SamuelBrand1:
> I in general also still find chain objects really clunky and really just want them in a long format data frame with variables by index and name (i.e expected obs in one column and index in another) - I think I've just been spoiled by `tidybayes` here.
> 
> This PR also cheekily closes #345 and updates observation stacking to avoid mutating a tuple.

Wide format conversion from `Chains` to dataFrame (e.g. `EpiAware.EpiAwareUtils.spread_draws`) was very easy to implement; want to make an issue?

seabbs:
> Its not quite zero e.g

What I mean here is we don't use anything that uses the gradients which is the most important thing for actually using the methods in the real world. All the prior methods are just sampling directly which is good to check but it leaves a gap where bugs can happen

SamuelBrand1:
> > Its not quite zero e.g
> 
> What I mean here is we don't use anything that uses the gradients which is the most important thing for actually using the methods in the real world. All the prior methods are just sampling directly which is good to check but it leaves a gap where bugs can happen

But it is [sampling with `NUTS`](https://github.com/CDCgov/Rt-without-renewal/blob/929deb47e7523f5aac64a0ac5c1add828192a3cf/EpiAware/test/EpiLatentModels/manipulators/CombineLatentModels.jl#L89) so a grad error is a sufficient condition to fail this test because a silent grad error will almost certainly cause an incorrect posterior. But I know what you mean.

seabbs:
> But it is [sampling with NUTS](https://github.com/CDCgov/Rt-without-renewal/blob/929deb47e7523f5aac64a0ac5c1add828192a3cf/EpiAware/test/EpiLatentModels/manipulators/CombineLatentModels.jl#L89) so a grad error is a sufficient condition to fail this test because a silent grad error will almost certainly cause an incorrect posterior. But I know what you mean.

Oh right yes sorry but that is the only example of that I think

EpiAware/src/EpiLatentModels/manipulators/ConcatLatentModels.jl
```diff
@@ -110,22 +110,23 @@ Generate latent variables by concatenating multiple latent models.
     @assert all(x -> x > 0, dims) "Non-positive dimensions are not allowed"
     @assert sum(dims)==n "Sum of dimensions must be equal to the dimension of the latent variables"
 
-    @submodel final_latent, latent_aux = _concat_latents(
-        latent_models.models, 1, Real[], [], dims, latent_models.no_models)
+    @submodel final_latent = _concat_latents(
+        latent_models.models, 1, nothing, dims, latent_models.no_models)
```
SamuelBrand1, line 114: This has an arg as a `Nothing` type but then it varies to another type... this might be a case where `map` works better than recursive function definition. Possible issue.

EpiAware/src/EpiLatentModels/manipulators/ConcatLatentModels.jl
```diff
@@ -110,22 +110,23 @@ Generate latent variables by concatenating multiple latent models.
...
+        return acc_latent
     else
-        @submodel latent, new_aux = generate_latent(models[index], dims[index])
-        @submodel updated_latent, updated_aux = _concat_latents(
-            models, index + 1, vcat(acc_latent, latent),
-            (; acc_aux..., new_aux...), dims, n_models)
-        return updated_latent, (; updated_aux...)
+        @submodel latent = generate_latent(models[index], dims[index])
+
+        acc_latent = isnothing(acc_latent) ? latent : vcat(acc_latent, latent)
```
SamuelBrand1, line 126: And around here.

SamuelBrand1 (approved):
Good changes. I'm starting to think that `mapreduce` or `map` is going to be more effective than the recursive function approach in the `concat` methods. Thoughts? Possible issue?

seabbs:
> Good changes. I'm starting to think that mapreduce or map is going to be more effective than the recursive function approach in the concat methods. Thoughts? Possible issue?

Yes I think you could be right.

EpiAware/src/EpiLatentModels/manipulators/ConcatLatentModels.jl
seabbs: I think this is really another issue to explore? If we see a performance regression that might be an indication of a problem.

seabbs:
I think the issue is that it can't detect the branch name when its been updated via a rebase - a bit annoying but not the end of the world

seabbs:
I'm testing benchmarks on #367. I've played around a bunch and I can't really see what in this PR would be causing this new benchmark failure (the change since it worked was the hotfixes to main which included all of the EpiObsModel benchmarks which had been missed out before).Any thoughts. If not thoughts and if #367 doesn't shed any light perhaps we merge as is and make a issue to explore?

SamuelBrand1:
> The error message seems to be about loading previous benchmark results. It could be because `suite` is misconfigured the earlier guess about the rebase (but I think not)

Yes there is a `MethodError`

```bash
ERROR: MethodError: no method matching loadparams!(::BenchmarkGroup, ::BenchmarkTools.Parameters, ::Symbol, ::Symbol)
```
Has the pattern changed, because the method signature

```bash
loadparams!(::BenchmarkGroup, !Matched::BenchmarkGroup, ::Any...)
```
Is quite close?

seabbs:
Yes I am seeing that and no not that I can see. 

#367 works and so this is either an issue with a change in this PR or its an issue with the configuration of the PR (which would be weird). Going to cherry pick some bits into their own PRs

seabbs:
https://github.com/JuliaCI/PkgBenchmark.jl/blob/6aab4956fbdf7e79a57cda260a2175bf5e13252a/src/runbenchmark.jl#L277

This is the line that is throwing the error I believe and I believe this indicates that the structure of the benchmarking job is changing between main and this PR. I struggle to understand why that might be given they share the same dependencies and benchmarks haven't been changed but am testing this but forcing a retune to see if that resolves the issue. If it does I guess we keep it off, make and issue, and investigate at our leisure. It could be some weird interaction from `make_turing_suite` and changes in the underlying Turing models resulting in modifications to that between PRs?

seabbs:
This didn't address the problem but reading the code it seems like `loadparam` shouldn't be being hit at all. This makes me wonder if its using code from main here. Going to hotfix there and rebase.

SamuelBrand1 (approved):
LGTM. 

As context, during f2f discussion we isolated a problem with BenchmarkCI; it was making its own call to `pkgbenchmark` which was not receiving a `retune` command to create a new branch -> main tuning file which seems to be required for running the Turing benchmarks.

seabbs:
There is still an issue here but it's now further in than the original. It still looks like the structure of suite is different between the two branches. I'll step through again and look locally using debugger

seabbs:
The problem is now in `PkgBenchmark` 

```julia
 PkgBenchmark: benchmark results written to .benchmarkci/result-baseline.json
┌ Info: Finish running benchmarks.
│ * Target: 47 minutes 58 seconds
└ * Baseline: 41 minutes 34 seconds
ERROR: MethodError: no method matching judge(::BenchmarkTools.TrialEstimate, ::BenchmarkTools.BenchmarkGroup)
Closest candidates are:
  judge(::BenchmarkTools.TrialEstimate, !Matched::BenchmarkTools.TrialEstimate; kwargs...)
   @ BenchmarkTools ~/.julia/packages/BenchmarkTools/QNsku/src/trials.jl:221
  judge(!Matched::BenchmarkTools.BenchmarkGroup...; kwargs...)
```

Same problem in the sense that of the outputs isn't in the expected format for judge

seabbs:
I think our conclusion is we should merge now and deal with benchmarks as an issue

Issue #351, opened 2024-07-09
Preliminary results visualisation for all scenarios


SamuelBrand1:
This is a meta-issue for getting everything in place for prelim results following on from discussion #24 .

- [ ] As of #349 we have `Example` mode only drawing a small number of posterior samples. Useful to have a full pass with a small number of posterior samples to have a concrete set of output to test visualisation too (this can be done by passing down to the `make_epimethod` `ndraws` kwarg.
- [ ] Update pipeline script to ingest multiple scenarios
- [ ] Fig1 code
- [ ] Fig2 code
- [ ] Fig3 code
- [ ] Full run.

PR #359, opened 2024-07-10
Create run_testmode_pipeline.jl

SamuelBrand1:
Small PR to add one script which runs the pipeline for all scenarios but with a (small) number of samples/draws per model. The point of this script is to generate small data examples to work forward the visualisation issues in #351.


SamuelBrand1:
Actually, running this has made me spot that different inference times are copying on top so I'm making this also a small bug fix.

SamuelBrand1:
Now added fixes to close #360 and #361

seabbs (approved):
`_prefix_from_pipeline` love to see it!

PR #371, opened 2024-07-15
Parameter recovery tests

SamuelBrand1:
This draft PR is working on parameter recovery tests for `EpiAware` model combinations.

So far I've added two examples of an underlying random walk recovering parameters from self-generated data: 
1. With Normal errors
2. With Neg bin errors (after exp transform from random walk to mean).

The unit test is on posterior predictive p values for scalar parameters of these models (e.g. the step size std, the initial value of the random walk, the cluster factor of the neg bin errors etc) being in the central 99% of the posterior probability mass. This might be a bit too unspecific? Alternative/additions include a PIT test on the posterior predictive time series (against a theoretical uniform distribution).


SamuelBrand1:
I've added the skeleton some bare bones (e.g. no helper function) tests and a "use epiproblem" version of a test.

SamuelBrand1:
I've added a more general approach to testing the model def + inference methods: a @testset group that goes over the keys of the true parameters (generated from the model with no data) and then checks that the true value is within the 99.9% central posterior marginal probability for that parameter.

The idea here is that the false fail rate should be low even with multiple parameters. Obviously, noting that the prob of avoiding false fail rate, conditional on having a correct model, goes approx. like $1 - (0.001)^m$ when $m$ parameters are getting checked.

A few pain points/questions: 

- The key names for the posterior chn and the parameters vary depending on whether we are drawing multivariate vs univariate random values. This has been an ongoing hassle with Turing/DynamicPPL. At the moment, only the scalar parameters (e.g. std of RWs, cluster factor of neg bin observation noise etc) are getting tested this way.
- Does this testing approach run the risk of having a significant false pass rate? My experience is that mis-specified models usually give wildly bad parameter estimates that are captured at this credibility level but thats not a rigourous analysis statement. @dylanhmorris @damonbayer Have you guys met a false pass problem in the `pyrenew` test suite?

SamuelBrand1:
@seabbs Not sure why Benchmark CI has failed?

seabbs:
>   27 dependencies successfully precompiled in 59 seconds. 291 already precompiled.
ERROR: LoadError: MethodError: no method matching logpdf(::EpiAware.EpiAwareUtils.SafePoisson{Float64}, ::Float64)

seabbs:
>  My experience is that mis-specified models usually give wildly bad parameter estimates that are captured at this credibility level

Agree

EpiAware/src/EpiObsModels/ObservationErrorModels/PoissonError.jl
```diff
@@ -21,5 +21,5 @@ The observation error model for Poisson errors. This function generates the
 observation error model based on the Poisson error model.
 "
 function observation_error(obs_model::PoissonError, Y_t)
-    return Poisson(Y_t)
+    return SafePoisson(Y_t)
```
seabbs, line 24: I assume this is what is causing benchmarks to fail

EpiAware/test/EpiLatentModels/models/RandomWalk.jl
```diff
@@ -38,3 +38,39 @@ end
...
+    using Random, Turing, FillArrays, Distributions, LinearAlgebra, DynamicPPL, StatsBase,
+          ReverseDiff
+    Random.seed!(1234)
+
+    rw_process = RandomWalk()
+    obs_nb = NegativeBinomialError()
+
+    @model function test_negbin_errors(rw, obs, y_t)
+        n = length(y_t)
+        @submodel Z_t = generate_latent(rw, n)
```
seabbs, line 52: you could save duplication by hard coding the rw and obs struct in the gen calls vs passing through

EpiAware/test/EpiLatentModels/models/RandomWalk.jl
```diff
@@ -38,3 +38,39 @@ end
...
+
+    generative_mdl = test_negbin_errors(rw_process, obs_nb, fill(missing, 40))
+    θ_true = rand(generative_mdl)
+    Z_t_obs, y_t_obs = condition(generative_mdl, θ_true)()
+
+    mdl = test_negbin_errors(rw_process, obs_nb, Int.(y_t_obs))
+    chn = sample(
+        mdl, NUTS(adtype = AutoReverseDiff(; compile = Val(true))), 1000, progess = false)
+
+    #Check that are in central 99.9% of the posterior predictive distribution
```
seabbs, line 65: we could maybe make this a bit higher i.e 99%

EpiAware/test/EpiLatentModels/models/RandomWalk.jl
```diff
@@ -38,3 +38,39 @@ end
...
+    #Therefore, this should be unlikely to fail if the model is correctly implemented
+    @testset "Check true parameters are within 99.9% central post. prob.: " begin
```
seabbs, line 67: this should be a test helper function as its reused multiple times

EpiAware/test/EpiObsModels/modifiers/LatentDelay.jl
```diff
@@ -148,16 +148,156 @@ end
...
+
+    @model function test_negbin_errors_with_delays(rw, obs, y_t)
+        n = length(y_t)
+        @submodel Z_t = generate_latent(rw, n)
+        @submodel gen_y_t = generate_observations(obs, y_t, exp.(Z_t))
+        return exp.(Z_t), gen_y_t
+    end
+
+    #Generate data from model
+    generative_mdl = test_negbin_errors_with_delays(
```
seabbs, line 184: can we take different models as inpuits and make this a helper fn as well

EpiAware/test/EpiObsModels/modifiers/LatentDelay.jl
```diff
@@ -148,16 +148,156 @@ end
...
+        ∇ℓ_rd = LogDensityProblemsAD.ADgradient(Val(:ReverseDiff), ℓ)
+        ∇ℓ_rd2 = ADgradient(:ReverseDiff, ℓ; compile = Val(true))
+        ∇ℓ_fd = LogDensityProblemsAD.ADgradient(Val(:ForwardDiff), ℓ)
+
+        x = randn(n)
+        val1, g1 = LogDensityProblems.logdensity_and_gradient(∇ℓ_rd, x)
+        val2, g2 = LogDensityProblems.logdensity_and_gradient(∇ℓ_fd, x)
+        val3, g3 = LogDensityProblems.logdensity_and_gradient(∇ℓ_rd2, x)
+
+        @test val1 ≈ val2 ≈ val3
```
seabbs, line 209: do we need this given we have benchmarking also doing this? Thoughts?

seabbs (changes requested):
Really nice. Some thoughts about duplicated code.

EpiAware/test/EpiLatentModels/models/RandomWalk.jl
SamuelBrand1: Good idea! From the point of view of using `TestItemRunner` probably easiest to have an `EpiAwareTestUtils` submodule?

SamuelBrand1: Maybe make this + refactor a new issue?

SamuelBrand1:
> > 27 dependencies successfully precompiled in 59 seconds. 291 already precompiled.
> > ERROR: LoadError: MethodError: no method matching logpdf(::EpiAware.EpiAwareUtils.SafePoisson{Float64}, ::Float64)

This makes sense right? `Poisson` is not defined floats only ints... why was it working before or not been erroring already?

EpiAware/test/EpiObsModels/modifiers/LatentDelay.jl
SamuelBrand1: Ah, this was my idiot checking that got a bit carried away. This should just replicate the Turing benchmarks so can get removed. 

Maybe is tests the `Turing.Model` -> `LogDensityProblem` conversion but... I think this is just exactly what Turing benchmarks does anyway.

EpiAware/test/EpiLatentModels/models/RandomWalk.jl
SamuelBrand1: This is where I'm a bit torn. If we have a test on 99% cred. int. then assuming the posterior dist has accurate coverage then if we test 10 params we expect it to pass $0.99^10 \approx 0.9$ proportion of the time. But a one in 10 stochastic CI failure rate could be a bit annoying?

Otoh, I set random seed so if we found one that worked (for Mac, Window and Ubuntu) then it will work until we do something that breaks it.

I don't have a lot of experience with optimizing an inference package so not sure where to pitch this.

EpiAware/test/EpiLatentModels/models/RandomWalk.jl
SamuelBrand1: Or loop over a few models to make this worthwhile?

SamuelBrand1:
> > This makes sense right? Poisson is not defined floats only ints... why was it working before or not been erroring already?
> 
> I assume `Poisson` converts to a Int internally maybe?

Yep, thats the one

EpiAware/test/EpiLatentModels/models/RandomWalk.jl
seabbs: but we are fixing seeds right and also I assume the intervals aren't independent?

SamuelBrand1: > but we are fixing seeds right and also I assume the intervals aren't independent?

Good point... failure is probably pretty strongly correlated here.

SamuelBrand1: Changed default to 0.99.

EpiAware/src/EpiObsModels/ObservationErrorModels/PoissonError.jl
SamuelBrand1: Think I've fixed that.

EpiAware/test/EpiLatentModels/models/RandomWalk.jl
SamuelBrand1: I've made a function `test_full_process` that ingests a latent model and a IGP model (the same matrix as the pipeline... not coincidently). The test matrix creates this array of models and runs `test_full_process` on each.

SamuelBrand1:
My main new addition are a test suite of the infection generating process models and latent process model combinations that are used in pipeline.

This test suite is in the `"LatentDelay parameter recovery with mix of IGP + latent processes: Negative binomial errors + EpiProblem interface"` test item. Its formed of the three main IGPs we consider: `DirectInfections`, `ExpGrowthRate` and `Renewal` with an arbitrary (non-disease specific) GI. The three latent models are `RandomWalk`, `AR` and a differenced `AR` process. 

The test model priors determined by the combination of IGP + latent model, this is influenced by #405, and whilst it shouldn't be technically necessary to wrangle the priors since this is a recovery test, my belief is that `ExpGrowthRate` having priors that imply high probability of gigantic infection rates will negatively influence testing the model.

Following discussion with @seabbs on this PR I've set the default test to be each true parameter value for each model is in the central 99% posterior probability distribution.

NB: I've removed an older test item which was effectively identical to using `DirectInfections` IGP.

SamuelBrand1:
OK initial CI run at 99% gave:

> ERROR: LoadError: Some tests did not pass: 440 passed, 1 failed, 0 errored, 0 broken.

I'm going to try at 99.5%.

seabbs (approved):
This is great. I think we might want a new issue to make a testutils to 1. Make some of the repeated code into test utilities and 2. make plots of the output. We can then generate these plots at test time and load them as an environment object in CI (for visual checking). We could even set up posting a comment with the model output if we think that is useful.

PR #370, opened 2024-07-15
Gathering samples into interpretable analysis and fig1 code

SamuelBrand1:
This PR does two things:

1. Provides a set of functions aimed at gathering saved samples from the analysis pipeline into a dataframe with columns for:
  - Reference time (e.g. when analysis was done in time line of the data).
  - Scenario outline
  - Latent process used
  - Mean GI used in generating truth data
  - Mean GI used in analysis (this is for misspecification of GI)
  - Infection generating process used.
2. A set of functions which checks input dataframe for comprehension and generates subplots for figure 1. The groupings are by Scenario, latent process and infection generating process. Reference time is fixed to date of biggest Rt shift. The subplots are combined using `AlgebraOfGraphics.jl` and rendering is done by `CairoMakie`.

Closes #362

PR #375, opened 2024-07-17
New figure 1 function

### LINKED ISSUE #374: Modify Fig1 to split between main and SI figs

SamuelBrand1:
This is a small PR which modifies the figure 1 generation to meet discussion points in #370 (and elsewhere). The new generation script generates a figure for each latent model choice (one to be used in main ms and others for SI).

Close #374

### DIFF
--- a/pipeline/scripts/create_figure1.jl
+++ b/pipeline/scripts/create_figure1.jl
@@ -1,4 +1,4 @@
-## Script to make figure 1
+## Script to make figure 1 and alternate latent models for SI
 using Pkg
 Pkg.activate(joinpath(@__DIR__(), ".."))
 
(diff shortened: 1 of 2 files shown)


seabbs (dismissed):
Nice looks good. I am a bit surprised you've needed to make subplots and compose vs using `AlgebraofGraphics` facet mechanics (which is what I would do if making this in `ggplot2` for example. 

What do you think about moving the legend to below the plot vs on the side?

SamuelBrand1:
> Nice looks good. I am a bit surprised you've needed to make subplots and compose vs using `AlgebraofGraphics` facet mechanics (which is what I would do if making this in `ggplot2` for example.

My problem here was that I/search engines couldn't work out how to vary the ylims figure by figure... I'm being a bit basic here I think. I agree because unlike the first pass fig 1 now the subplots are slightly different sized.

SamuelBrand1:
I forgot the reference date indication (mentioned by @seabbs f2f)! Added now.

This now closes #377

seabbs:
Can you do it by specifying it as a variable in the data? I think that would be the ggplot2 solution?

seabbs (approved):

PR #380, opened 2024-07-17
EpiAware specific predict method to handle prediction edge cases

seabbs:
This PR was spun out of #369 and implements a custom predict method and predict context. See that PR for discussion


seabbs:
I'm still stuck on this but tempted to pause work on it as the fix in `Turing` seems to be gaining traction (https://github.com/TuringLang/Turing.jl/issues/2239)

SamuelBrand1:
> I'm still stuck on this but tempted to pause work on it as the fix in `Turing` seems to be gaining traction ([TuringLang/Turing.jl#2239](https://github.com/TuringLang/Turing.jl/issues/2239)) and so maybe we can just wait on that.

I think thats a good idea.

seabbs:
It seems like progress on this on the Turing side might be quite slow so we might need to implement something here sooner rather than later.

SamuelBrand1:
How is this going?

seabbs:
it was passed to wait for Turing. Do you think we need it soon?

PR #381, opened 2024-07-17
Issue 364: Remove Zygote from benchmarks

### LINKED ISSUE #364: Remove Zygote from benchmarks
At the moment we include Zygote in our benchmarking but as per #339 support across our tools is very poor and performance is bad. To reduce the number of warnings we could remove Zygote for now and aim to fix in #339 (and then turn the benchmarks back on).

seabbs:
This PR closes #364 by removing Zygote from all Turing benchmarks. It does this by making a custom `EpiAware` wrapper around `TuringBenchmarking.make_turing_suite()` that sets some defaults (i.e clean is true and the backends we are testing vs they are). 

The downside of this change is we will need to watch their defaults/use to see if they think its time to test a new backend.

### DIFF
--- a/benchmark/bench/EpiInfModels/DirectInfections.jl
+++ b/benchmark/bench/EpiInfModels/DirectInfections.jl
@@ -13,5 +13,5 @@ let
     expected_incidence = exp.(log_init_scale .+ log_incidence)
 
     mdl = generate_latent_infs(direct_inf_model, log_incidence)
-    suite["DirectInfections"] = make_turing_suite(mdl; check = true)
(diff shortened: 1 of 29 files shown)


seabbs:
Note that the benchmarks on this branch don't run Zygote but as expected those from main do (so this is working as expected).

SamuelBrand1 (approved):
LGTM.

SamuelBrand1:
> The downside of this change is we will need to watch their defaults/use to see if they think its time to test a new backend.

Agreed. Concretely, we should keep our eyes out for future integration of `Enzyme` and/or `Tapir`.

PR #382, opened 2024-07-18
Issue 276: Add information to FAQ

### LINKED ISSUE #276: "Hints and tips" documentation
As Julia is so modular we have no need to reimplement some of our helper functionality as it can be used from other packages. However, this attitude across the Julia ecosystem makes packages very hard to use for everyone but experts. 

I think a way to mitigate this might be to have a hints and tips section (or something like useful functionality from other packages) where we collect some of these ideas.

Things that come to mind are: 

- Remaking objects using `Accessors.jl` (#90)
- Conditioning on Turing models using `|` and `condition`
- Tools to access parts `MCMCChain` objects and to visualise them for diagnostics etc.

_Are there any examples of good practice at this in the ecosystem we can be inspired by?_

seabbs:
This PR closes #276 by adding the currently discussed information to the FAQ. For new FAQ updates we should make specific issues.

### DIFF
--- a/EpiAware/docs/src/getting-started/faq.md
+++ b/EpiAware/docs/src/getting-started/faq.md
@@ -1,6 +1,6 @@
 # Frequently asked questions
 
-This page contains a list of frequently asked questions about the EpiAware package. If you have a question that is not answered here, please open a discussion on the GitHub repository.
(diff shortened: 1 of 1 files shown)


SamuelBrand1 (approved):
I like it.

PR #383, opened 2024-07-18
Issue 366: Add record expected obs and latent models

### LINKED ISSUE #366: ExpectObs and ExpectLatent helpers
As flagged in #358 it is useful to be able to extract expected observations but I think it might mean a performance hit if we tried to do so everywhere. One option is to create helper wrappers (`EpiObs` and `EpiLatent`) that uses `:=` to indicate to `Turing.jl` to track these variables. This would mean redefining a model post fitting if you wanted to extract something and then having to pass things back into `generated_quantities` (I haven't tested this but I assume it would work) which could be a bit clunky.

seabbs:
This PR closes #366  by adding `RecordExpectedObs` and `RecordExpectedLatent`.


SamuelBrand1 (approved):
Handy functionality!

PR #387, opened 2024-07-19
378 fig2 code

### LINKED ISSUE #378: Fig2 code

SamuelBrand1:
This commits the figure 2 plotting code. The variation here from figure 1 is that instead of going for a grid by grid loop for subplots I've created a new col which combines the combinations of target and scenario. This turned out to be easier to work with.

I also updated `AlgebraOfGraphics` which allowed easier customization of the legend so that truth data and models are easier to distinguish.

Closes #378

NB: this plot still need some tuning, but I propose final polishing is done as an issue because we'll want to fix all the plots.


seabbs (dismissed):
Implementation looks good.

seabbs (approved):

Issue #357, opened 2024-07-10
Integration benchmarks of Ascertainment fail in certain combinations


seabbs:
In developing integration benchmarks I see the following issue when a Ascertainment model is passed to a LatentDelay and only where the latent model the ascertainment model uses contains random variables (i.e `FixedIntercept` does not trigger this). 

I think this is something to do with line 76 which is where a link field from `Ascertainment` is used but I think the actual issue is in the passing of the `Ascertainment` struct (?). This is low priority whilst we set up more infra but I think could point to a deeper issue.

```julia
using BenchmarkTools, TuringBenchmarking, EpiAware
suite = BenchmarkGroup()
suite["integration"] = BenchmarkGroup()

let
    using Distributions

    obs = LatentDelay(
        Ascertainment(
            NegativeBinomialError(), Intercept(Normal(0, 1)); link = x -> exp.(x)
        ),
        LogNormal(1.6, 0.2)
    )
    I_t = fill(100, 10)
    gen_obs = generate_observations(obs, I_t, I_t)
    suite["Integration"]["LatentDelay - Ascertainment"] = make_turing_suite(gen_obs; check = true)
end
ERROR: ArgumentError: type does not have a definite number of fields
Stacktrace:
  [1] fieldcount(t::Any)
    @ Base ./reflection.jl:885
  [2] #s21#100
    @ ~/.julia/packages/ReverseDiff/p1MzG/src/derivatives/broadcast.jl:27 [inlined]
  [3] var"#s21#100"(F::Any, ::Any, ::Any)
    @ ReverseDiff ./none:0
  [4] (::Core.GeneratedFunctionStub)(::UInt64, ::LineNumberNode, ::Any, ::Vararg{Any})
    @ Core ./boot.jl:602
  [5] istypeorclosure
    @ ~/.julia/packages/ReverseDiff/p1MzG/src/derivatives/broadcast.jl:21 [inlined]
  [6] mayhavetracked
    @ ~/.julia/packages/ReverseDiff/p1MzG/src/derivatives/broadcast.jl:29 [inlined]
  [7] _any_tuple
    @ ./reduce.jl:1241 [inlined]
  [8] any
    @ ./reduce.jl:1237 [inlined]
  [9] mayhavetracked
    @ ~/.julia/packages/ReverseDiff/p1MzG/src/derivatives/broadcast.jl:33 [inlined]
 [10] get_implementation
[...]

Issue #386, opened 2024-07-18
Create a new submodule EpiDist/EpiDelay


seabbs:
We currently use delays in several places and could be doing so via a submodule. This could be called several things. 

I see it being used to: 

- Model reporting delays
- Model generation times
- Model non-parametric and parametric models

To start (i.e to resolve this issue) we would need to move our current Fixed delay functionality into a submodule. We would then need to make issues to implement our `censoring_pmf` approach as a in model fitted delay and to move our generation time model to be a model input.

seabbs:
@SamuelBrand1 I am going to have a go at this soon. Any thoughts on naming?

SamuelBrand1:
I'm leaning towards `EpiDistributions` or possibly `EpiAwareDistributions` although I think that is maybe too anchored. 

My reasoning is that what we are fundamentally doing in the three examples you quote is operating on distributions to create (in some sense) good discrete distributions for epidemiological modelling. Either because of enforcing discrete time dynamics or reasoning on observation censoring (or both). 

I think more verbose and making an implicit link to `Distributions.jl` works in typical Julia module naming conventions.

seabbs:
So I am expecting this to also hold all the nonparametric hazard components that epinowcast deals with so for that reason I think it might make sense to use a more general name?

SamuelBrand1:
That includes things that are discretization of things like [`Survival.jl`](https://github.com/JuliaStats/Survival.jl) which is from the `JuliaStats` org (as is `Distributions.jl`)... Maybe `EpiStats`? 

More left field, the non-hazard models are really special cases of a hazard model so `EpiHazards.jl`?

seabbs:
I feel like EpiHazard might lose people (and not everything would need a hazard approach). Maybe its two interacting submodels that have the same types (and so can be composed together...)?

SamuelBrand1:
> I feel like EpiHazard might lose people (and not everything would need a hazard approach). Maybe its two interacting submodels that have the same types (and so can be composed together...)?

E.g. something like `EpiDists` _and_ `EpiSurvival`?

seabbs:
yes but they would both need to have the same abstract parent type so that they hook in to i.e latent delay in the same way. I think probably it would be EpiHazard though as EpiSurvival would include dists (i.e distributional survival modelling)

seabbs:
Some proposed models (note these are really pseudo code) to go in this module in some form. 

- Double censoring adjusted daily cohort model

```julia
[...]

Issue #385, opened 2024-07-18
Make infection models composable and more modular


seabbs:
Currently infection models don't match the composability of the rest of the tooling. In particular you cannot compose multiple iterative processes together. It seems like we should do this. One choice we should make is how much we want to support this across infection models and how much we want to focus on the renewal based generative process.

I think we should support all equally but this probably means implementing slightly less efficient iterative direct infections and growth rate models (and for these models several of the wrapper models we might use might not make sense to use).

SamuelBrand1:
Conditionally deterministic infection models are a transformation on the random latent process + other rand vars.

It seems that the infrastructure we have allows us to start eliminating infection generating processes as a separate concept to latent processes and observation models

PR #388, opened 2024-07-19
Issue 340: Simplify LatentDelay

### LINKED ISSUE #340: Resolve gradient mismatches in benchmarks
For several utilities, benchmarking suggests that different backends given different gradients. This should be investigated as it may indicate performance issues.

seabbs:
This PR targets but may not close #340. It may also help with #357.

It simplifies the `LatentDelay` approach to use a dot product-based method rather than the current sparse matrix approach. It keeps the sparse matrix constructor as this may be useful in #386 (and generally changes here are temporary as #386 will likely improve this implementation substantially.

I initially implemented this as a simple map but then decided  to implement using the `accumulate_scan` framework as a test. That resulted in a bit more code overhead and complexity but maybe worth it for consistency. It definitely works in this setting though and confirms my belief that it will work for composable modelling in #385

### DIFF
--- a/EpiAware/src/EpiObsModels/EpiObsModels.jl
+++ b/EpiAware/src/EpiObsModels/EpiObsModels.jl
@@ -5,11 +5,11 @@ module EpiObsModels
 
 using ..EpiAwareBase
 
-using ..EpiAwareUtils: censored_pmf, HalfNormal, prefix_submodel
+using ..EpiAwareUtils
 
 using ..EpiLatentModels: HierarchicalNormal, broadcast_dayofweek, PrefixLatentModel
 
-using Turing, Distributions, DocStringExtensions, SparseArrays
+using Turing, Distributions, DocStringExtensions, SparseArrays, LinearAlgebra
 
 # Observation error models
 export PoissonError, NegativeBinomialError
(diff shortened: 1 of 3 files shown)


seabbs:
Benchmarks are a bit noisy (we need to look at tuning down the statistical threshold for flagging) but appears to show a clear speed up for ascertainment and latent delay methods. 

Noting that I still see gradient issues in the benchmarks (but still not fixed Tova source) so this doesn't close the target issue even if it speeds things up a little.

seabbs:
Thinking about it I assume the big variance in the ascertainment model is related to #383

seabbs:
@SamuelBrand1 thoughts on this?

SamuelBrand1:
> > However, there is no apparent speed up for compiled reverse auto-diff.
> 
> If something is going wrong with the compiled branch of the code for `AutoReverseDiff(;compile = true)` then I wouldn't expect the speed to change, but I still haven't isolated the problem here.

Urgh... I hit edit rather than quote reply.

SamuelBrand1 (approved):
LGTM. I think this was worth doing despite flexibility issues for the future and noting that it hasn't resolved #357 .

Also noting that we did consider memoization (e.g. discussed in #161); but haven't got that to work (yet). We'll be revisiting this I suspect.

seabbs:
Yes agree memoization or a turing side resolution to this would be preferable longer term

PR #389, opened 2024-07-22
LatentDelay passes a view of the obs into comparison with expected_obs -- do not merge

SamuelBrand1:
This draft looks at a very quick fix to the problem noted in #357 without making a confident claim that this is the best way forward.

The simple idea is to only pass along case obs `y_t` that aren't affected by left truncation and compare against `expected_obs`.

The upside is that this fixes the MRE of the error I found in #357 for reversediff and compiled reversediff by avoiding constructing arrays with eltypes that are not a subtype of `Real`. The downsides are it causing fails among unit tests that are expecting that kind of input/output _and_ unassessed potential interactions with `Ascertainment`.

This PR is more for reasoning on this rather than a target for merging.


seabbs:
Nice. Closing in favour of #392 which I think should work similarly but be more user friendly.

PR #393, opened 2024-07-22
Update return in `simulate` with more unit tests

SamuelBrand1:
This is a small PR which updates `EpiAwarePipeline` to work with changes in `EpiAware`. 

Only slightly confusing thing was that `calculate_processes` is slightly further from machine precision compared to a direct calculation (passes at `atol = 1e-6` rather than default), since no change should have impacted that.


seabbs (approved):

seabbs:
That is very odd. No ideas from me otherwise looks good

PR #394, opened 2024-07-22
Issue 319: Make all code in Mishra 2020 et al. visible

### LINKED ISSUE #319: Include all code blocks in `replication` example
From f2f conversation, some people find the hidden code blocks generating the plots in the "getting started" example confusing and would prefer if these were visible.

seabbs:
As the title. This PR closes #319 by removing all let statements from the Mishra et al showcase.


seabbs:
Sadly its not as simple as this

SamuelBrand1:
> As the title. This PR closes #319 by removing all let statements from the Mishra et al showcase.

Why not leave `let` visible? Its valid Julia code.

PR #395, opened 2024-07-23
Expose code blocks in replication and explain REPL usage

### LINKED ISSUE #319: Include all code blocks in `replication` example

SamuelBrand1:
This PR expands the replication of Mishra 2022 in two ways:

- All code blocks are exposed.
- There is explanation of how to run this notebook from the REPL from the root directory of `RtWithoutRenewal`.

I think this addresses the problem of this replication notebook requiring far too much from a first time user (e.g. close familiarity with `Pluto`, a desire to run the notebook locally in a way not described here) whilst not ditching the Pluto approach entirely. I've tried local build and running from the REPL... but I'm maybe not the best person to judge user friendliness.

Very interested in views from @kaitejohnson.

Closes #319


EpiAware/docs/src/showcase/replications/mishra-2020/index.jl
```diff
@@ -50,6 +47,29 @@ This tutorial introduces the basic functionality of `EpiAware`. `EpiAware` is a
...
+
+_NB:_ The difference between `let ... end` blocks and `begin ... end` blocks are that the `let ... end` type of code block only adds the final output/return value of the block to scope, like an anonymous function, whereas `begin ... end` executes each line and adds defined variables to scope.
+
+## Environment for this notebook and running it from the REPL
+We are going to use the `docs/` Julia environment for this notebook which is set by the `Project.toml` file in the `EpiAware/docs` folder, and check-out the latest version of `EpiAware` directly from its source code using `Pkg.develop`.
+
+From the context of this notebook we can find these _relative_ to this notebook using the `@__DIR__` macro, which returns the path of this notebook. If you are running this from the REPL from the root of the `RtWithoutRenewal` repository then replace the following:
+
```
kaitejohnson, line 60: Is the idea that you would want users to be running from the root of `RtwithoutRenewal` generally? 

Or would this also work after installing the package in a different repo? These questions come from my lack of understanding of the environment management in Julia, and over dependence on the R package structure :)

EpiAware/docs/src/showcase/replications/mishra-2020/index.jl
```diff
@@ -50,6 +47,29 @@ This tutorial introduces the basic functionality of `EpiAware`. `EpiAware` is a
...
+docs_dir = \"EpiAware/docs\"
+pkg_dir = \"EpiAware\"
+```
+"
+
+# ╔═╡ 27d73202-a93e-4471-ab50-d59345304a0b
+md"
+## Dependencies for this notebook
+In the code block above we activated the environment we wanted, this determines the exact versions of the extra packages we want to use as dependencies. We also ran `Pkg.instantiate` to make sure we have locally downloaded these packages.
+
```
kaitejohnson, line 71: Ok-- so by setting `pkg_dir = \"EpiAware\"` this is the equivalent in python of installing requirements or in R as using a renv file to install dependencies, but we still need to import/load them downstream

EpiAware/docs/src/showcase/replications/mishra-2020/index.jl
```diff
@@ -76,6 +96,28 @@ C_t \sim \text{NegBin}(\text{mean} = I_t,~ \text{overdispersion} = \phi).
 In the examples below we are going to largely recreate the _Mishra et al_ model, whilst emphasing that each component of the overall epidemiological model is, itself, a stand alone model that can be sampled from.
 "
 
+# ╔═╡ 1d3b9541-80ad-41b5-a5ed-a947f5c0731b
+md"
+## Load the data into scope
+First, we make sure that we have the data we want to analysis in scope by reading from a local CSV file. As with the environment, this notebook works relative to its directory. If you are running this from the REPL from the root of the `RtWithoutRenewal` repository then replace the following:
+
+```julia
+notebook_dir = \"EpiAware/docs/src/showcase/replications/mishra-2020\"
```
kaitejohnson, line 105: Might want to make this system agnostic (e.g. aren't the slashes the other way in Windows?) In R I always use file.path("each", "component")

kaitejohnson (commented):
General comment is I am curious how I would run this vignette outside the root of the repo, as I had tried in my sandbox yesterday after installing the package. For me, this is typically how I gain familiarity with a new package, I will try to work with it in a separate scratch repo somewhere. But if the idea is more to work within this repo and load EpiAware, as long as it is clear to users how they could do this from their own project repo I think that's fine!

EpiAware/docs/src/showcase/replications/mishra-2020/index.jl
SamuelBrand1: The idea here is that we have a `docs` environment that guarantees that the packages you use are the right ones for our document examples. That is portable, basically you just need to activate the`EpiAware/docs` directory which holds the `Project.toml` file.

EpiAware/docs/src/showcase/replications/mishra-2020/index.jl
SamuelBrand1: Thats a good point... I've come up with a different solution where we grab directly from the repo like here https://turinglang.org/TuringGLM.jl/dev/tutorials/linear_regression/ .

I'll push that, and that makes a lot more sense!

SamuelBrand1:
These are really good points. So _eventually_ `EpiAware` is going to just be on Julia registry and that will be a lot easier.

SamuelBrand1:
> For me, this is typically how I gain familiarity with a new package, I will try to work with it in a separate scratch repo somewhere.

This should work too. You'll need add any other dependencies, but it should work without pinning to specific version. If you could test that out, that would be great.

EpiAware/docs/src/showcase/replications/mishra-2020/index.jl
kaitejohnson: Can you describe what you mean by activate the `EpiAware/docs`?

kaitejohnson:
> These are really good points. So _eventually_ `EpiAware` is going to just be on Julia registry and that will be a lot easier.

Because once it's on Julia registry, does the environment management that you have set up here:

>  docs environment that guarantees that the packages you use are the right ones for our document examples.

Set this up automatically?

EpiAware/docs/src/showcase/replications/mishra-2020/index.jl
```diff
@@ -50,6 +47,29 @@ This tutorial introduces the basic functionality of `EpiAware`. `EpiAware` is a
...
 
 `EpiAware` is built using the [`DynamicPPL`](https://github.com/TuringLang/DynamicPPL.jl) probabilistic programming domain-specific language, which is part of the [`Turing`](https://turinglang.org/dev/docs/using-turing/guide/) PPL. The structural concept behind `EpiAware` is that each module of an epidemiological model is a self-contained `Turing` [`Model`](https://turinglang.org/DynamicPPL.jl/stable/api/#DynamicPPL.Model-Tuple{}); that is each module is an object that can be conditioned on observable data and sampled from. A complete `EpiAware` model is the composition of these objects using the [`@submodel`](https://turinglang.org/DynamicPPL.jl/stable/api/#DynamicPPL.@submodel) macro.
+
+## Pluto code blocks
+Note that `Pluto.jl` notebooks operate on single lines of code or code blocks encapsulated by `let ... end` and `begin ... end`. The reason is that `Pluto` notebooks are reactive, and re-run downstream code after changes with downstreaming determined by a tree of dependent code blocks. If you are running this example from the REPL, you can either use the code blocks and/or run each line individually.
```
seabbs, line 52: I think ideally any information would be on the working with Julia doc or the faq? Otherwise this would need to be repeated on every pluto file.

EpiAware/docs/src/showcase/replications/mishra-2020/index.jl
seabbs: I think this is still to complicated for new users. I think most people would expect that they can copy and paste code into their local  Julia instance (which the easy copying of chunks suggests you can) and run it that way. 

We do have a docs environment but i also don't think we need "guarantees" of the packages as this is a live maintained package and all the docs should just work with the latest packages.

EpiAware/docs/src/showcase/replications/mishra-2020/index.jl
seabbs: No not really. This is all local dev stuff (I.e in the same way as in R) not what you would do with an actual package.

EpiAware/docs/src/showcase/replications/mishra-2020/index.jl
seabbs: Yeah I think aiming for copying and pasting we need to be downloading from a link. Sounds like a good change.

seabbs (commented):
Nice changes but overall I think we want to aim for plug and play (i.e copy and paste interaction) and not requiring a local clone or any mucking about with our environments or local package. 

I think either: 

1. People use it as a pluto notebook and we have instructions for them
2. People copy and paste into their local instance. 

I think those instructions should be in a central doc and we should aim to have as little boiler plate as possible in each doc.

seabbs:
Responding to the rest of the convo here all the current instructions etc just seem overly complicated and hacky. 

All we need is to show people what the dependencies are of a given tutorial in a block and let them install them via copying and pasting imo.

EpiAware/docs/src/showcase/replications/mishra-2020/index.jl
```diff
@@ -1,14 +1,18 @@
 ### A Pluto.jl notebook ###
-# v0.19.40
+# v0.19.43
 
 using Markdown
 using InteractiveUtils
 
 # ╔═╡ 34a06b3b-799b-48c5-bd08-1e57151f51ec
-# hideall
```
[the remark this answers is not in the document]
SamuelBrand1: I think so, Its whatever is hidden from your local Pluto view

EpiAware/docs/src/showcase/replications/mishra-2020/index.jl
seabbs: @SamuelBrand1 I really don't think we need any environment stuff here and can just ask users to install via using  prompts when now they can see the dependencies that this has.

SamuelBrand1:
> All we need is to show people what the dependencies are of a given tutorial in a block and let them install them via copying and pasting imo.

Agreed. I'd independently come to that view as well and just pushed this change.

EpiAware/docs/src/showcase/replications/mishra-2020/index.jl
SamuelBrand1: Yeah good point actually.

SamuelBrand1:
What I've done is to put the information about Pluto code blocks into FAQ, and the extra details about what env we want to use and checking out the branch version of EpiAware into developer docs.

I think this should be usable/readable now.

EpiAware/docs/src/developer/contributing.md
```diff
@@ -39,3 +39,25 @@ As is conventional for Julia packages, unit tests are located at `test/*.jl` wit
...
+    sp = splitpath(@__DIR__)
+    docs_dir = sp |> sp -> sp[1:(findfirst(sp .== "docs"))] |> joinpath
+    pkg_dir = sp |> sp -> sp[1:(findfirst(sp .== "EpiAware"))] |> joinpath
+
+    Pkg.activate(docs_dir)
+    Pkg.develop(; path = pkg_dir)
+    Pkg.instantiate()
+```
+
+This block searchs up the directory path from the location of the `Pluto.jl` script to find the `docs` directory and the `EpiAware` directory, and then activates the environment in the `docs` directory and checks out the `EpiAware` package from the `EpiAware` directory.
```
seabbs, line 63:
```suggestion
This block searches up the directory path from the location of the `Pluto.jl` script to find the `docs` directory and the `EpiAware` directory, and then activates the environment in the `docs` directory and checks out the `EpiAware` package from the `EpiAware` directory.
```

EpiAware/docs/src/developer/contributing.md
```diff
@@ -39,3 +39,25 @@ As is conventional for Julia packages, unit tests are located at `test/*.jl` wit
 Tests that build example package docs from source and inspect the results (end to end tests) are
 located in `/test/examples`. The main entry points are `test/examples/make.jl` for building and
 `test/examples/test.jl` for doing some basic checks on the generated outputs.
+
+### Pluto usage in end to end tests and showcases
+Some of the end to end tests and showcases use `Pluto.jl` scripts.
```
seabbs, line 44: Is this a contribution approach we want to support? Or is this where we say to contribute: 

1. Clone the repo
2. Run the pluto server script we already have
3. Make change to pluto doc? 

i.e do we really need to support this workflow? 

(I also think the pluto in tests are a legacy that we need to remove).

EpiAware/docs/src/getting-started/faq.md
```diff
@@ -10,8 +10,11 @@ Pages = ["lib/getting-started/faq.md"]
 
 We use [`Pluto.jl`](https://plutojl.org/) scripts as part of our documentation and testing. The scripts are located in `docs/src/examples` and can be run using the `Pluto.jl` package.
 We recommend using the version of `Pluto` that is pinned in the `Project.toml` file defining the documentation environment.
+
```
seabbs, line 13: i.e this line below seems like enough if we mention it in the developer docs vs here? 

or mention it here and link out from the developer docs?

EpiAware/docs/src/getting-started/faq.md
```diff
@@ -10,8 +10,11 @@ Pages = ["lib/getting-started/faq.md"]
...
 An entry point to running or developing this documentation is the `docs/pluto-scripts.sh` bash shell script. Run this from the root directory of this repository.
 
+Note that `Pluto.jl` notebooks operate on single lines of code or code blocks encapsulated by `let ... end` and `begin ... end`. The reason is that `Pluto` notebooks are reactive, and re-run downstream code after changes with downstreaming determined by a tree of dependent code blocks. The difference between `let ... end` blocks and `begin ... end` blocks are that the `let ... end` type of code block only adds the final output/return value of the block to scope, like an anonymous function, whereas `begin ... end` executes each line and adds defined variables to scope.
```
seabbs, line 16: can we link out to a tutorial for working with pluto docs

EpiAware/docs/src/showcase/replications/mishra-2020/index.jl
```diff
@@ -52,6 +54,12 @@ It is common to conceptualise the generative process of public health data, e.g
 `EpiAware` is built using the [`DynamicPPL`](https://github.com/TuringLang/DynamicPPL.jl) probabilistic programming domain-specific language, which is part of the [`Turing`](https://turinglang.org/dev/docs/using-turing/guide/) PPL. The structural concept behind `EpiAware` is that each module of an epidemiological model is a self-contained `Turing` [`Model`](https://turinglang.org/DynamicPPL.jl/stable/api/#DynamicPPL.Model-Tuple{}); that is each module is an object that can be conditioned on observable data and sampled from. A complete `EpiAware` model is the composition of these objects using the [`@submodel`](https://turinglang.org/DynamicPPL.jl/stable/api/#DynamicPPL.@submodel) macro.
 "
 
+# ╔═╡ 27d73202-a93e-4471-ab50-d59345304a0b
```
seabbs, line 57: wouldn't this flow more naturally if the dependencies were below/in this section vs above all the text?

seabbs (commented):
Some good updates. I think we are getting there. A few more comments for the developer docs and the showcase structure

EpiAware/docs/src/showcase/replications/mishra-2020/index.jl
SamuelBrand1: They are, or at least are when I render locally which seems to reflect CI.

I'm not sure why the `.jl` file is saved this way round (this might be a legacy of me trying out different order). But the actual display order is found under `# ╔═╡ Cell order:`

This is a case where the `Pluto` notebooks being nearly human readable compared to (say) `ipynb` is maybe a bit confusing?

EpiAware/docs/src/getting-started/faq.md
SamuelBrand1: Do you want Pluto or Pluto as documentation?

EpiAware/docs/src/developer/contributing.md
SamuelBrand1: I think having the Pluto in the showcases makes sense, especially now that we have an ok handle on making it easier to use. But I'm removing reference to testing.

SamuelBrand1:
Replying to a comment above:

> i.e do we really need to support this workflow?

I haven't done this... but an option is to stop using a development approach to `EpiAware` for the pluto based showcases and just install `EpiAware` from main. Additionally, if we want we can pin the dep versions using Pkg rather than relying on finding a local path to the `docs` env.

The main downside is that a change to `EpiAware` in a branch will no longer be represented in the output of the pluto notebook. The main upside would be that each pluto nb we use will be portable; you could download it by itself and it will run locally for you from the Pluto interface without needing to install `EpiAware` or any of the deps.

seabbs:
> I haven't done this... but an option is to stop using a development approach to EpiAware for the pluto based showcases and just install EpiAware from main. Additionally, if we want we can pin the dep versions using Pkg rather than relying on finding a local path to the docs env.

I think we should think about this in a new issue. There are upsides and downsides as you note and I'm not clear what the best choice is or if we can mitigate the downsides of either approach in some way to make the answer clear

seabbs (approved):
This is looking really nice now. @kaitejohnson is you can have another go that would be amazing and report back on how you go. 

IMO these things always benefit from a few rounds of user feedback to get a good pattern set up that most people can get the hang of quickly.

PR #397, opened 2024-07-24
Issue 251: Add aggregation functionality

### LINKED ISSUE #251: Aggregate expected obs

seabbs:
This PR closes #251 by adding a new `Aggregate` struct which sums expected observations based on a supplied aggregation vector (which is first broadcast to the full length). This is similar to the handling of daily -> weekly reporting in `EpiNow2` but more flexible and might be of interest to @jessalynnsebastian (though I think it isn't what we would want when `EpiAware` supports broadcasting expected observations to a reporting triangle.

### DIFF
--- a/EpiAware/src/EpiObsModels/modifiers/Aggregate.jl
+++ b/EpiAware/src/EpiObsModels/modifiers/Aggregate.jl
@@ -42,18 +42,19 @@ end
 
 @model function EpiAwareBase.generate_observations(ag::Aggregate, y_t, Y_t)
     if ismissing(y_t)
-        y_t = Vector{Real}(0.0, length(Y_t))
+        y_t = fill(0.0, length(Y_t))
     end
 
     n = length(y_t)
     m = length(ag.aggregation)
 
-    aggregation = broadcast_n(RepeatEach(), ag.aggregation, n, m)
-    present = broadcast_n(RepeatEach(), ag.present, n, m)
+    aggregation = broadcast_rule(RepeatEach(), ag.aggregation, n, m)
+
+    present = broadcast_rule(RepeatEach(), ag.present, n, m)
 
     agg_Y_t = map(eachindex(aggregation)) do i
         if present[i]
-            exp_Y_t = sum(Y_t[min(1, i - aggregation[i] + 1):i])
+            exp_Y_t = sum(Y_t[max(1, i - aggregation[i] + 1):i])
         else
             exp_Y_t = 0.0
         end
@@ -61,10 +62,11 @@ end
     end
 
     @submodel exp_obs = generate_observations(ag.model, y_t[present], agg_Y_t[present])
-    return _return_aggregate(exp_obs, y_t, present)
+    return _return_aggregate(exp_obs, present, n)
 end
 
-function _return_aggregate(exp_obs, y_t, present)
+function _return_aggregate(exp_obs, present, n)
+    y_t = fill(0.0, n)
     y_t[present] = exp_obs
     return y_t
 end
--- a/EpiAware/src/EpiObsModels/modifiers/Aggregate.jl
+++ b/EpiAware/src/EpiObsModels/modifiers/Aggregate.jl
@@ -42,7 +42,7 @@ end
 
 @model function EpiAwareBase.generate_observations(ag::Aggregate, y_t, Y_t)
     if ismissing(y_t)
-        y_t = fill(0.0, length(Y_t))
+        y_t = Vector{Missing}(missing, length(Y_t))
     end
 
     n = length(y_t)
--- a/benchmark/bench/EpiObsModels/modifiers/Aggregate.jl
+++ b/benchmark/bench/EpiObsModels/modifiers/Aggregate.jl
@@ -1,0 +1,6 @@
+let
+    I_t = fill(10, 100)
+    weekly_agg = Aggregate(PoissonError(), [0, 0, 0, 0, 7, 0, 0])
+    mdl = generate_observations(weekly_agg, I_t, I_t)
+    suite["Aggregate"] = make_epiaware_suite(mdl)
+end
--- a/EpiAware/src/EpiLatentModels/EpiLatentModels.jl
+++ b/EpiAware/src/EpiLatentModels/EpiLatentModels.jl
@@ -23,7 +23,7 @@ export CombineLatentModels, ConcatLatentModels, BroadcastLatentModel
 export RepeatEach, RepeatBlock
 
 # Export helper functions
-export broadcast_dayofweek, broadcast_weekly, equal_dimensions
+export broadcast_rule, broadcast_dayofweek, broadcast_weekly, equal_dimensions
 
 # Export tools for modifying latent models
 export DiffLatentModel, TransformLatentModel, PrefixLatentModel, RecordExpectedLatent
--- a/EpiAware/src/EpiLatentModels/manipulators/broadcast/LatentModel.jl
+++ b/EpiAware/src/EpiLatentModels/manipulators/broadcast/LatentModel.jl
@@ -51,7 +51,6 @@ Generates latent periods using the specified `model` and `n` number of samples.
 
 ## Returns
 - `broadcasted_latent`: The generated broadcasted latent periods.
-- `latent_period_aux...`: Additional auxiliary information about the latent periods.
 
 "
 @model function EpiAwareBase.generate_latent(model::BroadcastLatentModel, n)
--- a/EpiAware/src/EpiLatentModels/manipulators/broadcast/rules.jl
+++ b/EpiAware/src/EpiLatentModels/manipulators/broadcast/rules.jl
@@ -7,7 +7,7 @@ It repeats the latent process at each period. An example of this rule is to repe
 ```julia
 using EpiAware
 rule = RepeatEach()
-latent = [1, 2, 3]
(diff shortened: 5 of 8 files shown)


EpiAware/src/EpiObsModels/modifiers/Aggregate.jl
```diff
@@ -0,0 +1,72 @@
...
+    end
+
+    n = length(y_t)
+    m = length(ag.aggregation)
+
+    aggregation = broadcast_rule(RepeatEach(), ag.aggregation, n, m)
+
+    present = broadcast_rule(RepeatEach(), ag.present, n, m)
+
+    agg_Y_t = map(eachindex(aggregation)) do i
```
SamuelBrand1, line 55: This seems to be type unstable which will hurt without an upside. I'm having a quick look at why type unstability comes in here

EpiAware/src/EpiObsModels/modifiers/Aggregate.jl
```diff
@@ -0,0 +1,72 @@
...
+
+@model function EpiAwareBase.generate_observations(ag::Aggregate, y_t, Y_t)
+    if ismissing(y_t)
+        y_t = Vector{Missing}(missing, length(Y_t))
+    end
+
+    n = length(y_t)
+    m = length(ag.aggregation)
+
+    aggregation = broadcast_rule(RepeatEach(), ag.aggregation, n, m)
```
SamuelBrand1, line 51: There is a slight tension here because `RepeatEach` treats the _last_ repeat iteration as (maybe) truncated, whereas the logic of aggregation is that the _first_ iteration block is truncated.

EpiAware/src/EpiObsModels/modifiers/Aggregate.jl
SamuelBrand1: This might have been a feature of passing `Y_t` as integers, worth looking out for though.

EpiAware/src/EpiObsModels/modifiers/Aggregate.jl
```diff
@@ -0,0 +1,72 @@
...
+
+    present = broadcast_rule(RepeatEach(), ag.present, n, m)
+
+    agg_Y_t = map(eachindex(aggregation)) do i
+        if present[i]
+            exp_Y_t = sum(Y_t[max(1, i - aggregation[i] + 1):i])
+        else
+            exp_Y_t = 0.0
```
SamuelBrand1, line 59:
```suggestion
            exp_Y_t = zero(eltype(Y_t))
```

EpiAware/src/EpiObsModels/modifiers/Aggregate.jl
```diff
@@ -0,0 +1,72 @@
...
+        end
+        return exp_Y_t
+    end
+
+    @submodel exp_obs = generate_observations(ag.model, y_t[present], agg_Y_t[present])
```
SamuelBrand1, line 64: I don't think we need to construct all the agg_Y_t values with `0.0` then take only presents.

Maybe, better to do 

```julia
agg_Y_t = map(findall(present)) do i
    sum(Y_t[max(1, i - aggregation[i] + 1):i])
end
```

EpiAware/src/EpiObsModels/modifiers/Aggregate.jl
```diff
@@ -0,0 +1,72 @@
...
+    present = broadcast_rule(RepeatEach(), ag.present, n, m)
+
+    agg_Y_t = map(eachindex(aggregation)) do i
+        if present[i]
+            exp_Y_t = sum(Y_t[max(1, i - aggregation[i] + 1):i])
+        else
+            exp_Y_t = 0.0
+        end
+        return exp_Y_t
+    end
```
SamuelBrand1, line 62:
```suggestion
    agg_Y_t = map(findall(present)) do i
        sum(Y_t[max(1, i - aggregation[i] + 1):i])
    end
```

EpiAware/src/EpiObsModels/modifiers/Aggregate.jl
SamuelBrand1:
```suggestion
    @submodel exp_obs = generate_observations(ag.model, y_t[present], agg_Y_t)
```

SamuelBrand1 (approved):
This looks good, but I think we can modify the implementation abit e.g. I think we unnecessarily map over too many indices to then slice into specific bits of the array etc.

Also, I noted when trying out the unit tests locally that even with `PoissonError` as the underlying obs error model the output was returning `Float64` eltype. I'm happy to make that an issue and move on, but its not what I expected (should I be expecting something else?).

EpiAware/src/EpiObsModels/modifiers/Aggregate.jl
seabbs: I'm not sure this is the case as the aggregation isn't happening here (but on the below lines) where it is assumed to be right truncated ?

seabbs: (i.e if you look at the test output the first week value is truncated (and is 5) and subsequent weeks (including the last are not truncated (i.e. they are all 7)

EpiAware/src/EpiObsModels/modifiers/Aggregate.jl
seabbs: 1. How did you detect
2. Do we need a test case as this is something very likely users will pass
3. What can we do to mitigate.

seabbs: oh wait sorry you passed Y_t as an integer? Okay I don't think we need to support that do we so perhaps this is just leave as is?

EpiAware/src/EpiObsModels/modifiers/Aggregate.jl
seabbs: you resolve without commiting?

EpiAware/src/EpiObsModels/modifiers/Aggregate.jl
```diff
@@ -0,0 +1,72 @@
...
+
+    @submodel exp_obs = generate_observations(ag.model, y_t[present], agg_Y_t[present])
+    return _return_aggregate(exp_obs, present, n)
+end
+
+function _return_aggregate(exp_obs, present, n)
+    y_t = fill(0.0, n)
```
seabbs, line 64:
```suggestion
    y_t = Vector{eltype(exp_obs)}(zero(eltype(exp_obs)), n)
```

seabbs:
```suggestion
     y_t = Vector{eltype(exp_obs)}(0.0, n)
```

seabbs:
Nice review. I've updated the output to be based on the type of expected obs so I should it should match your expectations now. Just a few open issues to discuss left I think. 

`findall` is good to know about as that definitely makes this cleaner.

EpiAware/src/EpiObsModels/modifiers/Aggregate.jl
SamuelBrand1: Yeah I saw that, this was a bit of idle musing/stream of consciousness based on if we could overload the broadcasting system we have to do `Aggregate` at low computational cost. But I think I'm happy with how we've gone.

EpiAware/src/EpiObsModels/modifiers/Aggregate.jl
SamuelBrand1: > oh wait sorry you passed Y_t as an integer? Okay I don't think we need to support that do we so perhaps this is just leave as is?

Yes, I was trying out the unit tests as an entry (which passed `fill(1, 28)`). I think this is fine tbf.

EpiAware/src/EpiObsModels/modifiers/Aggregate.jl
```diff
@@ -0,0 +1,68 @@
...
+        sum(Y_t[max(1, i - aggregation[i] + 1):i])
+    end
+
+    @submodel pred_obs = generate_observations(ag.model, y_t[present], agg_Y_t)
+    return _return_aggregate(pred_obs, present, n)
+end
+
+function _return_aggregate(pred_obs, present, n)
+    y_t = Vector{eltype(pred_obs)}(undef, n)
+    y_t = fill(zero(eltype(pred_obs)), n)
```
SamuelBrand1, line 65:
```suggestion
    y_t = zeros(eltype(pred_obs), n)
```

This does the trick in one pass I think

benchmark/bench/EpiObsModels/modifiers/Aggregate.jl
```diff
@@ -0,0 +1,6 @@
+let
```
SamuelBrand1, line 1: Love a benchmark :-)

SamuelBrand1 (approved):
LGTM. I've got one suggestion for doing `_return_aggregate` in one pass.

SamuelBrand1 (approved):
Nice.

PR #399, opened 2024-07-25
Add a PR comment that tells readers how to install the current dev version of EpiAware

seabbs:
As the title. I think this could help streamline review and make sure we more often test out changes locally. It might also help those new to Julia begin reviewing.


SamuelBrand1 (approved):
Nice idea!

PR #400, opened 2024-07-25
Remove test/predictive_checking/*

seabbs:
As the title remove out of date predictive checking that is now handled by unit tests. Potentially some of the content here could be moved to either docstrings or documenter docs but I think we should make new issues for that.


SamuelBrand1 (approved):
Good tidying up.

PR #401, opened 2024-07-25
Implement basic scoring as part of inference

### LINKED ISSUE #391: Decide and implement whether to do scoring as post-processing or pre-saving fit objects.

SamuelBrand1:
This PR offers basic CRPS scoring as part of the inference process _without_ outsourcing to `scoringutils`. The advantage of this approach is that we can run it without an `RCall` dependency.

The target time-varying processes for scoring are:

- incidence
- log incidence
- cases
- log cases
- Exponential growth rate
- Reproduction number
- log Reproduction number

The relative time (relative to start of inference period) and the scores at each time point are now contained within the dictionary that is returned by inference configuration. The reference time (i.e. the data horizon for inference) is already recorded there.

**Upsides**: This is a light-weight approach that matches our current pattern of analysis. If we become storage limited in terms of saving all model samples for all model configurations this becomes useful rather than multiple calls to `RCall` and having to set up `RCall` in our compute environment.

**Downsides**: We miss out on the full functionality of `scoringutils`. However, if we _aren't_ storage limited the saved inference outputs still allow us to return to using `scoringutils`.

Closes #391 

### Minor changes
- Updated the returned values from truth data to align with latest version of `EpiAware`.
- Modified the pipeline type to include a `prefix` field, which turns out to be useful compared to previous approach fully dependent on type dispatch.


pipeline/test/end-to-end/test_scoring.jl
```diff
@@ -2,103 +2,38 @@ using DrWatson, Test
...
+    ## plot scores
+
+    score_results = inference_results["score_results"]
+    targets = filter(ky -> ky != "ts", keys(score_results))
+    ts = score_results["ts"]
+    plt_scores = plot(;
+        xlabel = "Time", ylabel = "CRPS", title = "CRPS Scores (relative to max)")
+    for target in targets
+        scores = score_results[target]
+        plot!(plt_scores, ts, scores ./ maximum(scores), label = target)
```
seabbs, line 31: I think we might want to reasses this (scaled to the max) depending on how we want to talk about performance (i.e it doesn't really make sense if we are mostly interested in talking about well performing models.

seabbs (approved):
LGTM. Critically CRPS implementation looks/is correct

seabbs:
I'm a little unclear as to the precise motivation for this PR (as we don't know if we are storage limited) but now its here we might as well use!

PR #414, opened 2024-07-31
Issue 340: Add a prefix to benchmarks

seabbs:
This PR partially addresses #340 by adding a prefix to any warnings produced by `make_turing_suite` based on the input `model`. This should help us locate potential issues more easily. 

@damonbayer if you have time to review this that would be great. We are looking at the output of the `benchmark` CI to see if it has worked.

You can run the `prefix_warnings` wrapper using this example/ 

```julia
# Define a struct instead of a function for g
struct G
    name::String
    threshold::Float64
end

# Define a function that uses the struct G
function process(g::G, x)
    if x < g.threshold
        @warn "Input is below threshold"
    end
    return x^2
end

# Define a function f that uses process with G
function f()
    g = G("MyProcessor", 0.0)
    result = process(g, -5)
    println("Result from process: $result")
    return result
end

# The prefix_warnings function remains the same
function prefix_warnings(f, g)
    original_stderr = stderr
    (read_pipe, write_pipe) = redirect_stderr()
    result = nothing
    try
        result = f()
    finally
        redirect_stderr(original_stderr)
        close(write_pipe)
        output = String(read(read_pipe))
        if !isempty(output)
            println(stderr, "\nWarnings from $(g):")
            println(stderr, output)
        end
    end
    return result
end

# Run the example
g = G("MyProcessor", 0.0)
result = prefix_warnings(f, g)
println("Final result: $result")
```

In principle yes the code in benchmarks should also have tests but I think we can leave for now as I am hoping that the Turing devs put something better than this in upstream.

### DIFF
--- a/benchmark/make_epiaware_suite.jl
+++ b/benchmark/make_epiaware_suite.jl
@@ -3,7 +3,28 @@
 """
 function make_epiaware_suite(model; check = true,
         adbackends = [:forwarddiff, :reversediff, :reversediff_compiled])
-    suite = TuringBenchmarking.make_turing_suite(
-        model; check = check, adbackends = adbackends)
+    suite = prefix_warnings(
+        () -> TuringBenchmarking.make_turing_suite(
+            model; check = check, adbackends = adbackends),
+        model
+    )
     return suite
+end
+
+function prefix_warnings(f, g)
+    original_stderr = stderr
+    (read_pipe, write_pipe) = redirect_stderr()
+    result = nothing
+    try
+        result = f()
+    finally
+        redirect_stderr(original_stderr)
+        close(write_pipe)
+        output = String(read(read_pipe))
+        if !isempty(output)
+            println(stderr, "\nWarnings from $(g):")
+            println(stderr, output)
+        end
+    end
+    return result
 end


benchmark/make_epiaware_suite.jl
```diff
@@ -3,7 +3,28 @@
...
-        model; check = check, adbackends = adbackends)
+    suite = prefix_warnings(
+        () -> TuringBenchmarking.make_turing_suite(
+            model; check = check, adbackends = adbackends),
+        model
+    )
     return suite
 end
+
+function prefix_warnings(f, g)
```
damonbayer, line 14: Can you provide more descriptive variable names?

seabbs:
Annoyingly it looks like we might not get any transcient gradient warnings on this go which makes checking this really is working a bit of a pain!

benchmark/make_epiaware_suite.jl
seabbs: I mean yes in principle we can. I wonder if we need to for an internal function that is stored directly below its use case? That could just be letting it being late here talk though...

damonbayer:
> Annoyingly it looks like we might not get any transcient gradient warnings on this go which makes checking this really is working a bit of a pain!

Doesn't seem wise to approve without a demonstration of it working in CI. Can you force it to warn and then fix it so we see a demonstration?

seabbs:
urggh checking... also it doesn't work :(. I think that this means it is calling the check after the suite has been constructed i.e in benchmarktools which is why wrapping `make_turing_suite` doesn't work. That seems quite tricky to resolve on our end so maybe will have to push that this is a feature `TuringBenchmarking` needs.

seabbs:
So running this https://github.com/CDCgov/Rt-without-renewal/blob/fc41656a63d4002276b89dda94db519b8893fc65/benchmark/bench/EpiObsModels/EpiObsModels.jl#L3 

locally it catches the depreciation warnings as you might expect but that showed no gradient warnings (the target of interest). Running that suite now to see if this is where they show up (vs when building the suite which was how I thought it was set up to work).

seabbs:
So these are rendered stochastically when you make the suite and they are happening in the obs model (which is good to know for the future). It seems like they aren't actually warnings as expected hence not being caught

seabbs:
Hmm weird but they definitely appear to be based on the output of the CI


> ┌ Warning: There is disagreement in the log-density values!
>└ @ TuringBenchmarking ~/.julia/packages/TuringBenchmarking/0wcdT/src/TuringBenchmarking.jl:268
>┌────

benchmark/make_epiaware_suite.jl
damonbayer: Better names and type annotations would help me understand what the code is doing.

seabbs:
Okay so locally I am seeing these (when they happen): 

```julia
Warnings from Model{typeof(generate_observations), (:obs_model, :y_t, :Y_t), (), (), Tuple{NegativeBinomialError{HalfNormal{Float64}}, Vector{Float64}, Vector{Float64}}, Tuple{}, DefaultContext}(EpiAware.EpiAwareBase.generate_observations, (obs_model = NegativeBinomialError{HalfNormal{Float64}}(HalfNormal{Float64}(μ=0.01)), y_t = [10.0, 10.0, 10.0, 10.0, 10.0, 10.0, 10.0, 10.0, 10.0, 10.0], Y_t = [10.0, 10.0, 10.0, 10.0, 10.0, 10.0, 10.0, 10.0, 10.0, 10.0]), NamedTuple(), DefaultContext()):
┌ Warning: There is disagreement in the log-density values!
└ @ TuringBenchmarking ~/.julia/packages/TuringBenchmarking/0wcdT/src/TuringBenchmarking.jl:268
┌ Warning: There is disagreement in the gradients!
└ @ TuringBenchmarking ~/.julia/packages/TuringBenchmarking/0wcdT/src/TuringBenchmarking.jl:275
┌ Warning: There is disagreement in the log-density values!
└ @ TuringBenchmarking ~/.julia/packages/TuringBenchmarking/0wcdT/src/TuringBenchmarking.jl:268
┌ Warning: There is disagreement in the gradients!
└ @ TuringBenchmarking ~/.julia/packages/TuringBenchmarking/0wcdT/src/TuringBenchmarking.jl:275
```

So that is good. I think they aren't showing in the Ci as the thing throwing the warning there is `main`. (which doesn't have this functionality running). Based on that I think this is safe to merge.

benchmark/make_epiaware_suite.jl
seabbs: I'm think this is fine for an internal function but if you feel strongly we can reopen.

seabbs (commented):
As all potential reviewers are on AL self reviewing. Based on the output in benchmarks I think this is working as expected. Merging is low risk as this is the benchmark suite. There approving.

PR #415, opened 2024-08-01
Issue 340: Remove nan handling

seabbs:
This PR explores some of the findings in #340  and #414 which indicate that NegativeBinomialError has unstable gradients when a tape is used. As the majority of the code is shared with PoissonError this indicates the problem is likely to be within the observation_error function. Looking at this branch it seems like the ifelse branch is most likely to be causing issues. 

It isn't clear to me if the lack of nan handling will outweigh any improvement in gradient stability. Really opening this PR for comparative benchmarks in the first instance.

I checked out some model implementations by the Turing folks and when they have implemented different parameterizations they haven't added anything for numerical stability. I've removed all the `clamp` statements to test this.

*I haven't been able to find anything on the stability on `clamp`.*


seabbs:
So the overflow tests semi-obviously fail now but waiting on the benchmarks. The documenter issue appears to be unrelated but unclear

seabbs:
So no warnings in the benchmarks about gradient issues and some evidence of a speed up for NB error models. Not clear if this is worth it if we then have overflow issues? Personally I am minded to remove and see how we go?

SamuelBrand1:
> So no warnings in the benchmarks about gradient issues and some evidence of a speed up for NB error models. Not clear if this is worth it if we then have overflow issues? Personally I am minded to remove and see how we go?

The main problem here is that an overflow error it can be irritating inside of an inference/NUTS run. Obviously, we can put a `try ... catch` block in, but this would v likely also cause gradient issues with reverse + `compile = true`.

A feature of this issue/PR is that the `logpdf` grad calls don't have a problem:

```julia
using Distributions, ReverseDiff, FiniteDifferences

function NegativeBinomialMeanClust(μ, α)
           μ² = μ^2
           ex_σ² = α * μ²
           p = μ / (μ + ex_σ²)
           r = μ² / ex_σ²
           return NegativeBinomial(r, p)
       end

log_μ = 48. #Plausible large value to hit with a log scale random walk over a number of time steps
α = 0.1
nb = NegativeBinomialMeanClust(exp(log_μ), α)

#A few checks
logpdf(nb, 100)
# -427.89305594455965

# Make a helper function for grad calls
f(x) =  NegativeBinomialMeanClust(exp(x[1]), α) |> nb -> logpdf(nb, 100)

# Compare finite diff and rev diff at log_mean = log_μ
g_fd = grad(central_fdm(5, 1), f, [log_μ])[1]

#Precompile ReverseDiff using arbitrary input value to compile the tape
input = randn(1)
const f_tape = ReverseDiff.GradientTape(f, input)
const compiled_f_tape = ReverseDiff.compile(f_tape)
cfg = ReverseDiff.GradientConfig(input)
g_rvd = ReverseDiff.gradient(f, [log_μ], cfg)

g_fd ≈ g_rvd
#true
```

The the overflow issue seems to occur when we make a sample call to a NegativeBinomial with a very large mean (in this example we have `exp(log_μ) = 7.016735912097631e20`).

```julia
rand(nb)
```
> ERROR: InexactError: trunc(Int64, 3.7682298331533404e20)
Stacktrace:
 [1] trunc
   @ ./float.jl:905 [inlined]
 [2] floor
   @ ./float.jl:383 [inlined]
 [3] PoissonADSampler
   @ ~/.julia/packages/Distributions/oGEEM/src/samplers/poisson.jl:54 [inlined]
 [4] PoissonADSampler
   @ ~/.julia/packages/Distributions/oGEEM/src/samplers/poisson.jl:49 [inlined]
 [5] sampler
   @ ~/.julia/packages/Distributions/oGEEM/src/univariate/discrete/poisson.jl:144 [inlined]
 [6] rand(rng::Random.TaskLocalRNG, d::Poisson{Float64})
   @ Distributions ~/.julia/packages/Distributions/oGEEM/src/univariate/discrete/poisson.jl:148
 [7] rand(rng::Random.TaskLocalRNG, d::NegativeBinomial{Float64})
   @ Distributions ~/.julia/packages/Distributions/oGEEM/src/univariate/discrete/negativebinomial.jl:125
 [8] rand(::NegativeBinomial{Float64})
   @ Distributions ~/.julia/packages/Distributions/oGEEM/src/genericrand.jl:22
 [9] top-level scope
   @ REPL[21]:1

Following this stack trace, the `rand` call lowers to a Poisson `rand` call. This has a specialist method that branches depending on the size of mean (for large mean it uses a sampling algo that has a `floor` method after normal sampling). The overflow problem is that `floor` fails with `InexactError` at large values.

This suggests a few different options:
- Flag this problem with `floor` with Julia `Base` maintainers OR flag with `Distributions` depending on what the underlying problem is here.
- Try and isolate why `rand` calls are occurring in inference, when in theory they aren't needed for sampling from the posterior with NUTS.
- Make a custom version of `NegativeBinomial` that uses a relatively inefficient Poisson sampler but which avoids the `floor` issue.

seabbs:
I think we should make an issue upstream and if we can use a sampler that doesn't overflow (this requires us to know if its a sampling context right).

seabbs:
Another option is to make this a true Distribution and to write our own `rand` that doesn't overflow in such an annoying way

seabbs:
I think this is yet another example that Turing needs a sampling context

SamuelBrand1:
> Another option is to make this a true Distribution and to write our own `rand` that doesn't overflow in such an annoying way

This was my thinking.

SamuelBrand1:
I think Turing might be the right place to do this, but I might lower it down to `Distribution.jl`.

https://github.com/TuringLang/Turing.jl/issues/2301

SamuelBrand1:
> Another option is to make this a true Distribution and to write our own `rand` that doesn't overflow in such an annoying way

Upon reflection this is sufficiently low effort I'm going to give this a go.

SamuelBrand1:
#418 resolves the documenter CI problem here by shifting to "safe" version of the distributions.

SamuelBrand1 (approved):
Looks good.

PR #418, opened 2024-08-14
Add `rand` safe version of Poisson and Negative binomial distributions

SamuelBrand1:
This PR contributes a possible solution to #415 by creating distributions that work with `Distributions` but don't trigger a `InexactError` when `rand` is called on them with a large mean value (noting that a large mean value can triggered by random walks on the log-mean). These distributions are:
- `SafePoisson`
- `SafeNegativeBinomial`

Largely this was boilerplate (if we do more of this we should consider a macro to avoid this), with the difference that I replace `floor(Int, x)` calls with a helper function that determines whether `floor(BigInt, x)` should be used rather than throwing an `InexactError`.

## Notable features/limitations
- The poly algo I use for sampling Poissons is from `PoissonRandom` after checking that its still faster than `Distribution.Poisson` sampling.
- For `SafeNegativeBinomial` I'm using the parameterisation with mean $\mu$ and variance relationship:

```math
\text{var} = \mu + \alpha^2 \mu^2.
```

My reasoning here is in the doc string; I think this is the easiest parameterisation to reason on priors because $\alpha \approx \text{std} / \text{mean}$ when the mean is larger than $1/ \alpha$. This differs from `NegativeBinomialMeanClust` so I'm happy to change.
- I've added AD libraries `ReverseDiff` and `ForwardDiff` to the test env. This was because we need to be sure that we can backprop through `logpdf` calls with `SafePoisson` or `SafeNegativeBinomial` distribution structs. I've added unit tests here with very large mean values to check validity in extreme parameter regimes.
- I think `rand` calls are slower to these distributions (but not done large formal analysis).
- If we want to add these distributions then it suggests a number of issues to replace and remove `NegativeBinomialMeanClust`.

I know @damonbayer has been interested in Neg bin safety too.


SamuelBrand1:
Note that this is branched from `remove-nan-handling` and therefore has same documenter CI problem which derives from removing the safety handling of `NegativeBinomialMeanClust`. 

We can either address that with a new issue, or use that as an opportunity to test using the new dists?

seabbs (commented):
I think this looks good and I think we should replace our usage everywhere with this (so that i.e Documenter passes). I also think we should push ideas about reducing copy and paste here into a issue but definitely think it should be investigated as it should help with our partial pooling problems

EpiAware/src/EpiAwareUtils/SafeNegativeBinomial.jl
```diff
@@ -0,0 +1,140 @@
...
+function SafeNegativeBinomial(μ::T, α::T) where {T <: Real}
+    return SafeNegativeBinomial{T}(μ, α)
+end
+
+SafeNegativeBinomial(μ::Real, α::Real) = SafeNegativeBinomial(promote(μ, α)...)
+
+# helper function
+function _negbin(d::SafeNegativeBinomial)
+    μ² = d.μ^2
+    ex_σ² = d.α^2 * μ²
```
seabbs, line 83: I think we should directly paramterise to `NegativeBinomial` and pass to this from `NegativeMeanClust`.

seabbs:
This needs a benchmark before we merge it.

SamuelBrand1:
> This needs a benchmark before we merge it.

If we call `SafeNegativeBinomial` from `NegativeMeanClust` that should be a benchmark test automatically?

EpiAware/src/EpiAwareUtils/SafeNegativeBinomial.jl
SamuelBrand1: Yeah this makes sense.

SamuelBrand1:
I got irritated by a stochastic CI failure on how different the sample variance is from sampling from `SafeNegativeBinomial` vs `NegativeBinomial` so I made it more principled test based on the std of the variance estimator.

SamuelBrand1:
Documenter CI works now.

To recap:
- `SafeNegativeBinomial` and `SafePoisson` resolve a problem with `floor` inside `rand` calls when the sampler mean is very large.
- Setting `check_args = false` as a kwarg for `Distribution` object construction switches off checking that causes occasion fails. The NUTS sampler seems to be smart enough to handle resultant `logpdf` values.

seabbs (approved):
Nice looks good. As we discussed its a lot of copy and paste so ideally we should have an issue to track changes we can make to reduce this (either here or upstream).

Issue #406, opened 2024-07-25
Consider moving from broadcasting to weekly to daily increments in underlying latent processes


SamuelBrand1:
We've been guided by the common approach of weekly (i.e. 7 time step) broadcasting for $R_t$, and done the same for `DirectInfections` and `ExpGrowth` infection generating processes. However, they make a bit less sense in those contexts IMO.

For consistency we should consider moving all latent models used in the pipeline to not having broadcasting.

The obvious downside is that this mean approximately x7 parameters to do inference upon, and maybe a switch to reverse-mode AD?

SamuelBrand1:
It looks like #415 could be resolving #340 i.e. numerical instability with compiled tape AD. Conditional on this, that increases the upside of this change by diminishing the down side of swapping to reverse mode (e.g. we can use compiled tape to be faster).

PR #427, opened 2024-08-16
Remove most pipeline compat and add a fetch call in testing

SamuelBrand1:
Small PR ahead of PR to address #406 . 

Informed by working on `compat` for `EpiAware`, I noted that the compat requirements in the pipeline had drifted a bit (e.g. fixing `ADTypes` to 1.2). My current thinking is that as a _project_ having compat requirements for `EpiAwarePipeline` is not appropriate since the version of the packages used is fixed anyway.

Also, I noted that the pipeline run tests were passing _before_ running had finished. I realised that this was because the execution was not waiting for the `EagerThunk` to resolve before resolving the test outcome. I've added a `Dagger.fetch` call to force the process to wait (this didn't change test from pass to fail but might in the future).


seabbs (approved):
Nice good spot

PR #430, opened 2024-08-16
Move latent models from broadcasting to weekly to daily increments and change default AD to reverse-mode

### LINKED ISSUE #406: Consider moving from broadcasting to weekly to daily increments in underlying latent processes

### LINKED ISSUE #424: Add Changelog to manuscript
As per title.

SamuelBrand1:
As per title and discussion in #406 . This closes #406 .

This variation in the analysis plan is recorded in a change log at `manuscript/changelog.md` and creating this file closes #424 .


seabbs (approved):
LGTM.

Issue #435, opened 2024-08-20
Bug: Mishra et al replication error


SamuelBrand1:
Unfortunately, our replication of Mishra et al has failed silently. 

From local experimentation, the chance of the sampler getting stuck in a poor part of parameter regime is removed by increasing the `nadapts` steps in warm-up along with a higher target acceptance rate.

PR #436, opened 2024-08-20
Attempt to fix replication issue

SamuelBrand1:
This PR tries to fix the issue #435 by reverting pathfinder bump (which also reverts Turing). I find this very sub-optimal and would prefer to merge a different solution.


seabbs:
Do we know what the cause of this was now? is it the Turing update (as we saw it both with and without pathfinder)? If so do we have any ideas for what to do about this?

SamuelBrand1:
> Do we know what the cause of this was now? is it the Turing update (as we saw it both with and without pathfinder)? If so do we have any ideas for what to do about this?

I'm afraid not... reversing the Pathfinder update 0.8 -> 0.9 seems to have fixed the wild error we saw. However, I don't really see what the problem was. I'd prefer to use the more update version and isolate the error.

PR #437, opened 2024-08-27
Explore using self contained pluto notebooks for documentation

### LINKED ISSUE #396: Explore using self-contained Pluto notebooks for documentation
> i.e do we really need to support this workflow?

I haven't done this... but an option is to stop using a development approach to `EpiAware` for the pluto based showcases and just install `EpiAware` from main. Additionally, if we want we can pin the dep versions using Pkg rather than relying on finding a local path to the `docs` env.

The main downside is that a change to `EpiAware` in a branch will no longer be represented in the output of the pluto notebook. The main upside would be that each pluto nb we use will be portable; you could download it by itself and it will run locally for you from the Pluto interface without needing to install `EpiAware` or any of the deps.

_Originally posted by @SamuelBrand1 in https://github.com/CDCgov/Rt-without-renewal/issues/395#issuecomment-2247668692_

SamuelBrand1:
A draft PR exploring the use of self contained Pluto notebooks.


seabbs (commented):
I like the aim here (to make things independent). I am a little unsure we want to throw away the package manager utility.

Any help here: https://discourse.julialang.org/t/using-unregistered-packages-with-pluto/72483/4

and here: https://github.com/fonsp/Pluto.jl/discussions/1558

specifically Pluto.activate_notebook_environment

This seems like the ultimate solution here: https://github.com/fonsp/Pluto.jl/issues/1304#issuecomment-1983629828

i.e. use `[Source]` in `Project.toml`: https://github.com/JuliaLang/Pkg.jl/pull/3783

I think to get this to work you would need to manually edit the internal Project.toml of each Pluto doc (i.e. as here for the compact rec: https://discourse.julialang.org/t/update-project-toml-compat-entries-in-pluto/86396/3) which is not great but I guess will work.

seabbs:
Shall we close this PR as I think we agree the correct approach is to use source? 

Using: 

> I think to get this to work you would need to manually edit the internal Project.toml of each Pluto doc (i.e. as here for the compact rec

Or shall we update doing this? 

I think this still leaves an issue in that it won't just work for people so potentially we want to regex in the source URL vs using a relative path (so that the pluto truely stands alone).

SamuelBrand1:
> Shall we close this PR as I think we agree the correct approach is to use source?
> 
> Using:
> 
> > I think to get this to work you would need to manually edit the internal Project.toml of each Pluto doc (i.e. as here for the compact rec
> 
> Or shall we update doing this?
> 
> I think this still leaves an issue in that it won't just work for people so potentially we want to regex in the source URL vs using a relative path (so that the pluto truely stands alone).

Agreed.

PR #441, opened 2024-08-28
Change ARStep to reverse AR parameters

### LINKED ISSUE #440: AR parameters in reverse order
We have a usual dot-product-on-recent state pattern for AR [here](https://github.com/CDCgov/Rt-without-renewal/blob/f5b4ab7bdbb21f069b5d178bd365373e7aff8635/EpiAware/src/EpiLatentModels/models/AR.jl#L99). However, in the constructor it seems more natural to define the `damp_priors` $\psi_1 \sim \pi_1,\dots,\psi_p \sim \pi_p$ in the usual order, rather than reverse order (that how it is in the replication note).

I think just putting reverse into the `ARStep` constructor is the best way here. Noting that `Renewal` should follow this convention too for style unity across `EpiAware`. This can be address by addressing #384 (`Renewal` is only place that uses `scan` now.)

SamuelBrand1:
This fixes #440 by choosing to reverse the AR parameters, whilst maintaining everything else as is.

To cover this bug I've also added a unit test that would have picked up swapping the AR parameters in an AR(2) process by comparing a long single draw to its known theoretical stationary distribution.

Closes #440

### DIFF
--- a/EpiAware/src/EpiLatentModels/models/AR.jl
+++ b/EpiAware/src/EpiLatentModels/models/AR.jl
@@ -81,23 +81,26 @@ Generate a latent AR series.
     damp_AR ~ latent_model.damp_prior
     ϵ_t ~ filldist(Normal(), n - p)
 
-    ar = accumulate_scan(ARStep(damp_AR), ar_init, σ_AR * ϵ_t)
+    ar_step = ARStep(reverse(damp_AR))
+    ar = accumulate_scan(ar_step, ar_init, σ_AR * ϵ_t)
 
     return ar
 end
 
 @doc raw"
-The autoregressive (AR) step function struct
+The autoregressive (AR) step function struct.
+
+Note that the AR parameters are stored in reverse order.
 "
 struct ARStep{D <: AbstractVector{<:Real}} <: AbstractAccumulationStep
-    damp_AR::D
+    rev_damp_AR::D
 end
 
 @doc raw"
 The autoregressive (AR) step function for use with `accumulate_scan`.
 "
 function (ar::ARStep)(state, ϵ)
-    new_val = dot(ar.damp_AR, state) + ϵ
+    new_val = dot(ar.rev_damp_AR, state) + ϵ
     new_state = vcat(state[2:end], new_val)
     return new_state
 end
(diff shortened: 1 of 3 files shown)


EpiAware/src/EpiLatentModels/models/AR.jl
```diff
@@ -81,23 +81,26 @@ Generate a latent AR series.
     damp_AR ~ latent_model.damp_prior
     ϵ_t ~ filldist(Normal(), n - p)
 
-    ar = accumulate_scan(ARStep(damp_AR), ar_init, σ_AR * ϵ_t)
+    ar_step = ARStep(reverse(damp_AR))
```
seabbs, line 84: For efficieny I don't think we want to do this at run time?

SamuelBrand1: To avoid doing `reverse`? That would suggest reversing at the constructor level I think.

SamuelBrand1: e.g. we add a `rev_damp_prior` field to `AR` and have `damp_AR ~ latent_model.rev_damp_prior`. Obvs we could just reverse `damp_prior` but that might get confusing for users?

SamuelBrand1:
@seabbs had a good shout about not doing the reverse at run time. I've added a `rev_damp_priors` field to `AR`. The only real gotcha now is that you have to condition on `rev_damp_prior` parameter if you want to do conditional AR things now, but I think this is an ok price for not doing a `reverse` call every time we call the model.

seabbs:
Nice change. I think we should pause on this until #438 is in (ideally today) as that changes the names and structure of a lot of the AR model.

SamuelBrand1:
> Nice change. I think we should pause on this until #438 is in (ideally today) as that changes the names and structure of a lot of the AR model.

Right. But I'm going to merge this into some other branches because this is the actual likely problem which makes #436 hard to resolve (I think!).

seabbs:
ah okay. Shall we just merge this in then and I will deal with it in #438

seabbs (commented):
I'm quite unclear why this would have a performance impact?

SamuelBrand1:
> I'm quite unclear why this would have a performance impact?

It shouldn't in general (right?) but in the replication notebook we effectively had strong priors on the AR parameters in the wrong order (which strongly effects the prior variance of the process etc).

seabbs:
This is blocked by #438

seabbs:
This is unblocked now but because of all the annoying changes from me we might want to close and start again

SamuelBrand1:
Agreed!

Issue #420, opened 2024-08-14
Make Mishra et al replication a full replication


SamuelBrand1:
At the moment the latent AR process for $\log R_t$ showcase of _Mishra et al_ is still really showcase of the broadcasting of `EpiAware`. Demoing that should be a different part of the docs. I propose actually showcasing the replication by moving the _getting started with epiaware_ stuff away and replicating the daily increments in the AR process.

seabbs:
Yes, strongly agree

SamuelBrand1:
Closed by #442

PR #442, opened 2024-08-28
Fix the replication note and turn it into a full replication

### LINKED ISSUE #435: Bug: Mishra et al replication error

SamuelBrand1:
This PR builds on #441 as well as rewriting the Mishra et al replication note to fully replicate that pre-print.

In doing this I think it became clear that the structure of the model makes it hard to identify the AR process std parameter (the other parameters have strong priors) from the neg bin observation cluster factor. In the preprint no prior for neg bin was found (apols if I missed it!). Therefore, I demonstrated conditioning the value and running the inference on the model with conditioned cluster factor.

Closes #435 and #420.

### DIFF
--- a/EpiAware/docs/src/showcase/replications/mishra-2020/index.jl
+++ b/EpiAware/docs/src/showcase/replications/mishra-2020/index.jl
@@ -543,24 +543,24 @@ let
         c = :black,
         lab = "prior")
 
-    p3 = histogram(inference_results.samples["latent.rev_damp_AR[2]"],
+    p3 = histogram(inference_results.samples["latent.damp_AR[2]"],
         lab = "chain " .* string.([1 2 3 4]),
         fillalpha = 0.4,
         lw = 0,
         norm = :pdf,
         title = "Posterior dist: rho_1")
-    plot!(p3, ar.damp_prior.v[1],
+    plot!(p3, ar.damp_prior.v[2],
         lw = 3,
         c = :black,
         lab = "prior")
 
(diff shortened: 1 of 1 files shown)


seabbs (commented):
I really like this rewrite. I am not clear why it needed to be built on top of #441 though?

SamuelBrand1:
> I really like this rewrite. I am not clear why it needed to be built on top of #441 though?

We had strong priors on the AR parameters in the wrong order (which strongly effects the prior variance of the process etc). Obviously, I could have just swapped them (as well as the other changes) but that felt silly because we'd have just had to swap back after #441.

seabbs:
I'm still a bit uncomfortable about #441. I know its annoying but any chance we can decouple?

SamuelBrand1:
> I'm still a bit uncomfortable about #441. I know its annoying but any chance we can decouple?

Yeah, sure. Probably easiest to wait until we've decided what we want with `AR` then merge that into this?

SamuelBrand1:
Ok, I `git reset` the commits, reversed the notebook `damp_prior` vector for correctness and force pushed that back.

seabbs:
ty - sorry for being a pain

seabbs (approved):

seabbs (dismissed):

seabbs (approved):

Issue #439, opened 2024-08-28
renaming step functions to iteration or update functions?


SamuelBrand1:
Minor issue, but I think us using step functions as terminology for the iteration/update steps in `accumulate_scan` is possibly a bit confusing? Because "step function" is a pretty well known maths concept (https://en.wikipedia.org/wiki/Step_function), and we are using it in a different context e.g. "step-by-step". 

This might just be reflecting my personal biases!

SamuelBrand1:
e.g. `AbstractAccumulationStep` -> `AbstractAccumulationUpdate`

seabbs:
good point. Would your suggestion be `iterator` or `updator` or other?

SamuelBrand1:
My suggestion would be `update` or `updater` because `iterator` is a bit overloaded in Julia.

seabbs:
so I don't think we should use update as we are likely to want a function called update to i.e update turing models (and in my head update means update an existing object). I don't love that it doesn't clearly identify the iterative process. Hmm

SamuelBrand1:
Some thing with `recurrence` or `recurrent` or an abbreviation?

seabbs:
hmm maybe

SamuelBrand1:
Its the same operation as a recurrent NN (just with a dot product rather than a NN call)... so at least some people would find it intuitive description.

seabbs:
good point

SamuelBrand1:
Have we in fact just settled on `Steps`?

seabbs:
I think we have more paused thinking about it until we have a better vision for a more generic composable iterator vision

Issue #408, opened 2024-07-28
EpiSteps submodule


seabbs:
We should consider moving the `accumulate_scan`,`ModelSteps`, and `get_state` functionality into its own submodule. At the same time we should think about the changes needed to make these utilities composable so they can represent say renewal processes with susceptible depletion and also ARIMAs etc from the same code base. I think we should look to do this after #407 is merged as this will give more examples to work from (i.e to inform the design).

seabbs:
A potential pattern that could work here for composing steps is the following: 

```julia
function compose_steps(steps...)
    return (state, ϵ) -> foldl((s, step) -> step(s, ϵ), steps, init=state)
end

# Usage
arma_step = compose_steps(ar_step, ma_step)
result = accumulate_scan(arma_step, initial_state, ϵ_t)
```

A more manual approach to this would be: 

```julia
struct CompositeStep{T<:Tuple{Vararg{AbstractAccumulationStep}}} <: AbstractAccumulationStep
    steps::T
end

function (cs::CompositeStep)(state, ϵ)
    for step in cs.steps
        state = step(state, ϵ)
    end
    return state
end

# Usage example
ar_step = ARStep([0.5, 0.3])
ma_step = MAStep([0.2, 0.1])
arma_step = CompositeStep((ar_step, ma_step))

result = accumulate_scan(arma_step, initial_state, ϵ_t)
```

seabbs:
Some obvious questions that come to mind are that different states can be of different lengths (i.e a p length AR and a q length MA). If  the same state is passed then it would need to be the longest available one and only the last step in the iterator should drop a value from the state. 

It might therefore be simpler if we constrain the design such that all states are independently tracked but this would mean that the values would only represent those at the current step of the composition vs the final value).

seabbs:
In some sense I think this suggests detecting the state updating from the iterative process of applying the steps i.e it is 

1. Buildout recursive functions that each track a state and take inputs from the initial values
2. Add a final layer that increments the state of each recursive step (i.e each step is a 3 things 1. Update a new value 2. Update the state 3. Extract the state of interest).
3. Apply these using accumulate_scan to each step
4. Extract the states of interest to rebuild the timeseries

seabbs:
[...]

PR #444, opened 2024-08-29
Replace scan usage with accumulate scan

### LINKED ISSUE #384: Replace scan usage with accumulate_scan
In #369 we added new functionality for the `AR` latent model that introduce an abstract approach to extending `Base.accumulate` via step functions. This can be used to replace our current usage of the `scan` function. 

This is likely to be attracted as it uses a base function which will likely be a major target of Auto-diff optimisations. It also makes it easier for us to introduce composability to our `scan` functions (i.e #385).

SamuelBrand1:
This PR primarily changes usage of `scan` in Renewal model to using `accumulate_scan`.

To do this meant defining callable `Step` structs. The relevant design decisions I made here were:

- To create a `AbstractConstantRenewalStep <: AbstractAccumulationStep` type to group recurrence/step functions that are intended to be static. 
- These become a field in `Renewal`, indicating that for this kind of modelling inference of the GI is not in scope, and avoiding redeclaring the Step at runtime (hence `ConstantStep`).
- This allowed me to ditch the `RenewalWithPopulation` struct type, this behaviour is now defined by the `ConstantRenewalWithPopulationStep` struct.
- I've swapped to reversing the generation interval (I was doing essentially the "left-hand" version of Renewals before which is possible suboptimal).

Closes #384 .


EpiAware/src/EpiInfModels/Renewal.jl
```diff
@@ -31,9 +31,11 @@ number.
 `initialisation_prior` for the prior distribution of ``\hat{I}_0``. The default
 `initialisation_prior` is `Normal()`.
 
-## Constructor
+## Constructors
 
-- `Renewal(; data, initialisation_prior)`.
+- `Renewal(; data, initialisation_prior)`. Construct a `Renewal` model with default update steps.
+- `Renewal(data; initialisation_prior)`. Construct a `Renewal` model with default update steps.
+- `Renewal(data, initialisation_prior, renewal_step)` Construct a `Renewal` model with `renewal_step` update step function.
```
seabbs, line 38: if its in a renewal struct does it need to be called renewal step?

EpiAware/src/EpiInfModels/RenewalSteps.jl
```diff
@@ -0,0 +1,125 @@
...
+and `g_i` is the generation interval.
+
+# Arguments
+- `recent_incidence`: Array of recent incidence values order least recent to most recent.
+- `Rt`: Reproduction number.
+
+# Returns
+- Updated incidence array.
+"""
+function (renewal_step::ConstantRenewalStep)(recent_incidence, Rt)
```
seabbs, line 39: My hope is we can do this just with the AR step in the future

EpiAware/src/EpiInfModels/RenewalSteps.jl
```diff
@@ -0,0 +1,125 @@
...
+    return last.(state)
+end
+
+@doc raw"
+The renewal process iteration/step function struct with constant generation interval and a fixed
+population size.
+
+Note that the generation interval is stored in reverse order.
+"
+struct ConstantRenewalWithPopulationStep{T} <: AbstractConstantRenewalStep
```
seabbs, line 66: my hope that we can start composing these vs duplicating in #408

seabbs (commented):
This is a really nice step forward. I think we can iterate on this to make the renewal steps use more of the latent model steps and to also be more composable but I think this is a great first step. The only outstanding comment here is that the name in the struct might be a bit duplicative everything else is really a musing.

EpiAware/src/EpiInfModels/RenewalSteps.jl
SamuelBrand1: That would be very neat.

SamuelBrand1: It would be interesting to start reasoning on this. Both `AR`, `MA` and `Renewal` etc have/can be represented as having a convolution on recent variables. Otoh, the "noise" process is different.

Is this related to the compositional idea, e.g. one "Step" for doing a conv-like operation and another that specialises?

EpiAware/src/EpiInfModels/Renewal.jl
SamuelBrand1: I'm interpreting this as a comment on argument naming rather than the naming of the type system? The type system is pretty verbose, but personally I like that.

EpiAware/src/EpiInfModels/RenewalSteps.jl
SamuelBrand1: I've renamed to `recurrent_step` following our discussion in #439

seabbs:
There was an outstanding issue with renewal models that meant we couldn't benchmarks so no useful benchmarks to compare against. We should add some benchmarks in in another PR.

seabbs (approved):
LGTM

Issue #404, opened 2024-07-25
Upgrade the "testmode" pipeline script to just pipeline script


SamuelBrand1:
At the moment, the "testmode" script https://github.com/CDCgov/Rt-without-renewal/blob/main/pipeline/scripts/run_testmode_pipeline.jl is very close to the actual pipeline script. I think it just needs renamed, but maybe we want some more functionality?

At the moment, you use from the command line as:

```bash
julia path\to\script ndraws
```
and it will sample `ndraws` for each scenario in each pipeline.

PR #445, opened 2024-08-30
Upgrade the `testmode` pipeline script to pipeline scripts: prior predictive + full

SamuelBrand1:
This Draft PR has the improvements I'm making to the pipeline towards resolving #404. There were quite a number of improvements _before_ I noticed that the forecast function was now broken:

- `pipeline` testmode boolean field which is useful for the test running.
- prefixes also determined by the pipeline.
- "testmode" pipeline script upgraded to main script.
- Better end-to-end tests.

This is tested using the updated `EpiAware` hash [9c52763](https://github.com/CDCgov/Rt-without-renewal/pull/444/commits/9c527635c82863e75d40d2e8a6786fcef33c2fa8).

Currently forecast function in pipeline is still broken, but this is a good place to fix from.


SamuelBrand1:
I've merged in removal of all basic plots in favour of `CairoMakie` plotting under self-review.

SamuelBrand1:
Steps towards full PR:

- [ ] Prior predictive checks
- [ ] use findings of above to guide more informative priors
- [ ] check that inference methods don't fail under better priors.

SamuelBrand1:
This branch now runs both the prior predictive and the full pipelines, with in cases of numerical error in any of the steps it serialises the `Exception` string return.

Since, it is no longer broken code, but rather improvable code I propose merging into `main`.

seabbs (approved):

Issue #306, opened 2024-06-27
Adding an MAP optimization `EpiMethod`; possibly tempered


SamuelBrand1:
Following the issues with #238 we could experiment with initialisation of `pathfinder`.

I've had success with pre-heating the pre-heating of `pathfinder` but doing a sequence of tempered MAP optimizations to initialize `pathfinder`, which subsequently gives an informed starting point for NUTS metric selection e.g. 

https://github.com/cdcent/cfa-forecast-mechanistic-julia/blob/main/src/inference-methods/map-optimization.jl

seabbs:
Noting that this could be a fun issue for a new contributor...

SamuelBrand1:
Some recent developments: Turing 0.33 has a nicer interface to mode optimization https://turinglang.org/docs/tutorials/docs-17-mode-estimation/ so this should be nicer to do.

PR #464, opened 2024-10-02
Add chatzilena et al as a replication example

### LINKED ISSUE #446: Add Chatzilena et al as a replication example
This paper https://www.sciencedirect.com/science/article/pii/S1755436519300325 was really good (IMO) at introducing Bayesian leaning mechanistic epi modellers to NUTS/ADVI. I think this would be a good next replication project. 

It would also be a nice place to try out SciML DEProblem interface into EpiAware.

SamuelBrand1:
This PR replicates two parts of the demonstration of epidemic inference using stan from [_Chatzilena et al 2019_](https://www.sciencedirect.com/science/article/pii/S1755436519300325):

- The "deterministic" SIR model fit to a classic boarding school influenza data set. This is a SIR model with Poisson errors.
- The "Stochastic" SIR model fit to same data. This is a SIR model with an autoregressive process on the log-residuals.

In both cases, I use a combination of SciML and EpiAware functionality which worked well.

Closes #446 

## Aside comments

The prior predictive checking (not done in Chatzilena et al) reveals very poor prior specification in the sense that I don't think anyone would reasonably expect the prior marginal median for numbers of ill children. 

This means that I use a combination of MLE and MAP optimisation to start the NUTS sampling in reasonable parts of parameter space; in practise I would change the prior specifications but I felt this deviated from replication/comparison.

Doing the pre-sampling optimisation means that we are getting to reasonable ess is about 10-60x faster than reported in the paper (acknowledging here that the paper is 4 years old and other confounding factors). However, without this optimization the method is prone to stochastic failures. Therefore, pre-sampling steps are quite important to consider going forwards if we use this kind of approach in applications.


seabbs:
> Therefore, pre-sampling steps are quite important to consider going forwards if we use this kind of approach in applications.

I think this is a generally true observation in most real-world use cases. Do we already have a initialisation method that samples from the prior (I forget but can check)

seabbs:
I see the following Documenter error: 


Code:
  begin
      using EpiAware.EpiAwareUtils: censored_pmf, censored_cdf, ∫F
      using Random, Distributions, StatsBase #utilities for random events
      using DataFramesMeta #Data wrangling
      using CairoMakie, PairPlots #plotting
      using Turing #PPL
  end

Error:
  ArgumentError: Package StatsBase not found in current path.
  - Run `import Pkg; Pkg.add("StatsBase")` to install the StatsBase package.
  Stacktrace:
  
  FYI I think we nearly critically need to move towards pluto wrapped notebooks vs the doc level environment as this kind of thing is easy to do.

seabbs:
I think you need to add this to pages.jl (perhaps we can automate this at a later date to avoid this step)

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
```diff
@@ -0,0 +1,622 @@
...
+
+Alongside the `EpiAware` package we will use the `OrdinaryDiffEq` and `SciMLSensitivity` packages for interfacing with `SciML` ecosystem; this is a lower dependency usage of `DifferentialEquations.jl` that, respectively, exposes ODE solvers and adjoint methods for ODE solvees; that is the method of propagating parameter derivatives through functions containing ODE solutions. Bayesian inference will be done with `NUTS` from the `Turing` ecosystem. We will also use the `CairoMakie` package for plotting and `DataFramesMeta` for data manipulation.
+"
+
+# ╔═╡ 943b82ec-b4dc-4537-8183-d6c73cd74a37
+md"
+# Single population SIR model
+
+As mentioned in _Chatzilena et al_ disease spread is frequently modelled in terms
+of ODE-based models. The study population is divided into compartments representing a specific stage of the epidemic status. In this case, susceptible, infected, and recovered individuals.
```
seabbs, line 74: is this line break intentionnal?

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
```diff
@@ -0,0 +1,622 @@
...
+"
+
+# ╔═╡ 0e78285c-d2e8-4c3c-848a-14dae6ead0a4
+md"
+We can interface to the `SciML` ecosystem by writing a function with the signature:
+
+> `(du, u, p, t) -> nothing`
+
+Where:
+- `du` is the _vector field_ of the ODE problem, e.g. ${dS \over dt}$, ${dI \over dt}$ etc. This is calculated _in-place_.
```
seabbs, line 97:
```suggestion
- `du` is the _vector field_ of the ODE problem, e.g. ${dS \over dt}$, ${dI \over dt}$ etc. This is calculated _in-place_ (commonly denoted using ! in function names in Julia).
```

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
```diff
@@ -0,0 +1,622 @@
...
+    du[1] = -β * I * S
+    du[2] = β * I * S - γ * I
+    du[3] = γ * I
+
+    return nothing
+end
+
+# ╔═╡ f16eb00b-2d77-45df-b767-757fe2f5674c
+md"
+We combine vector field function `sir!` with a initial condition `u0` and the integration period `tspan` to make an `ODEProblem`. We do not define the parameters, these will be defined within an inference approach.
```
seabbs, line 118: Suggest add a link out to `EpiProblem` concept we use.

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
```diff
@@ -0,0 +1,622 @@
...
+"
+
+# ╔═╡ d64388f9-6edd-414d-a191-316f75b35b2c
+md"
+
+## Data for inference
+
+There was a brief, but intense, outbreak of Influenza within the (semi-) closed community of a boarding school reported to the British medical journal in 1978. The outbreak lasted from 22nd January to 4th February and it is reported that one infected child started the epidemic and then it spread rapidly. Of the 763 children at the boarding scholl, 512 became ill.
+
+We downloaded the data of this outbreak using the R package `outbreaks` which is maintained as part of the [R Epidemics Consortium(RECON)](http://www. repidemicsconsortium.org).
```
seabbs, line 128: maintained is doing a lot of work here

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
```diff
@@ -0,0 +1,622 @@
...
+data = "https://raw.githubusercontent.com/CDCgov/Rt-without-renewal/refs/heads/446-add-chatzilena-et-al-as-a-replication-example/EpiAware/docs/src/showcase/replications/chatzilena-2019/influenza_england_1978_school.csv2" |>
+       url -> CSV.read(download(url), DataFrame) |>
+              df -> @transform(df,
+    :ts=(:date .- minimum(:date)) .|> d -> d.value + 1.0,)
+
+# ╔═╡ aba3f1db-c290-409c-9b9e-6065935ede54
+N = 763;
+
+# ╔═╡ bb07a580-6d86-48b3-a79f-d2ed9306e87c
+sir_prob = ODEProblem(
```
seabbs, line 142: is it useful here to talk a bit about the ODEProblem object or link somewhere that does

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
```diff
@@ -0,0 +1,622 @@
...
+md"
+From `EpiAware`, we have the `PoissonError` struct which defines the probabilistic structure of this observation error model.
+"
+
+# ╔═╡ 87509792-e28d-4618-9bf5-e06b2e5dbe8b
+obs = PoissonError()
+
+# ╔═╡ 81501c84-5e1f-4829-a26d-52fe00503958
+md"
+Now we can write the observation model using the `Turing` PPL.
```
seabbs, line 179: is it not the probabliistic or full model (which contains the obs model)

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
```diff
@@ -0,0 +1,622 @@
...
+    )
+
+    ##Solve remade ODE model##
+
+    sol = solve(_prob, solver;
+        saveat = ts,
+        verbose = false)
+
+    ##log-like accumulation using obs##
+    λt = N * sol[2, :] .+ upjitter #expected It
```
seabbs, line 205: is the jitter needed here I thought we had handled that with the obserror types?

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
```diff
@@ -0,0 +1,622 @@
...
+    @submodel obsYt = generate_observations(obs, Yt, λt)
```
seabbs, line 206: obsYt and following doesn't really follow our variable naming conventions. I really don't think this is a big deal FYI but if want to change to _ I would do so

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
```diff
@@ -0,0 +1,622 @@
...
+
+    ##Generated quantities##
+    return (; sol, obsYt, R0 = β / γ)
+end
+
+# ╔═╡ e7383885-fa6a-4240-a252-44ae82cae713
+md"
+We instantiate the model in two ways:
+
+1. `deterministic_mdl`: This conditions the generative model on the data observation. We can sample from this model to find the posterior distribution of the parameters.
```
seabbs, line 216: Again suggest link out to other docs we have

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
```diff
@@ -0,0 +1,622 @@
...
+end
+
+# ╔═╡ 4c0759fb-76e9-4de5-9206-89e8bfb6c3bb
+md"
+The prior predictive checking suggests that _a priori_ our parameter beliefs are very far from the data. Approaching the inference naively can lead to poor fits.
+
+We do three things to mitigate this:
+
+1. We choose a switching ODE solver which switches between explicit (`Tsit5`) and implicit (`Rosenbrock23`) solvers. This helps avoid the ODE solver failing when the sampler tries extreme parameter values. This is the default `solver = AutoTsit5(Rosenbrock23())` above.
+2. To avoid the effect of numerically negative small values of `λt` we add a small `upjitter`.
```
seabbs, line 284: sounds like my above question answer is no

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
```diff
@@ -0,0 +1,622 @@
...
+3. We locate the maximum likelihood point, that is we ignore the influence of the priors, as a useful starting point for `NUTS`.
```
seabbs, line 285: another good place for a callout to elsewhere in our docs

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
```diff
@@ -0,0 +1,622 @@
...
+"
+
+# ╔═╡ 8d96db67-de3b-4704-9f54-f4ed50a4ecff
+nmle_tries = 100
```
seabbs, line 289: this makes me sad. Can't we do this using our EpiProblem framework? I think this might be out of scope for this PR and be a new issue as I can see it needing a feature?

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
```diff
@@ -0,0 +1,622 @@
...
+    (1 - exp(-2 * ϕ)) * σ² / (2 * ϕ)
+end
+
+# ╔═╡ 89c767b8-97a0-45bb-9e9f-821879ddd38b
+md"
+We define the AR(1) process by matching means of `HalfNormal` prior distributions for the damp parameters and std deviation parameter to the calculated the prior means from the _Chatzilena et al_ definition.
+"
+
+# ╔═╡ 71a26408-1c26-46cf-bc72-c6ba528dfadd
+ar = AR(
```
seabbs, line 412: could be cool here to put in a prior predictive check of this sub model

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
```diff
@@ -0,0 +1,622 @@
...
+)
+
+# ╔═╡ 9309f7f8-0896-4686-8bfc-b9f82d91bc0f
+@model function stochastic_ode_mdl(Yt, ts, logobsprob, obs, prob, N;
+        solver = AutoTsit5(Rosenbrock23()),
+        upjitter = 1e-2#0.1,
+)
+
+    ##Priors##
+    β ~ LogNormal(0.0, 1.0)
```
seabbs, line 425: wow what if this was rewritten as an infection EpiAware infection model (new issue/PR)

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
```diff
@@ -0,0 +1,622 @@
...
+    ##Solve ODE model##
+    sol = solve(_prob, solver;
+        saveat = ts,
+        verbose = false
+    )
+
+    ##Sample the log-residual AR process##
+    nobs = length(Yt)
+    @submodel κₜ = generate_latent(logobsprob, nobs)
+    λt = @. N * sol[2, :] * exp(κₜ) + upjitter
```
seabbs, line 444: Can we be pro and use any of the helpers (transformLatent eetc etc) if no can we have an issue to make one that would let us do this?

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
SamuelBrand1: I would love that.

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
SamuelBrand1: Yes, really good shout

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
seabbs: I think the answer is to write the infection model as a EpiAware infection model and then do directly?

SamuelBrand1:
Hey, @damonbayer . Was just wondering if you have any thoughts on whether this is a good intro to SciML + EpiAware inference?

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
seabbs: This would currently be a bit painful so maybe needs to wait for a refactor of the infection submodels

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
seabbs: ha fair. One option we could take going forward is to put all new sentences on a new line. Makes it easier to see changes in git and maybe plays nicer with precommit. Downside is very boring.

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
SamuelBrand1: yes indeed!

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
```diff
@@ -0,0 +1,622 @@
...
+ar = AR(
+    damp_priors = [HalfNormal(mean(sampled_AR_damps))],
+    std_prior = HalfNormal(mean(sampled_AR_stds)),
+    init_priors = [Normal(0, 0.001)]
+)
+
+# ╔═╡ 9309f7f8-0896-4686-8bfc-b9f82d91bc0f
+@model function stochastic_ode_mdl(Yt, ts, logobsprob, obs, prob, N;
+        solver = AutoTsit5(Rosenbrock23()),
+        upjitter = 1e-2#0.1,
```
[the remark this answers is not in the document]
SamuelBrand1: Good catch

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
SamuelBrand1: That is a really good point:

1) https://github.com/CDCgov/Rt-without-renewal/issues/306 We have MAP/MLE opt as an issue
2) I could do this with the pathfinder functionality we already have

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
seabbs: Shall we make a new issue for this? I don't think this is to be addressed here? Then we can close this and linked issues.

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
seabbs: As this works as is lets leave and update in an issue. I like the idea that we support multiple options. I think it makes sense though to also link out to the idea we have lots of options when we update this.

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
SamuelBrand1: For a person with a taut readable style this would be no problem...

SamuelBrand1: Ok I've done this except for bullets

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
SamuelBrand1: Good point, I think its worth being consistent here.

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
SamuelBrand1: No not the negative safety (unless I missed something v possbile)

EpiAware/docs/src/showcase/replications/chatzilena-2019/index.jl
SamuelBrand1: I've got a bit mixed up... is this about not using `apply_method` I'm going to make a comment about this at the bottom.

SamuelBrand1:
I think I've addressed the comments from @seabbs . 

An issue to note, which will be an issue to address, is that in an experiment with using the `ManyPathfinder` -> `NUTS` helper functions (not committed) I found that it made inference both slower (approx x4 but I didn't fully benchmark) and one of the chains for the stochastic model diverged to a nonsense zone of parameter space. This is out of scope for this PR but needs followed up.

SamuelBrand1:
Pertinent to above and @seabbs comment on using `upjitter`. Using the softplus transform instead of an upjitter (or clamp to do relu) stabilises the sampling. Also, using `LogExpFunctions.xexpy` to implement the effect of the AR process on observations in the stochastic model.

After these changes, an experiment to not use initial values to start NUTS still converges onto the correct posterior distributions, so at least in this context this is a big improvement in the stability of the model. I know @damonbayer has wrangled with this in Bayesian inference of ODEs so I wonder what your thoughts on using special functions to stabilise sampling are? 

I've left the initial condition finding steps in the note because it still improves things a bit, and I think is a good example.

seabbs (approved):
This looks great now. Love it. Looking forward to v2 once we had added more integration to EpiAware.

PR #470, opened 2024-10-03
change exp to xexpy

### LINKED ISSUE #469: Change calculation method in `EpiAwareBase.generate_latent_infs(epi_model::ExpGrowthRate, rt)`

SamuelBrand1:
This is a small PR to replace the calculation of `ExpGrowthRate` infection generating process with `LogExpFunctions.xexpy` to get their overflow properties. 

The reason I targeted this specific bit of the code base it that everywhere else we can define `transformation` whereas here the `exp` transform is baked into the meaning of the igp.

Closes #469 .

NB: I also changed from `cumsum` to `accumulate(+, rt; init = init_incidence)` to avoid a broadcast add before applying the "safe" `xexpy`. I wonder if anything shows in benchmarks.

### DIFF
--- a/EpiAware/src/EpiInfModels/EpiInfModels.jl
+++ b/EpiAware/src/EpiInfModels/EpiInfModels.jl
@@ -6,7 +6,7 @@ module EpiInfModels
 using ..EpiAwareBase
 using ..EpiAwareUtils
 
-using Turing, Distributions, DocStringExtensions, LinearAlgebra
+using Turing, Distributions, DocStringExtensions, LinearAlgebra, LogExpFunctions
 
 #Export models
 export EpiData, DirectInfections, ExpGrowthRate, Renewal
--- a/EpiAware/src/EpiInfModels/ExpGrowthRate.jl
+++ b/EpiAware/src/EpiInfModels/ExpGrowthRate.jl
@@ -116,5 +116,5 @@ I_t = generated_quantities(latent_inf, θ)
 "
 @model function EpiAwareBase.generate_latent_infs(epi_model::ExpGrowthRate, rt)
     init_incidence ~ epi_model.initialisation_prior
-    return xexpy.(1.0, accumulate(+, rt; init = init_incidence))
+    return xexpy.(1.0, init_incidence .+ cumsum(rt))
 end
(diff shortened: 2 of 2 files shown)


seabbs (approved):
Nice. Agree we should hold off merging until we see what happens in the benchmarks

SamuelBrand1:
updated a hotfix for the doc fix

SamuelBrand1:
> ID | time ratio | memory ratio
-- | -- | --
["EpiInfModels", "ExpGrowthRate", "evaluation", "linked"] | 1.13 (5%) ❌ | 1.06 (1%) ❌
["EpiInfModels", "ExpGrowthRate", "evaluation", "standard"] | 1.13 (5%) ❌ | 1.06 (1%) ❌
["EpiInfModels", "ExpGrowthRate", "gradient", "ADTypes.AutoForwardDiff(chunksize=0)", "linked"] | 1.09 (5%) ❌ | 1.12 (1%) ❌
["EpiInfModels", "ExpGrowthRate", "gradient", "ADTypes.AutoForwardDiff(chunksize=0)", "standard"] | 1.10 (5%) ❌ | 1.12 (1%) ❌
["EpiInfModels", "ExpGrowthRate", "gradient", "ADTypes.AutoReverseDiff()", "linked"] | 0.43 (5%) ✅ | 0.79 (1%) ✅
["EpiInfModels", "ExpGrowthRate", "gradient", "ADTypes.AutoReverseDiff()", "standard"] | 0.43 (5%) ✅ | 0.79 (1%) ✅
["EpiInfModels", "ExpGrowthRate", "gradient", "ADTypes.AutoReverseDiff(compile=true)", "linked"] | 1.19 (5%) ❌ | 0.35 (1%) ✅
["EpiInfModels", "ExpGrowthRate", "gradient", "ADTypes.AutoReverseDiff(compile=true)", "standard"] | 1.18 (5%) ❌ | 0.35 (1%) ✅

Seems the relevant bit of the benchmarking? I think thats overall bad? But very big performance gain with not compiled tape AD?

seabbs:
Yes agree. I think ideally we would look at: 

1. xexpy in isolation
2. Doing cumsum(preprend etc in isolation)

SamuelBrand1:
> Yes agree. I think ideally we would look at:
> 
> 1. xexpy in isolation
> 2. Doing cumsum(preprend etc in isolation)

Good thinking and I've noted #472

seabbs:
So the `expy` change appears to have little impact on current benchmarks for forward diff. Making things significantly faster on reverse diff but hurt when using compiled reverse diff. 

So quite a mixed bag. I don't have a strong mental model to explain that result

SamuelBrand1:
> So the `expy` change appears to have little impact on current benchmarks for forward diff. Making things significantly faster on reverse diff but hurt when using compiled reverse diff.
> 
> So quite a mixed bag. I don't have a strong mental model to explain that result

Me neither.

PR #472, opened 2024-10-04
Replace cumsum broadcast with accumulate

seabbs:
Tests a single change from #470

### DIFF
--- a/EpiAware/src/EpiInfModels/ExpGrowthRate.jl
+++ b/EpiAware/src/EpiInfModels/ExpGrowthRate.jl
@@ -116,5 +116,5 @@ I_t = generated_quantities(latent_inf, θ)
 "
 @model function EpiAwareBase.generate_latent_infs(epi_model::ExpGrowthRate, rt)
     init_incidence ~ epi_model.initialisation_prior
-    return exp.(init_incidence .+ cumsum(rt))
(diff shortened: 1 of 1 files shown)


seabbs:
So this has a different pattern of impact vs #470. 

Here I see slower evaluation and forward diff, no real impact on reversediff with compilation and much faster reverse diff. 

Is this due to a specialisation that exists for `accumulate` do we think or other? 

I think this one is worth changing given these results (this makes me want to get Enzyme and Tapir into the benchmarks asap)

SamuelBrand1:
> Is this due to a specialisation that exists for `accumulate` do we think or other?

This is possible, but `cumsum` calls down to `accumulate!` which I'd imagine has similar things happening.

My mental model _was_ that `accumulate` saves either a broadcast sum or a vcat by using the `init` keyword arg... but I'm not sure thats the win I thought it would be now.

SamuelBrand1:
I've refreshed this test with the `xexpy` based function committed on main by #473

SamuelBrand1:
This is still ambiguous: some cases worse, some better.

As per f2f discussion with @seabbs I think that mean not going forward with this unless we get extra evidence. For example, if in a realistic inference exercise some kind of extra sampling stability improves sampling overall even if individual primal+grad calls are slower.

PR #476, opened 2024-10-06
Full priorpred check pipeline script

### LINKED ISSUE #468: Full prior predictive plotting step in pipeline

SamuelBrand1:
This PR closes #468 .

The contribution of this PR is to add `priorpredictive` mode as a pipeline option. This choice causes modifications down the pipeline so that:

- truthdata generation is turned off
- Sampler is set to direct prior sampling.
- The save object for a config does not include forecasting or scoring and the sampling result is sent to the plotting function created in #462 . The idea here is that prior plotting is more informative than serialising the samples, therefore we can only save the plots.

### Minor changes
I spotted and fixed a bug where saved objects would not get saved in the pipeline directories expected.


seabbs (approved):
Nice

PR #478, opened 2024-10-07
Changed saved prior predictive data

### LINKED ISSUE #477: Change saved prior predictive data

SamuelBrand1:
This PR changes the serialisation behaviour of the pipeline to reduce unnecessary objects being serialised (less important) and deals with problems in deserialising around reconstructing arbitrary Julia types (more important).

Closes #477 .

NB: While doing this PR, I was passing more pipeline data to `infer`... so it became easier to just pass the `pipeline` struct instance. This lets us select the inference post-processing based on the pipeline (e.g. testmode, priorpredictive etc). There might be a slicker way to do this which we can revisit in the future.


seabbs (approved):
Nice. I like the move to passing the pipeline type here vs the boolean.

PR #488, opened 2024-10-09
Update pipeline for julia 1.11

SamuelBrand1:
This is a small PR to turn off the jitter in the `EpiAwarePipeline` version of calculate processes. 

The reason is because in `julia-1.11` the basic functions e.g. `log` seem to be more accurate and we are failing tests at `atol=1e-6` because of a `jitter` of size `1e-6` that used to pass with `julia-1.10`.


seabbs (approved):
Nice. Note there is also a jitter in the observation model

seabbs (approved):

PR #491, opened 2024-10-10
Change transformation for Renewal model in test suite

SamuelBrand1:
This PR closes #490 

This PR makes the CI parameter recovery more robust by changing the `transformation` defined by `EpiData` from `exp` to `softplus`.

NB: Since the `exp` transform is very commonly used for $R_t$ for example in [EpiNow2](https://epiforecasts.io/EpiNow2/articles/estimate_infections.html#time-varying-reproduction-number) we might want to consider the problem in more detail despite this fix.

## Mathematical details

Softplus transformation and its inverse:

```math
\begin{aligned}
\text{softplus}(x) &= \log(1 + \exp(x)) \\
\text{invsoftplus}(y) &= \log(\exp(y) - 1)
\end{aligned}
```

In the parameter recovery test I have modified to do:

```math
\text{invsoftplus}(R_t) \sim Z_t
```
Where $Z_t$ is the latent process being tested (e.g. random walk, AR(1) and Diff AR(1)).

PR #494, opened 2024-10-10
issue 492: Create distribution that returns `<: Real`

SamuelBrand1:
This draft PR introduces `RealUnivariateDistribution` with an `eltype` method 

```julia
Base.eltype(::Type{<:Sampleable{F,RealValued}}) where {F} = Real
```

This means that `rand` is no longer hard encoded to expect either `Float64` or `Int` values.

Flagging for @seabbs. I want to see the CI and benchmarks before possible merge.

NB: This branch is _not_ using the change in CI from #491 ... which was accidental but actually might be a good test case.


SamuelBrand1:
Note that the benchmarks for observations are worse. 

This might suggest going for something a bit more specialised like `Union{Int,BigInt}`

SamuelBrand1:
Closing this as #497 looks much more promising.

Issue #471, opened 2024-10-03
Check, and if possible apply, `Transform...` struct using softplus etc for chatzilena et al replication.


SamuelBrand1:

SamuelBrand1:
The tricky thing about this is that one of the transformations _combines_ the infection time series with the latent process e.g.

```julia
λt = xexpy.(log1pexp.(N * sol[2, :]), κₜ)
```

SamuelBrand1:
Another thing is that we don't have a `Transform` for processes that aren't `AbstractLatentModel` subtypes. This is mainly because we are aiming to refactor infection processes as per #385 (I think).

This is an easy change though, which should make `Transform...` easy to include in any refactor.

seabbs:
So is this in some sense a ascertainment model?

SamuelBrand1:
Its more a question of _if_ we use this on things like the output of an ODE solve then do we want `Transform` to exist as a rectifier for that. The other option being that it lives "inside" the struct defining the ODE solve behaviour.

I think we do want a Transform struct modifier here because different choices have a big impact on inference (potentially) without changing anything else about the model.

seabbs:
without being very clever if we then combined this with a Deterministic `TransformObsModel` (i.e like https://github.com/CDCgov/Rt-without-renewal/blob/main/EpiAware/src/EpiLatentModels/modifiers/TransformLatentModel.jl but on expected cases vs on a latent model) the we could express this in our current framework

seabbs:
Also checking we do still have a hard coded pad: https://github.com/CDCgov/Rt-without-renewal/blob/0414f5f9da803950042e54ecf5032d67c43fd434/EpiAware/src/EpiObsModels/ObservationErrorModels/methods.jl#L21

SamuelBrand1:
> Also checking we do still have a hard coded pad:
> 
> https://github.com/CDCgov/Rt-without-renewal/blob/0414f5f9da803950042e54ecf5032d67c43fd434/EpiAware/src/EpiObsModels/ObservationErrorModels/methods.jl#L21

I think not.

seabbs:
thats literally a link to it?

SamuelBrand1:
[...]

PR #495, opened 2024-10-10
Issue 471: Changing ascertainment to transform x and y vs just y

seabbs:
As the title this PR partially addresses #471 by changing the structure of `Ascertainment` to take a transform function rather than a link. It also updates the default here to use `LogExpFunctions`. This is then propagated through the ascertainment day of week helper. I've slightly expanded the test suite and added a check in the `Ascertainment` struct that just checks for the input having two arguments - these could be expanded at a later date.


SamuelBrand1 (approved):
This looks really good to me!

PR #496, opened 2024-10-10
Issue 471: Add TransfromObservationModel

seabbs:
This PR adds `TransformObservationModel` in the same style as `TransformLatentModels`. It helps to address #471. 

I noted during this PR that the obs models names are getting very long and unwieldy. One obvious way to address this would be to use obs everywhere. I also think we should update `TransformLatentModels` to have a transform argument to match that used here.


EpiAware/src/EpiObsModels/modifiers/TransformObservationModel.jl
```diff
@@ -0,0 +1,56 @@
+@doc raw"
+The `TransformObservationModel` struct represents an observation model that applies a transformation function to the expected observations before passing them to the underlying observation model.
+
+## Fields
+- `model::M`: The underlying observation model.
+- `transform::F`: The transformation function applied to the expected observations.
+
+## Constructors
+- `TransformObservationModel(model::M, transform::F = x -> log1pexp.(x)) where {M <: AbstractTuringObservationModel, F <: Function}`: Constructs a `TransformObservationModel` instance with the specified observation model and a default transformation function.
```
SamuelBrand1, line 9: It turns out that the `softplus` alias for `log1pexp` exists so probably marginally better to write that for readability

SamuelBrand1: In doc strings I mean. I'm very agnostic on this in the src code

EpiAware/test/EpiObsModels/modifiers/TransformObservationModel.jl
```diff
@@ -0,0 +1,48 @@
...
+    @test trans_obs_kwarg.model == PoissonError()
+    @test trans_obs_kwarg.transform == custom_transform
+end
+
+@testitem "Test TransformObservationModel generate_observations" begin
+    using Turing, LogExpFunctions, Distributions
+
+    # Test with default log1pexp transform
+    trans_obs = TransformObservationModel(NegativeBinomialError())
+    gen_obs = generate_observations(trans_obs, missing, fill(10.0, 1))
```
SamuelBrand1, line 27: Should we have a test for a mix of missing and obs? I'd like to check that the transform returns `missing` as appropriate rather than fails?

(apols if that is covered but I missed it)

SamuelBrand1 (changes requested):
Looks good. Just a couple of comments

EpiAware/src/EpiObsModels/modifiers/TransformObservationModel.jl
seabbs: Do you think so? I wondere if log1pexp is better as its the explicit math and if our target users don't know what the softplus is that might be clearer? I don't feel very strongly either way

EpiAware/test/EpiObsModels/modifiers/TransformObservationModel.jl
seabbs: So as the transform is on the expected value surely this isn't something that comes from this function but rather the lower level submodel

seabbs: We can unit test it here you would like but I would expect that functionality to be independent of this functon

SamuelBrand1: > So as the transform is on the expected value surely this isn't something that comes from this function but rather the lower level submodel

Yes, ofc. Sorry was reviewing a bit late last night

EpiAware/src/EpiObsModels/modifiers/TransformObservationModel.jl
SamuelBrand1: I lean towards believing that expressions common in machine learning have a higher chance of being understood, but I don't really have strong evidence for that

seabbs: fair enough if that is our audience but its really epis and I think they probably won't have come across it. If you don't feel that strongly I think I would prefer to leave as is but what about adding something to the doc string to make it clear it is softplus?

seabbs: I added a very little signposting. Do you think that is enough?

SamuelBrand1: Yeah I think so.

SamuelBrand1 (dismissed):
LGTM

seabbs:
Hmm does this repeat failure indicate we should change the test?

seabbs:
@SamuelBrand1 thoughts on something to change here to make this safer for merging

SamuelBrand1:
> @SamuelBrand1 thoughts on something to change here to make this safer for merging

What was the fail?

seabbs:
its an inexact error for the benchmark for this feature which given its simplicity I find very confusing

SamuelBrand1:
> ERROR: LoadError: InexactError: Int64(-225.81712039285281)

Not sure what is causing this, but its quite a wild error in benchmarks

SamuelBrand1:
The error in the benchmarking is happening at the `varinfo` stage of the benchmark call. This suggests that its to do with the `linked` version of the benchmark (e.g. sampling in the unconstrained domain).

seabbs:
@SamuelBrand1 can you take a look at this and see if you can spot the problem? It would be great to be able to merge. this

EpiAware/src/EpiObsModels/modifiers/TransformObservationModel.jl
```diff
@@ -0,0 +1,56 @@
...
+end
+
+function TransformObservationModel(model::M;
+        transform::F = x -> log1pexp.(x)) where {
+        M <: AbstractTuringObservationModel, F <: Function}
+    return TransformObservationModel(model, transform)
+end
+
+@doc raw"
+Generates observations based on the `TransformObservationModel`.
```
SamuelBrand1, line 38:
```suggestion
Generates observations or accumulates log-likelihood based on the `TransformObservationModel`. 
```

SamuelBrand1 (approved):
This looks really handy. Noting the benchmark CI problem, which I'll also investigate.

SamuelBrand1:
To note: my attempt to "be helpful" collided with @seabbs simultaneous and functionally identical fix. I've left @seabbs fix up.

I've added a section in the contributor guide explaining the "gotcha" here with the benchmarking

SamuelBrand1 (dismissed):

SamuelBrand1 (approved):

PR #503, opened 2024-10-11
Tweak parameter recovery testing

seabbs:
This PR attempts to hotfix the parameter recovery unit tests by: 

1. Reducing the number of chains from 4 -> 2
2. Moving from a union data type to just missing

It also turns off progress in the sampling to make testing less noisy.

### DIFF
--- a/EpiAware/test/EpiObsModels/modifiers/LatentDelay.jl
+++ b/EpiAware/test/EpiObsModels/modifiers/LatentDelay.jl
@@ -222,10 +222,10 @@ end
 
         #Inference method
         inference_method = EpiMethod(
-            pre_sampler_steps = [ManyPathfinder(nruns = 4, maxiters = 100)],
+            pre_sampler_steps = [ManyPathfinder(nruns = 4, maxiters = 50)],
             sampler = NUTSampler(adtype = ad,
                 ndraws = 1000,
-                nchains = 4,
+                nchains = 2,
                 mcmc_parallel = MCMCThreads())
         )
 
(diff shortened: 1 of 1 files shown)


seabbs:
Any thoughts on this failing? It would be nice to get this in to speed up CI

SamuelBrand1:
I'm looking at this.

SamuelBrand1:
Alternative is to use a longer average and/or less relatively variable delay process: this decouples the obs noise from the process noise better.

seabbs:
Good investigation. Agree that this makes sense. We don't need these tests to be incredibly hard and right now we just want them to be stable and pass!

SamuelBrand1 (dismissed):
I think this is a good change. Lets see the CI.

EpiAware/test/EpiLatentModels/models/RandomWalk.jl
```diff
@@ -44,23 +44,23 @@ end
...
         return Z_t, y_t
     end
 
-    generative_mdl = test_negbin_errors(rw_process, obs_nb, fill(missing, 40))
+    generative_mdl = test_poisson_errors(process, obs, fill(missing, 40))
     θ_true = rand(generative_mdl)
     Z_t_obs, y_t_obs = condition(generative_mdl, θ_true)()
 
-    mdl = test_negbin_errors(rw_process, obs_nb, Int.(y_t_obs))
+    mdl = test_poisson_errors(rw_process, obs, Int.(y_t_obs))
```
SamuelBrand1, line 61:
```suggestion
    mdl = test_poisson_errors(process, obs, Int.(y_t_obs))
```

SamuelBrand1: Mini error I just spotted

SamuelBrand1 (approved):

seabbs (commented):
Agree. Great to see a speed up here.

Issue #492, opened 2024-10-10
Change SafePoisson and SafeNegativeBinomial to return value from `rand` calls


SamuelBrand1:
In light of discussion here https://github.com/JuliaStats/Distributions.jl/pull/1905#issuecomment-2389426188

We can expect the behaviour of `rand(d::Distributions)` to change in a way that will ameliorate edge-case numerical instability, e.g. `InexactError`s. The _origin_ of these issues as I understand it is the behaviour of `Distributions.eltype`, which can't always precisely predict the output type of a `rand` call from the type of the `Distribution`. This tracks the point made by @devmotion on the PR.

The simplest fix I can think of before (possible) implementation of linked PR is to either 
- make `Safe...` Continuous and treat integers as `x.0` 
- make `Safe...` `Real` valued.

seabbs:
Status of this?

PR #505, opened 2024-10-15
Safe distributions as float x.0 distributions

SamuelBrand1:
This is another test PR which aims at testing #492 by treating Safe integer dists as returning x.0 floats.


seabbs:
I think this was a test/dead so closing

PR #506, opened 2024-10-15
Add ascertainment model to chatzilena

SamuelBrand1:
This PR contributes to #471 by adding the `Ascertainment` functionality @seabbs added with #495 to the Chatzilena replication note, along with explanation.

I think that this improves the readability of the code, and anchors the code base in the description of the model as having a time-varying ascertainment factor.


seabbs (approved):
Nice. We can also use `TransformObservationModel` to get rid of line 523 I think

PR #510, opened 2024-10-17
Issue 465: Add an infection generating model for ODE problems

### LINKED ISSUE #465: New EpiAware infection generating process: generate from `SciML.solve`
wow what if this was rewritten as an infection EpiAware infection model (new issue/PR)

_Originally posted by @seabbs in https://github.com/CDCgov/Rt-without-renewal/pull/464#discussion_r1784631345_
            
The idea here is to make a new infection generating model which ingests some `DEProblem` and generates the infections from `solve` on this object.

SamuelBrand1:
This PR closes #465 .

## Contributions

This PR contributes the following:
- An `InfectionODEProcess` subtype of `AbstractTuringEpiModel` which defines the behaviour of a class of ODE based epi model.
- An `ODEParams` type for passing to the method of `generate_latent_infs` that dispatches the ODE solver.
- The `generate_latent_infs` method.

### ODE problem details

The ODE problem specialises to

1. The initial condition `u0` must be a vector or matrix.
2. The parameters passed to the ODE solver must be a vector or matrix.
3. The generated latent infections must be definable as a map on the `sol` object that is returned from an ODE `solve` call.

### Minor contributions and things to note

- Added a benchmark for `InfectionODEProcess` based generation.
- A version change changed the precise output of a docstring causing the doctest to fail, I caught this.
- I think `sciml` style might have changed because `JuliaFormatter` made a large number of small whitespace changes to files outside the scope.
- I have tested updating the _Chatzilena et al_ showcase with the new constructors and it works just as well with more elegant code. I've interpreted that as another issue but happy to commit to this PR if it adds.


EpiAware/src/EpiAwareUtils/prefix_submodel.jl
```diff
@@ -17,9 +17,11 @@ Generate a submodel with an optional prefix. A lightweight wrapper around the `@
 using EpiAware, DynamicPPL
 
 submodel = prefix_submodel(FixedIntercept(0.1), generate_latent, string(1), 2)
-submodel
```
seabbs, line 20: I wonder if we really gain anything from this doc test now and should instead make it an @example?

EpiAware/src/EpiInfModels/EpiInfModels.jl
```diff
@@ -6,10 +6,14 @@ module EpiInfModels
 using ..EpiAwareBase
 using ..EpiAwareUtils
 
-using Turing, Distributions, DocStringExtensions, LinearAlgebra, LogExpFunctions
+using Turing, Distributions, DocStringExtensions, LinearAlgebra, LogExpFunctions,
+      OrdinaryDiffEq
 
 #Export models
-export EpiData, DirectInfections, ExpGrowthRate, Renewal
+export EpiData, DirectInfections, ExpGrowthRate, Renewal, InfectionODEProcess
```
seabbs, line 13: very clunky have ODEParams but InfeectionODEProcess

I think ODEParams and ODEProcess seems reasonable. 

Small detail by the current export comments (models and parameters) aren't really correct. I think ODEParams and EpiData should maybe go together as parameter helpers and everything is a process model?

EpiAware/src/EpiInfModels/InfectionODEProcess.jl
```diff
@@ -0,0 +1,182 @@
...
+```jldoctest
+using EpiAware
+params = ODEParams(; u0 = ones(10), p = [2, 3])
+
+# output
+
+ODEParams{Float64}([1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0], [2.0, 3.0])
+```
+"""
+struct ODEParams{T}
```
seabbs, line 25: why not use @kwdef?

EpiAware/src/EpiInfModels/InfectionODEProcess.jl
```diff
@@ -0,0 +1,182 @@
...
+    F <: Function}
+    prob::P
+    ts::Vector{T}
+    solver::S
+    sol2infs::F
+end
+
+function InfectionODEProcess(prob::ODEProblem; ts,
+        solver = AutoTsit5(Rosenbrock23()),
+        sol2infs = sol -> sol[end, :])
```
seabbs, line 119: maybe call in transform to match other args elsewhere that do a similar thing?

EpiAware/src/EpiInfModels/InfectionODEProcess.jl
```diff
@@ -0,0 +1,182 @@
+"""
```
seabbs, line 1: What happened to @doc and @doc raw?

EpiAware/src/EpiInfModels/InfectionODEProcess.jl
```diff
@@ -0,0 +1,182 @@
...
+    u0::VecOrMat{T}
+    p::VecOrMat{T}
+end
+
+function ODEParams(; u0::VecOrMat, p::VecOrMat)
+    T = promote_type(eltype(u0), eltype(p))
+    return ODEParams{T}(T.(u0), T.(p))
+end
+
+"""
```
seabbs, line 35: @doc / @doc raw?

EpiAware/src/EpiInfModels/InfectionODEProcess.jl
```diff
@@ -0,0 +1,182 @@
...
+The purpose of this structure is to define the behaviour of modelling an infection process
+using an ODE. We use the [`SciML` ecosystem](https://sciml.ai/) to define and solve the ODE.
+For `InfectionODEProcess` structs we focus on defining from a restricted set of ODE problems:
+- The initial condition `u0` must be a vector or matrix.
+- The parameters `p` must be a vector or matrix.
+- The output of the ODE should be interpreted as the infection incidence at each time point in
+`ts` via the function `sol2infs` which maps the solution object `sol` of the ODE solve to infection counts.
+
+# Fields
+- `prob::P`: The ODE problem instance, where `P` is a subtype of `ODEProblem`.
```
seabbs, line 49: for clarity seems like we could afford problem unless this is matching from sciml?

EpiAware/src/EpiInfModels/InfectionODEProcess.jl
```diff
@@ -0,0 +1,182 @@
...
+- `ts::Vector{T}`: A vector of time points, where `T` is the type of the time points.
+- `solver::S`: The solver used for the ODE problem.
```
seabbs, line 51: if docing like this (I'd prefer in the struct docs with @doc raw need to say what the defaults are

EpiAware/src/EpiInfModels/InfectionODEProcess.jl
```diff
@@ -0,0 +1,182 @@
...
+- `sol2infs::F`: A function that maps the solution object of the ODE to infection counts.
+
+# Constructors
+- `InfectionODEProcess(prob::ODEProblem; ts, solver, sol2infs)`: Create an `InfectionODEProcess`
+object with the ODE problem `prob`, time points `ts`, solver `solver`, and function `sol2infs`.
+
+# Type Parameters
```
seabbs, line 58: We don't have this doc anywhere else (LLM spam?) do we need it?

EpiAware/src/EpiInfModels/InfectionODEProcess.jl
```diff
@@ -0,0 +1,182 @@
...
+using EpiAware, OrdinaryDiffEq
+r = log(2) / 7 # Growth rate corresponding to 7 day doubling time
+u0 = [1.0]
+p = [r]
+params = ODEParams(u0 = u0, p = p)
+
+# Define the ODE problem using SciML
+# We use a simple exponential growth model
+
+function expgrowth(du, u, p, t)
```
seabbs, line 76: nice! For a later PR we should ship some helpers for common models I think

seabbs: (also I wonder how this benchmarks vs the discrete version...)

EpiAware/src/EpiInfModels/InfectionODEProcess.jl
```diff
@@ -0,0 +1,182 @@
...
+expgrowth_model = InfectionODEProcess(prob::ODEProblem; ts = 0:1:10,
+        solver = Tsit5(),
+        sol2infs = sol -> sol[1, :])
+
+# Generate the latent infections
+I_t = generate_latent_infs(expgrowth_model, params)()
+```
+"""
+@model function EpiAwareBase.generate_latent_infs(
+        epi_model::InfectionODEProcess, params::ODEParams)
```
seabbs, line 171: As written this doesn't look like it will work within the generate_epiaware framework (https://github.com/CDCgov/Rt-without-renewal/blob/c5f47ff512df19f919a477e2af0cd2d03413602e/EpiAware/src/EpiAwareUtils/turing-methods.jl#L14)

We have agreed elsewhere to update the infection processes so that they don't depend on a latent model (once they are composable). 

One option could be to encode the initial conditions and parameters as submodel dispatched via a field in the ODEProcess struct (i.e ODEParams is a field of ODEProcess) and have a deterministic default (i.e the current struct).

seabbs (changes requested):
This looks really nice. Mostly renaming along with some questions where I think the SciML approach is clashing with our current approach elsewhere. 

My particular concern is the pattern you envisage for integrating this with a current (or future generate_epiaware call).

EpiAware/src/EpiInfModels/InfectionODEProcess.jl
SamuelBrand1: I thought about this, but since I have outer constructors with `kwargs` it seemed pretty unnecessary

EpiAware/src/EpiInfModels/InfectionODEProcess.jl
SamuelBrand1: I thought about this, but we have a few things going on:

- The concept of reduction (rather than transformation), e.g. going from an `ODESolution` struct to an `Array`.
- The concept of transforming that `Array` for the purpose of easier AD/ catching non-negs without needing to use `DiffEqCallbacks` etc
- The concept of transforming as a part of a model (as separate from convenience), e.g. our `Transform...` model structs.

So it felt that using `transform` willy-nilly might be even more confusing?

EpiAware/src/EpiInfModels/InfectionODEProcess.jl
SamuelBrand1: Yes, this is a match for `ODEProblem` from `SciML` which differentiates from using the (bigger) `DEProblem`. I'm using the lower dependency `OrdinaryDiffEq` so we aren't loading methods for wider classes of problem e.g. SDEs, boundary value problems, Delay diff eqs etc.

EpiAware/src/EpiInfModels/InfectionODEProcess.jl
SamuelBrand1: No we don't need it. This is indeed uncaught LLM spam

EpiAware/src/EpiInfModels/InfectionODEProcess.jl
SamuelBrand1: We aren't docing like this.

EpiAware/src/EpiInfModels/InfectionODEProcess.jl
SamuelBrand1: In fact, just to add clarity. The outer constructor for `ODEParams` also does a `promote_type` step (which is useful for keeping the parametric struct type simple), but it would conflict with the constructor generated by `@kwdef` here.

EpiAware/src/EpiInfModels/InfectionODEProcess.jl
seabbs: fine can use something that doesn't include infs. (reduction, selection

EpiAware/src/EpiInfModels/InfectionODEProcess.jl
```diff
@@ -0,0 +1,182 @@
...
+    ts::Vector{T}
+    solver::S
+    sol2infs::F
+end
+
+function InfectionODEProcess(prob::ODEProblem; ts,
+        solver = AutoTsit5(Rosenbrock23()),
+        sol2infs = sol -> sol[end, :])
+    P = typeof(prob)
+    T = eltype(ts)
```
seabbs: I'm pretty unclear why you need to do this?

EpiAware/src/EpiInfModels/InfectionODEProcess.jl
```diff
@@ -0,0 +1,182 @@
...
+ 1.4859941865014936
+ 1.640671113705054
+ 1.8114471151863056
+ 1.9999990356297939
+ 2.2081789476865237
+ 2.438027196361022
+ 2.6918002758361723
+```
+"""
+struct InfectionODEProcess{P, T, S, F} <: EpiAwareBase.AbstractTuringEpiModel where {
```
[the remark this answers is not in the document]
seabbs: If we follow this pattern everywhere this is going to add quite a lot more boiler plate and increase the distance between docs of args and there defaults which I am not a fan of. I would prefer we stick with @kwdef until we at least test and find its a problem to do so

seabbs: why are you not declaring any of the types here as we do in all other structs?

seabbs: I think we can discuss a style change elsewhere but it shouldn't happen piecemeal

SamuelBrand1: Good spot! This was unplanned, I literally misremembered our style here.

SamuelBrand1: Fixed this now.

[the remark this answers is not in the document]
SamuelBrand1: Rethink: this is a bit different to wanting to use `promote_type` so I'll changed this to our usual pattern.

seabbs (commented):
This is starting to get there

EpiAware/src/EpiAwareUtils/EpiAwareUtils.jl
```diff
@@ -19,7 +19,8 @@ export HalfNormal, DirectSample, SafePoisson, SafeNegativeBinomial, SafeIntValue
        SafeDiscreteUnivariateDistribution
 
 #Export functions
-export scan, spread_draws, censored_cdf, censored_pmf, get_param_array, prefix_submodel, ∫F
+export scan, spread_draws, censored_cdf, censored_pmf, get_param_array, prefix_submodel, ∫F,
+       _expand_dist
```
seabbs, line 23: as this is exported it needs to be renamed away from the private naming scheme

EpiAware/src/EpiInfModels/ODEProcess.jl
```diff
@@ -0,0 +1,156 @@
...
+params = ODEParams(; u0 = ones(10), p = [2, 3])
+
+# output
+
+ODEParams{Float64}([1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0], [2.0, 3.0])
+```
+"""
+struct ODEParams{T}
+    "The initial condition(s) for the ODE, which can be a vector or matrix of type `T`."
+    u0::VecOrMat{T}
```
seabbs, line 20: Can you clarify why `VecOrMat isn't in the {} definition?

seabbs: To me this suggests a much more flexible typing approach than we have used elsewhere

EpiAware/src/EpiInfModels/ODEProcess.jl
```diff
@@ -0,0 +1,156 @@
...
+ 2.438027196361022
+ 2.6918002758361723
+```
+"""
+@kwdef struct ODEProcess{P <: ODEProblem, T, S, F <: Function} <:
+              EpiAwareBase.AbstractTuringEpiModel
+    "The ODE problem instance, where `P` is a subtype of `ODEProblem`."
+    prob::P
+    "A vector of time points, where `T` is the type of the time points."
+    ts::Vector{T}
```
seabbs, line 94: Sam question why is T not Vector{<: Real} or similar?

seabbs:
🙏

EpiAware/src/EpiInfModels/ODEProcess.jl
SamuelBrand1: Good shout

EpiAware/src/EpiInfModels/ODEProcess.jl
```diff
@@ -0,0 +1,170 @@
...
+p = [r]
+params = ODEParams(u0 = u0, p = p)
+
+# Define the ODE problem using SciML
+# We use a simple exponential growth model
+
+function expgrowth(du, u, p, t)
+    du[1] = p[1] * u[1]
+end
+prob = ODEProblem(expgrowth, u0, (0.0, 10.0), p)
```
seabbs, line 147: is it more efficient to do it this way (make a problem with fixed params and then remake) vs make it for the first time inside the call?

EpiAware/src/EpiInfModels/ODEProcess.jl
```diff
@@ -0,0 +1,170 @@
...
+```julia
+using EpiAware, OrdinaryDiffEq, Distributions
+# Define the time span for the ODE problem
+tspan = (0.0, 30.0)
+
+# Define prior distributions
+infectiousness_prior = LogNormal(log(0.3), 0.05)
+incubation_rate_prior = LogNormal(log(0.1), 0.05)
+recovery_rate_prior = LogNormal(log(0.1), 0.05)
+initial_prop_infected_prior = Beta(1, 99)
```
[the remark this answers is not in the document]
SamuelBrand1: I've thought about it, but it can get pretty annoying. 

For example, the appropriate initial conditions are often functions of parameter samples, e.g. if you sample incubation and infectious periods then the initial distribution among the population are heavily correlated to this (this can cause really annoying sampling problems). Or it can happen vice versa, e.g. sampling an inital growth rate can be an initial condition in some forms of epi model which is correlated to parameters.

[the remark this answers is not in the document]
seabbs: Okay fair makes sense.

EpiAware/src/EpiInfModels/ODEProcess.jl
```diff
@@ -0,0 +1,170 @@
...
+Where `S` is the proportion of the population that is susceptible, `I` is the proportion of the
+population that is infected and `R` is the proportion of the population that is recovered. The
+parameters are the infectiousness `β` and the recovery rate `γ`.
+
+```julia
+using EpiAware, OrdinaryDiffEq, Distributions
+# Define the time span for the ODE problem
+tspan = (0.0, 30.0)
+# Define prior distributions
+infectiousness_prior = LogNormal(log(0.3), 0.05)
```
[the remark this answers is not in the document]
SamuelBrand1: Done this, and agree looks better!

SamuelBrand1:
Planned change: after f2f conversation with @seabbs it'll be smoother, both now and possible future changes, to have the parameter generation as a `generate_latent` method rather than creating a new type system for parameters.

SamuelBrand1:
> Planned change: after f2f conversation with @seabbs it'll be smoother, both now and possible future changes, to have the parameter generation as a `generate_latent` method rather than creating a new type system for parameters.

I've implemented this change:

- No more `AbstractParam` type.
- `ODEParams` are not a `LatentModel` subtype with `generate_latent` methods.
- Reorganised code/tests to reflect this.

seabbs:
> I think I've nearly fully addressed this concern. Where we might be missing something is not having a NullLatentModel type which just generates nothing values. I wanted to leave the option open for the ODE models to be conditional on a latent process outcome, e.g. as per a Renewal model, but that is not currently used. A nothing returning process would let us use the current infra for model specification.

Good idea

EpiAware/src/EpiAwareUtils/EpiAwareUtils.jl
```diff
@@ -37,5 +37,6 @@ include("get_param_array.jl")
 include("SafeInt.jl")
 include("SafePoisson.jl")
 include("SafeNegativeBinomial.jl")
+include("_expand_dist.jl")
```
seabbs, line 40: if we are using this across modules I think we need to export it or at least update the name to a non-internal pattern

EpiAware/src/EpiInfModels/ODEProcess.jl
```diff
@@ -0,0 +1,253 @@
...
+```julia
+# Sampled parameters
+θ = rand(sir_infections_mdl)
+expected_infections = (sir_infections_mdl | θ)()
+test_data = expected_infections .|> λ -> rand(Poisson(λ))
+```
+
+Now, we can refit the model using a `PoissonError` observation model
+
+```julia
```
seabbs, line 124: I think we need to move away from using julia and only use @example

EpiAware/src/EpiInfModels/ODEProcess.jl
```diff
@@ -0,0 +1,253 @@
...
+We can generate some test data from the model by sampling from the model and then calling the model
+
+```julia
+# Sampled parameters
+θ = rand(sir_infections_mdl)
+expected_infections = (sir_infections_mdl | θ)()
+test_data = expected_infections .|> λ -> rand(Poisson(λ))
+```
+
+Now, we can refit the model using a `PoissonError` observation model
```
seabbs, line 122: it feels a bit weird to me that this isn't using a EpiAware wrapper. Do you think it adds value?

EpiAware/src/EpiLatentModels/odemodels/SEIRParams.jl
```diff
@@ -0,0 +1,175 @@
...
+# Returns
+  A tuple `(u0, p)` where:
+  - `u0`: A vector representing the initial state of the system `[S₀, E₀, I₀, R₀]` where `S₀`
+  is the initial proportion of susceptible individuals, `E₀` is the initial proportion of exposed
+  individuals,`I₀` is the initial proportion of infected individuals, and `R₀` is the initial
+  proportion of recovered individuals.
+  - `p`: A vector containing the parameters `[β, α, γ]` where `β` is the infectiousness rate,
+  `α` is the incubation rate, and `γ` is the recovery rate.
+"""
+@model function EpiAwareBase.generate_latent(params::SEIRParams, Z_t)
```
seabbs, line 166: why have you written this to take in Z_t and not n like the other latent models?

EpiAware/src/EpiInfModels/ODEProcess.jl
```diff
@@ -0,0 +1,253 @@
...
+
+generated_It = generate_latent_infs(sir_process, nothing)()
+```
+
+"""
+@model function EpiAwareBase.generate_latent_infs(epi_model::ODEProcess, Z_t)
+    prob, solver, sol2infs, solver_options = epi_model.params.prob,
+    epi_model.solver, epi_model.sol2infs, epi_model.solver_options
+
+    @submodel prefix="params" u0, p=generate_latent(epi_model.params, Z_t)
```
seabbs, line 246: why apply a prefix here?

EpiAware/src/EpiInfModels/ODEProcess.jl
```diff
@@ -0,0 +1,253 @@
...
+```
+
+We can compare the summarized chain to the sampled parameters to see that the model is fitting
+the data well.
+
+# Custom ODE models
+
+To define a custom ODE model, you need to define:
+
+- Some `CustomParams <: AbstractTuringParamModel` struct (the name doesn't need to be `CustomParams`!)
```
seabbs, line 146: Is this out of date? If so highlights danger in using julia vs example

EpiAware/src/EpiLatentModels/odemodels/SEIRParams.jl
seabbs, line 166: The implicit promise of this being a latent model is that it will work with the library of latent model functionality. Does this implementation do so? If no do we need to rethink the idea of making this a latent model

seabbs (commented):
Nice. Close. One more round of tightening and some push back about the use of the latent model.

EpiAware/src/EpiLatentModels/odemodels/SEIRParams.jl
SamuelBrand1: This is a mistake, I'll fix it

EpiAware/src/EpiInfModels/ODEProcess.jl
SamuelBrand1: Left over from the `generate_parameter` approach

EpiAware/src/EpiAwareUtils/EpiAwareUtils.jl
SamuelBrand1: The move to `EpiAwareUtils` was motivated by an idea which I didn't go forwards with so I've reverted.

EpiAware/src/EpiInfModels/ODEProcess.jl
SamuelBrand1: And gone now.

EpiAware/src/EpiInfModels/ODEProcess.jl
SamuelBrand1: I think I should refactor this to generate data from the model; then refit. That would follow our pattern much better

EpiAware/src/EpiInfModels/ODEProcess.jl
SamuelBrand1: This is a really good point (and see below where this would have caught an outdated approach.

However, I think that `@example` doesn't mean execution inside docstrings.

Options:

1. Replace with `jldoctest` and end a block with something like `nothing` and just checking nothing. This make the doctest a test against the code not failing.
2. Have actual targets for doctest
3. Reduce size of docstrings and put complicated code into a (pluto or documenter-flavoured md) tutorial.

EpiAware/src/EpiInfModels/ODEProcess.jl
SamuelBrand1: I'll hold for a sec and we can decide how we want to validate the docstrings

EpiAware/src/EpiInfModels/ODEProcess.jl
SamuelBrand1: I'll offer option 1. to concretise this a bit; but happy to vary.

EpiAware/src/EpiInfModels/ODEProcess.jl
SamuelBrand1: Since we've gone with option 1 above, I've refactored the example to use our EpiAware wrappers... Its just better

EpiAware/src/EpiInfModels/ODEProcess.jl
SamuelBrand1: Implemented option 1

EpiAware/src/EpiInfModels/ODEProcess.jl
SamuelBrand1: Redone this, and its covered by a `jldoctest` to flag if we move away from this custom pattern

EpiAware/src/EpiLatentModels/odemodels/SEIRParams.jl
SamuelBrand1: I'm not sure off the top of my head... I think this an example of something we should think about generally #525

EpiAware/src/EpiInfModels/ODEProcess.jl
SamuelBrand1: Not resolved this conversation because interested in @seabbs view here.

seabbs (approved):
I think this looks good for now but I find the use of latent model here a bit jarring and indicative of a wider design flaw. I see the issue to discuss so let's do that there. 

Less fussily though this is great functionality to have

PR #512, opened 2024-10-18
Fix benchmark ci

SamuelBrand1:
This PR hotfixes `BenchmarkCI.jl` incompat by creating a local version with appropriate compatibility.

This should only be seen as temporary until https://github.com/tkf/BenchmarkCI.jl/pull/181 is resolved


SamuelBrand1:
Making `BenchmarkCI` a local resources seems to make it hard to run `PkgBenchmark`. Not really sure whats causing the problem with reading the generated `Manifest`

seabbs (approved):

seabbs:
This looks like its still having some problems?

seabbs:
@SamuelBrand1 problems here?

SamuelBrand1:
Yes, I'm really not understanding the error here. Because `PkgBenchmark` _is_ registered.

seabbs:
I think we want to rebase this off main vs trying to stack it in the ode process pr

SamuelBrand1:
> I think we want to rebase this off main vs trying to stack it in the ode process pr

Shall I close this, and we can revisit?

PR #515, opened 2024-10-24
Hotfix: remove prefix_submodel doctest

SamuelBrand1:
It turns out the precise struct description (i.e. non-pretty print `show` of the struct) has changed.

Since this is quite complicated and brittle, the hotfix is to just swap this to being a standard `julia` block in the docstring. Note that this docstring only really covered that call the prefix function didn't error which is also covered in our unit tests.

### DIFF
--- a/EpiAware/src/EpiAwareUtils/prefix_submodel.jl
+++ b/EpiAware/src/EpiAwareUtils/prefix_submodel.jl
@@ -13,13 +13,9 @@ Generate a submodel with an optional prefix. A lightweight wrapper around the `@
 
 # Examples
 
-```jldoctest
+```julia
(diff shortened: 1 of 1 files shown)


EpiAware/src/EpiAwareUtils/prefix_submodel.jl
```diff
@@ -13,13 +13,9 @@ Generate a submodel with an optional prefix. A lightweight wrapper around the `@
 
 # Examples
 
-```jldoctest
+```julia
```
seabbs, line 16:
```suggestion
```@example
```

seabbs (commented):
I think we want to be porting to examples

seabbs (approved):

Issue #511, opened 2024-10-18
BenchmarkCI is incompatible with DiffEqBase


SamuelBrand1:
I've noted that the `compat` requirements for `BenchmarkCI` are now misaligned with `DiffEqBase` with regards to (at least) using `SetField`.

I've got a PR towards `BenchmarkCI` that could fix this here https://github.com/tkf/BenchmarkCI.jl/pull/181 , but at the moment this blocks benchmarks running for #510

SamuelBrand1:
Flagging this for @seabbs after f2f conversation.

SamuelBrand1:
Closed by #524

PR #520, opened 2024-11-11
Patch: Switch to fork of benchmarkCI

SamuelBrand1:
The goal of this is to switch `main` to using the fork of benchmarkCI and contribute to closing #511 .

### DIFF
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -22,8 +22,6 @@ jobs:
         run: |
           julia --project=benchmark -e 'using Pkg;
           Pkg.rm("EpiAware");
-          Pkg.rm("BenchmarkCI");
-          Pkg.add(url="https://github.com/SamuelBrand1/BenchmarkCI.jl");
           Pkg.resolve();
           Pkg.instantiate();
           Pkg.develop(path = "./EpiAware")'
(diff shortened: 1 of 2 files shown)


.github/workflows/benchmark.yaml
```diff
@@ -22,6 +22,8 @@ jobs:
         run: |
           julia --project=benchmark -e 'using Pkg;
           Pkg.rm("EpiAware");
+          Pkg.rm("BenchmarkCI");
+          Pkg.add(url="https://github.com/SamuelBrand1/BenchmarkCI.jl");
```
seabbs, line 26: Why not update the project toml?

SamuelBrand1: I'm following the pattern in the BenchmarkCI README.

https://github.com/tkf/BenchmarkCI.jl

seabbs:
As in add a source here: https://github.com/CDCgov/Rt-without-renewal/blob/main/benchmark/Project.toml

So that this uses the correct package both in CI and when called locally

.github/workflows/benchmark.yaml
seabbs: that doesn't have a rm?

seabbs: and we didnn't follwo trhat patter up to now as we hade the project toml?

SamuelBrand1: Its now failing in ways I really don't expect. Putting in the `[sources]` section means that CI failing on `resolve` with an error that suggests that it can't register PkgBenchmark, which is really strange

seabbs (approved):
Wow a surprising pain! Nice fix

PR #524, opened 2024-11-14
Patch fix: add `Manifest.toml` to benchmarking

SamuelBrand1:
After some problems with the benchmarking CI, I came to the conclusion that the failure was due to swapping to `Pkg.add` in CI which was causing uncommitted changes to the branch in the CI run leading to a failure ("dirty commit").

To stabilise the changes on the checked out branch in CI I've add a `Manifest.toml` file for the benchmarking following https://github.com/tkf/BenchmarkCI.jl#setup-with-benchmarkmanifesttoml .

- The fork of `BenchmarkCI` we are using is specified there.
- `EpiAware` dep is specified there.

This reduces the CI command to just `instantiate` and stops an error occurring between running benchmarks for the target and the baseline commits.

Evidence that this works is the test draft PR #523 which is aimed at benchmarking the test branch `trigger-bci-2` against the test branch `trigger-bci`: both of which have this approach to specifying the `benchmark` project deps.

### DIFF
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -24,7 +24,7 @@ jobs:
           Pkg.instantiate();
           Pkg.update()'
       - name: Run benchmarks
-        run: julia --project=benchmark -e 'using BenchmarkCI; BenchmarkCI.judge(; baseline = "origin/main", retune = false)'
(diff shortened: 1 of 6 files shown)


seabbs (approved):
What a nightmare and this all comes back to it not respecting source? It’s weird they show the add pattern working in their docs

Issue #525, opened 2024-11-15
Guaranteed latent model functionality


SamuelBrand1:
The implicit promise of this being a latent model is that it will work with the library of latent model functionality. Does this implementation do so? If no do we need to rethink the idea of making this a latent model

_Originally posted by @seabbs in https://github.com/CDCgov/Rt-without-renewal/pull/510#discussion_r1842263147_
            
As @seabbs mentions above the latent model type (setting ODE model and parameter priors) offered by #510 might not make sense with the whole latent model stack. E.g. What does broadcasting a sample from this kind of latent model mean?

This opens up a conversation about what we want for the fundamental capabilities of latent model.

SamuelBrand1:
Agreed: latent random variable (sets) we expect to be indexable should be one branch and others on the other.

PR #530, opened 2024-11-20
Issue 529: Create null Latent model

### LINKED ISSUE #529: Create NullLatentModel
> I think I've nearly fully addressed this concern. Where we might be missing something is not having a NullLatentModel type which just generates nothing values. I wanted to leave the option open for the ODE models to be conditional on a latent process outcome, e.g. as per a Renewal model, but that is not currently used. A nothing returning process would let us use the current infra for model specification.

Good idea

_Originally posted by @seabbs in https://github.com/CDCgov/Rt-without-renewal/issues/510#issuecomment-2468572438_
            
As discussed here, this would be an easy and handy feature.

SamuelBrand1:
This small PR closes #529 .

This PR contributes a `Null` struct which defines the concept of having a `nothing` latent process. This is useful to reusing patterns that we already have (e.g. `EpiProblem`) in situations where we don't actually want a latent process.

_Potentially_ it might be better to sequence this to after resolving #525 , because like the ODE parameter objects this not a naturally _indexable_ latent model.


seabbs:
I think this is fine to go in for now. Noting that we only need this whilst we take `Z_t` as an input and once that is shifted to be model internal we should no longer need this and can remove?

seabbs (dismissed):

SamuelBrand1:
> I think this is fine to go in for now. Noting that we only need this whilst we take `Z_t` as an input and once that is shifted to be model internal we should no longer need this and can remove?

Yes, hard to imagine we'll need this with a suitably flexible interface.

PR #534, opened 2024-12-05
Issue 532: Change deps and remove Dagger for pipeline

### LINKED ISSUE #532: Reduce pipeline deps and simplify pipeline compute
From some preliminary experiments using cloud tools, I've noted two issues:

- Using `RCall.jl` is rather more painful than locally since R installation has a set up complexity.
- Using `Dagger.jl` doesn't seem to add much to the Rt-without project because the implied task DAG is very simple. However, it does seem to be brittle to different compute environments. I've found hard to debug "missing socket" errors. I'll try and recreate a MWE for an issue on their repo, but a quick experiment with `pmap` works very successfully.

SamuelBrand1:
This PR closes #532 .

The way in which I've addressed problems described in #532 around using `RCall` and `Dagger` in the pipeline execution is to:

1. Move the usage of `RCall` into an extension of `EpiAwarePipeline`. This means we can still call to `scoringutils` in post-processing but it doesn't need to installed for a pipeline main inference compute.
2. Removed `Dagger` and replace with `pmap`. This is a bit disappointing but it works fine for this problem.


seabbs (approved):
Nice good solution. A shame about Dagger

SamuelBrand1:
> Nice good solution. A shame about Dagger

Agreed.

Issue #405, opened 2024-07-25
Model specific priors capability


SamuelBrand1:
We have been concentrating on the infrastructure of the pipeline, but I think the priors need some reasoning upon because for different scenarios they end up having different interpretations.

For example, the typical variation over time for the exponential growth rate $r_t$ is going to be on a different scale to $\log(I_t)$ the log-incidence. Our prior beliefs should reflect this.

To be able to implement this we probably to vary away from a default constructor for priors and towards a constructor that dispatches on the IGP being selected in the list of possible model combinations.

EDIT: Adding a smaller task list to get here

- [x] #455
- [x] #468
- [x] #477
- [x] #481
- [ ] #522

SamuelBrand1:
A particular problem here is that we also have a (current default) 35 days look back for inference... this means that at least some of the priors should depend on state at beginning of lookback window (Or we revert to just fitting on whole time series).

SamuelBrand1:
This is part of resolving #4 .

PR #535, opened 2024-12-09
Issue 537: Fix making analysis dataframes and record priors

SamuelBrand1:
~This draft PR is for tracking changes to the `EpiAwarePipeline` code base done in order to run the full pipeline and to fix unanticipated changes in the analysis code.~

I've decided to upgrade the work so far to a full PR. This PR closes #537 and contributes to #405.

## Tasks so far

- [x] Record the tightened priors chosen to partially address #405; but still general to all models. 
- [x] Fix `make_prediction_dataframe_from_output` function to implement changes to saved objects due to #445. This closes #537


SamuelBrand1:
This PR modifies the `make_prediction_dataframe_from_output` function to: 

1. Fix the function interface to the output data structure by responding to pipeline changes.
2. Use new `inference_config` rather than logic on the filename of serialised object.

This PR also adds a small "test data" in the correct format for use in unit testing. From f2f discussion with @seabbs this is a temporary solution until we make a test which generates its own test data.

SamuelBrand1:
I was thinking about breaking this up, but I didn't like the sequence and ended up with a with set of commits.

seabbs (approved):

PR #540, opened 2024-12-10
Test: Add enzyme and mooncake to benchmark

seabbs:
This PR adds Enzyme.jl and Mooncake.jl to out benchmarking to see what happens. Potentially these should be split into two PRs and then kept updated as new versions of the autodiff backends are produced.

### DIFF
--- a/benchmark/Manifest.toml
+++ b/benchmark/Manifest.toml
@@ -1,13 +1,13 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.11.1"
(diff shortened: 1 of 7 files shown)


seabbs:
Potentially there is a setup issue as from what I can see both fail for all benchmarks

seabbs:
I have also inserted Enzyme and Mooncake into one of the case studies each as another small test

seabbs:
Ha nothing is loaded - I am clearly doing a great job

PR #548, opened 2024-12-12
Issue 541: Changelog for prior changes and fix the data ingestion step

### LINKED ISSUE #541: Record changes in priors and fix post-process
This is a sub-task in #432 . This records the next step use of priors and the fixed/refactored the script to ingest the raw `jld2` output.

SamuelBrand1:
As per title, this PR closes #541 .

Extra addition is that the data ingestion script also creates a table of failed inferences 54 / 1440 failed; this follows the table approach in the supporting information quarto nb and should be added to the SI nb (new issue though IMO).


seabbs (approved):
Nice. LGTM

PR #552, opened 2024-12-13
Issue #509: Remove remaining inner constructors that don't use new

### LINKED ISSUE #509: Outer constructor methods
I've seen some suggestions in the julia ecosystem that _outer_ constructor methods work better with AD. I'll outsource my findings to perplexity (search: `Are outer constructors better for AD in julia`)

## Summary

Outer constructors can indeed be advantageous for automatic differentiation (AD) in Julia, particularly when working with complex types or when you need to perform additional computations during object construction. Here's an explanation of why outer constructors are often preferred for AD:

## Advantages of Outer Constructors for AD

**Flexibility and Control**

Outer constructors provide more flexibility in how objects are created, which can be beneficial for AD[1]. They allow you to:

- Perform pre-processing or validation of input arguments
- Apply transformations to the inputs before creating the object
- Implement custom logic that may be necessary for proper AD

**Avoiding AD of Constructor Internals**

By using outer constructors, you can avoid automatic differentiation of the internal constructor logic, which may not be necessary or desirable for your AD computations[2]. This can lead to:

- Improved performance by reducing unnecessary AD computations
- Cleaner and more focused gradients that only consider the relevant parts of your code

**Compatibility with AD Tools**

Some AD tools in Julia work better with outer constructors. For example, Zygote.jl, a popular AD library, has better support for outer constructors compared to inner constructors[3].

## Implementation Considerations

When using outer constructors for AD-friendly code, consider the following:

1. **Keep constructors pure**: Ensure that your outer constructors are pure functions without side effects, as this is crucial for accurate AD.

2. **Use type parameters judiciously**: Leverage Julia's type system to create efficient and AD-friendly constructors:

```julia
struct MyType{T}
    x::T
    y::T
end

# Outer constructor
function MyType(x::Real, y::Real)
    T = promote_type(typeof(x), typeof(y))
    return MyType{T}(T(x), T(y))
end
```

3. **Avoid non-differentiable operations**: In your outer constructors, try to avoid operations that are not differentiable or may cause issues with AD, such as branching conditions based on input values.

4. **Provide custom adjoints**: For complex constructors, you may need to provide custom adjoint rules to ensure correct gradient computation:

```julia
using Zygote

function Zygote.pullback(::Type{MyType}, x::Real, y::Real)
    T = promote_type(typeof(x), typeof(y))
    my_type = MyType{T}(T(x), T(y))
    function my_type_pullback(Δ)
        return (nothing, Δ.x, Δ.y)
    end
    return my_type, my_type_pullback
end
```

## Conclusion

While inner constructors have their place in Julia programming, outer constructors often provide a more flexible and AD-friendly approach to object creation. By using outer constructors, you can maintain better control over the construction process, optimize for AD performance, and ensure compatibility with popular AD tools in the Julia ecosystem.

Citations:
[1] https://docs.julialang.org/en/v1/manual/constructors/
[2] https://gensoft.pasteur.fr/docs/julia/1.4.1/manual/constructors.html
[3] https://github.com/JuliaLang/julia/issues/8135
[4] https://docs.julialang.org/en/v1/manual/types/
[5] https://juliateachingctu.github.io/Scientific-Programming-in-Julia/stable/lecture_03/lecture/
[6] https://www.reddit.com/r/Julia/comments/11ucqtz/whats_julias_biggest_weakness/

seabbs:
As the title this PR closes #509 by removing the remaining inner constructors that don't depend on `new` (and so need to be inner constructors).


SamuelBrand1:
I see you are ditching `scan` as well?

seabbs (commented):
Yes sorry a bit of a drive by. In looking at the code i couldn't find it being used

seabbs (commented):
Yes sorry a bit of a drive by. In looking at the code i couldn't find it being used

SamuelBrand1 (approved):

seabbs:
I think we thought we would wait or see or something but I feel like we have waited now! :)

PR #555, opened 2024-12-16
Fix: figure 1 script

SamuelBrand1:
This PR fixes the script for creating figure 1 so that it runs with the current inference data structure.

Also, I noted that it was possible for the colours used to be inconsistent; I've refactored the figure generating function to be fully `Makie` based rather than using the `AlgebraOfGraphics` frontend package to `Makie`. The downside (if in fact it is a downside) is that the `figureone` function is imperative rather than declarative.


seabbs (approved):
Bit of a shame to move away for AoG but looks good.

PR #558, opened 2024-12-16
Fix: Figure 2 script

SamuelBrand1:
This PR fixes the script for creating figure 2 so that it runs with the current inference data structure. Also, I've refactored plotting functions within `EpiAwarePipeline` where ever I've found myself reusing for figure 2 functions that seems specific to fig 1.

As per #555 I've swapped to using `Makie` directly.


seabbs (approved):
Nice looks good to me

PR #560, opened 2024-12-18
Issue 559: Diagnostic analysis over all inference runs

### LINKED ISSUE #559: Report for MCMC quality

SamuelBrand1:
This PR closes #559 

This PR adds an analysis function which returns details around the MCMC convergence:

- Proportion of all parameters get less than a bulk ess of 400 (tunable).
- Proportion of all parameters get less than a tail ess of 100 (tunable).
- Proportion of all parameters with an R hat more than 0.02 away from 1.
- The tail ess of the cluster factor for neg bin sampling. Special attention paid to this parameter because its a likely candidate for identifiability problems.

The PR also adds a script to run this across all inferences, and adds recording successes to the success/fail analysis.


pipeline/scripts/create_postprocessing_dataframes.jl
```diff
@@ -1,8 +1,11 @@
-using EpiAwarePipeline, EpiAware, JLD2, DrWatson, DataFramesMeta, CSV
+using EpiAwarePipeline, EpiAware, JLD2, DrWatson, DataFramesMeta, CSV, MCMCChains
 
 ## Define scenarios
 scenarios = ["measures_outbreak", "smooth_outbreak", "smooth_endemic", "rough_endemic"]
 
+## Define true GI means
+true_gi_means = [2.0, 10.0, 20.0]
```
seabbs, line 7: This isn't connect to this PR but I see this and the scenarios being specified in a few repeated places now?

pipeline/src/analysis/make_mcmc_diagnostic_dataframe.jl
```diff
@@ -0,0 +1,48 @@
...
+
+    #Create the dataframe
+    df = mapreduce(vcat, info.used_gi_means) do used_gi_mean
+        DataFrame(
+            Scenario = scenario,
+            igp_model = info.igp_model,
+            latent_model = info.latent_model,
+            True_GI_Mean = true_mean_gi,
+            used_gi_mean = used_gi_mean,
+            reference_time = info.reference_time,
```
seabbs, line 35: we might as well also collect some summary statistics for ESS and Rhat (i.e mean, max min or other) as well as the threshold

seabbs (dismissed):
This looks great. It would be nice to add a test to show the core mcmc diagnostic functions works in and of itself. I think it would also be good for the future to add some convergence summaries beyond the thresholds (I think having a figure on this was planned in the analysis plan but it can be a follow up issue/PR as its not on the critical path for this feature

pipeline/src/analysis/make_mcmc_diagnostic_dataframe.jl
SamuelBrand1: Nice idea

pipeline/scripts/create_postprocessing_dataframes.jl
SamuelBrand1: The goal is that this script runs the other postprocessing scipts via `include` and as I hit repetitions in either the script or underlying functions I'm moving into single calls to avoid DRY. e.g. in this PR I found myself reusing some code and created the `_get_info_from_config` internal function to do this.

So anywhere you spot repetitions are in scope to bug me about (I'm going to look for a code analysis tool to try and keep this tidy).

SamuelBrand1: In this case, I've rewritten the top script to be based on the actual pipeline types we use which seems much more robust

SamuelBrand1:
I've gone for:

- proportion pass (relative to threshold)
- mean
- min
- max

seabbs (approved):
Nice looks good to me!

PR #562, opened 2024-12-19
Issue 561: Soft min transformation

### LINKED ISSUE #561: Test avoiding `InExactError` by using `TransformObservationModel`
As a reminder this is caused by the inference expecting `Int` valued returns but instead getting caught by a `BigInt`. This is possible with latent processes in the log domain which gets `exp` transformed and then linked with a neg binomial or Poisson distribution e.g.

```julia
julia> typemax(Int) |> log
# 43.66827237527655
```

We've tried a few approaches to stop this eg #492 . 

This issue is about trying a `TransformObservationModel` composition with the usual obs processes to implement:

```math
x \to - \log ( e^{-C} + e^{-x})
```

Where $C$ is a large float. In julia this is 

```julia
using LogExpFunctions
const C = some_big_number
x -> -logaddexp(-C, -x)
```

SamuelBrand1:
This PR closes #561.

This is a small PR which simply redefines the `Ascertainment` part of the observation model composition to have a custom transformation which soft mins the _expected_ observations at `1e15` chosen so that the `InexactErrors` are not caused by wild sampling in some models.

I also increased the pipeline test draws to better check if this is operating as expected.

PS 

I've done this as an extension of the work in #560 so if @seabbs is happy with that we can do this PR as a 2-fer.

Minor change: I noted that `fetch` was still being used in the tests as a legacy of using `Dagger.jl`. This didn't error because it has a sensible `Base` default but I've removed as unnecessary.

### DIFF
--- a/pipeline/src/constructors/make_observation_model.jl
+++ b/pipeline/src/constructors/make_observation_model.jl
@@ -1,5 +1,5 @@
 """
-Constructs an observation model for the given pipeline. This is the defualt method.
+Constructs an observation model for the given pipeline. This is the default method.
 
 # Arguments
 - `pipeline::AbstractEpiAwarePipeline`: The pipeline for which the observation model is constructed.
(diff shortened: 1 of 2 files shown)


seabbs (approved):
Nice this looks good to me! Phew we should have done this a while ago. 

Note I rebased this to your target PR vs `main` to catch this is a stacked PR. 

If `fetch` was from Dagger does this mean Dagger is living in the Project.toml still and not being used? Should we remove it?

SamuelBrand1:
> If `fetch` was from Dagger does this mean Dagger is living in the Project.toml still and not being used? Should we remove it?

Dagger isn't in the Project anymore, I missed this in testing because `Base.fetch` has a just pass through fallback method https://docs.julialang.org/en/v1/base/parallel/#Base.fetch-Tuple{Any}

PR #564, opened 2024-12-19
Issue 563: Change lookback to 90 time steps

### LINKED ISSUE #563: Remove inference windowing
This issue results from f2f conversation with @seabbs .

We should remove the 35 day window for inference in the Rt-without project:

- Not part of the original analysis plan (its there as a temporary feature)
- Although standard practise for long data series, in the Rt-without we are looking at more scenarios but none of great length (70 days).

To be revisited if we look at longer time series in an extension.

SamuelBrand1:
This small PR closes #563 .

All this PR does is change the lookback to 90 time steps. This removes the windowing effect on the Rt-without project whilst leaving a more reasonable default for any extension to a longer time series.


seabbs (approved):
LGTM. This brings the pipeline back into line with the original analysis plan as discussed f2f

PR #565, opened 2024-12-19
Issue 405: model specific priors

### LINKED ISSUE #405: Model specific priors capability

SamuelBrand1:
This PR closes #405.

This PR adds a `remake_latent_model` function which is placed into the pipeline and either dispatches a pass through of the default latent model, or has a specified choice which reacts to information on mcmc diagnostics from our last full run (but not to scores).

Following f2f with @seabbs the general pattern has been to reduce a priori variance especially for difference processes and `ExpGrowthRate` models. The exception after some prior predictive checking (no comparison to data) is `DirectInfection` models which now have higher correlations when in an `AR` mode. This makes sense upon reflection since the `DirectInfection` models the residuals away from inferred log-It level.

See next comment for prior predictive sampling:


SamuelBrand1:
priorpred_gi_mean=2 0_gi_std=2 0_igp=EpiAware EpiInfModels DirectInfections_latent_model=ar_priorpredictive=true_scenario=SmoothOutbreakPipeline_tspan=1_70
priorpred_gi_mean=2 0_gi_std=2 0_igp=EpiAware EpiInfModels DirectInfections_latent_model=diff_ar_priorpredictive=true_scenario=SmoothOutbreakPipeline_tspan=1_70
priorpred_gi_mean=2 0_gi_std=2 0_igp=EpiAware EpiInfModels DirectInfections_latent_model=rw_priorpredictive=true_scenario=SmoothOutbreakPipeline_tspan=1_70
priorpred_gi_mean=2 0_gi_std=2 0_igp=EpiAware EpiInfModels ExpGrowthRate_latent_model=ar_priorpredictive=true_scenario=SmoothOutbreakPipeline_tspan=1_70
priorpred_gi_mean=2 0_gi_std=2 0_igp=EpiAware EpiInfModels ExpGrowthRate_latent_model=diff_ar_priorpredictive=true_scenario=SmoothOutbreakPipeline_tspan=1_70
priorpred_gi_mean=2 0_gi_std=2 0_igp=EpiAware EpiInfModels ExpGrowthRate_latent_model=rw_priorpredictive=true_scenario=SmoothOutbreakPipeline_tspan=1_70
priorpred_gi_mean=2 0_gi_std=2 0_igp=EpiAware EpiInfModels Renewal_latent_model=ar_priorpredictive=true_scenario=SmoothOutbreakPipeline_tspan=1_70
priorpred_gi_mean=2 0_gi_std=2 0_igp=EpiAware EpiInfModels Renewal_latent_model=diff_ar_priorpredictive=true_scenario=SmoothOutbreakPipeline_tspan=1_70
priorpred_gi_mean=2 0_gi_std=2 0_igp=EpiAware EpiInfModels Renewal_latent_model=rw_priorpredictive=true_scenario=SmoothOutbreakPipeline_tspan=1_70
priorpred_gi_mean=10 0_gi_std=2 0_igp=EpiAware EpiInfModels Renewal_latent_model=ar_priorpredictive=true_scenario=SmoothOutbreakPipeline_tspan=1_70
priorpred_gi_mean=10 0_gi_std=2 0_igp=EpiAware EpiInfModels Renewal_latent_model=diff_ar_priorpredictive=true_scenario=SmoothOutbreakPipeline_tspan=1_70
priorpred_gi_mean=10 0_gi_std=2 0_igp=EpiAware EpiInfModels Renewal_latent_model=rw_priorpredictive=true_scenario=SmoothOutbreakPipeline_tspan=1_70
priorpred_gi_mean=20 0_gi_std=2 0_igp=EpiAware EpiInfModels Renewal_latent_model=ar_priorpredictive=true_scenario=SmoothOutbreakPipeline_tspan=1_70
priorpred_gi_mean=20 0_gi_std=2 0_igp=EpiAware EpiInfModels Renewal_latent_model=diff_ar_priorpredictive=true_scenario=SmoothOutbreakPipeline_tspan=1_70
priorpred_gi_mean=20 0_gi_std=2 0_igp=EpiAware EpiInfModels Renewal_latent_model=rw_priorpredictive=true_scenario=SmoothOutbreakPipeline_tspan=1_70

pipeline/src/constructors/make_model_priors.jl
```diff
@@ -16,7 +16,7 @@ deviation 1e-1.
 
 """
 function make_model_priors(pipeline::AbstractEpiAwarePipeline)
-    transformed_process_init_prior = Normal(0.0, 0.25)
+    transformed_process_init_prior = Normal(0.0, 0.1)
```
seabbs, line 19: This reduces the initial prior not the std?

seabbs (commented):
Sorry I haven't reviewed the priors in detail. Will do so in the morning. The implantation seems correct but um yeah a bit horrifying. 

Did you consider using double dispatch on remake latent model by infection model and latent model then using accessors to update the latent model?

pipeline/src/constructors/make_model_priors.jl
SamuelBrand1: Thats right

SamuelBrand1:
> Sorry I haven't reviewed the priors in detail. Will do so in the morning. The implantation seems correct but um yeah a bit horrifying.

I agree, the basic problem was that we identified:

- Diff AR is quite a different process to AR (obviously) because, approximately, its variance growth like $t$.
- `ExpGrowthRate` is quite different to other IGPs because its `cumsum` on the log-scale.
- `DirectInfections` on log-scale is therefore one less diff version of `ExpGrowthRate`.

Once I started reasoning on that you quite quickly end up with this horrible special case prior implementation. I couldn't think of a smooth dispatch, although maybe I just need to think a bit more.

Tidying this:
- Treating the default as one good choice at least reduces the bloat.
- Reasoning a bit more on how we expect the variance to be a priori and then reversing back towards a good choice of prior dists on the parameters would be much smoother.

pipeline/src/constructors/make_model_priors.jl
```diff
@@ -16,7 +16,7 @@ deviation 1e-1.
...
     std_prior = HalfNormal(0.025)
     damp_param_prior = Beta(1, 9)
```
seabbs, line 21: I thought we discussed changing this default?

pipeline/src/constructors/make_model_priors.jl
```diff
@@ -16,7 +16,7 @@ deviation 1e-1.
 
 """
 function make_model_priors(pipeline::AbstractEpiAwarePipeline)
-    transformed_process_init_prior = Normal(0.0, 0.25)
+    transformed_process_init_prior = Normal(0.0, 0.1)
     std_prior = HalfNormal(0.025)
```
seabbs, line 20: I thought we discussed changing this default?

pipeline/src/constructors/remake_latent_model.jl
```diff
@@ -0,0 +1,99 @@
...
+
+"""
+function remake_latent_model(inference_config::Dict,
+        pipeline::AbstractRtwithoutRenewalPipeline; tscale = sqrt(21.0))
+    #Baseline choices
+    prior_dict = make_model_priors(pipeline)
+    igp = inference_config["igp"]
+    latent_model_name = inference_config["latent_namemodels"].first
+
+    if latent_model_name == "diff_ar"
```
seabbs, line 34: Can we have an issue to refactor this to use dispatch and update the models via accessors?

pipeline/src/constructors/remake_latent_model.jl
```diff
@@ -0,0 +1,99 @@
...
+        if igp == ExpGrowthRate
+            ar = AR(damp_priors = [prior_dict["damp_param_prior"]],
```
seabbs, line 36: The structure of this code now makes me thing we should be dispatching twice with a remak_latent(IGP) and a remake(IGP, latent) within then and be updating the prior dict not the latnet model (and then defining the latent model once).

pipeline/src/constructors/remake_latent_model.jl
```diff
@@ -0,0 +1,99 @@
...
+                std_prior = HalfNormal(0.005 / tscale),
+                init_priors = [prior_dict["transformed_process_init_prior"]])
+            diff_ar = DiffLatentModel(;
+                model = ar, init_priors = [prior_dict["transformed_process_init_prior"]])
+            return diff_ar
+        elseif igp == Renewal
+            ar = AR(damp_priors = [prior_dict["damp_param_prior"]],
+                std_prior = HalfNormal(0.05 / tscale),
```
seabbs: I agree with the across model choices for std prior but the composition leaves something to be desired (as above). I can't really tell which model still uses the default

pipeline/src/constructors/remake_latent_model.jl
seabbs: maybe you actually need a remake_damp and so on?

pipeline/src/constructors/remake_latent_model.jl
```diff
@@ -0,0 +1,99 @@
...
+        end
+    elseif latent_model_name == "ar"
+        if igp == Renewal
+            ar = AR(damp_priors = [Beta(2, 8)],
+                std_prior = HalfNormal(0.25),
+                init_priors = [prior_dict["transformed_process_init_prior"]])
+            return ar
+        elseif igp == ExpGrowthRate
+            ar = AR(damp_priors = [prior_dict["damp_param_prior"]],
+                std_prior = HalfNormal(0.025),
```
seabbs, line 65: this is the default

pipeline/src/constructors/remake_latent_model.jl
```diff
@@ -0,0 +1,99 @@
...
+            return diff_ar
+        elseif igp == ExpGrowthRate
+            ar = AR(damp_priors = [prior_dict["damp_param_prior"]],
+                std_prior = HalfNormal(0.005 / tscale),
+                init_priors = [prior_dict["transformed_process_init_prior"]])
+            diff_ar = DiffLatentModel(;
+                model = ar, init_priors = [prior_dict["transformed_process_init_prior"]])
+            return diff_ar
+        elseif igp == DirectInfections
+            ar = AR(damp_priors = [Beta(9, 1)],
```
seabbs, line 50: this is the reversed default? So the idea being a lot more trendy following?

pipeline/src/constructors/remake_latent_model.jl
```diff
@@ -0,0 +1,99 @@
...
+                std_prior = HalfNormal(0.05 / tscale),
+                init_priors = [prior_dict["transformed_process_init_prior"]])
+            diff_ar = DiffLatentModel(;
+                model = ar, init_priors = [prior_dict["transformed_process_init_prior"]])
+            return diff_ar
+        end
+    elseif latent_model_name == "ar"
+        if igp == Renewal
+            ar = AR(damp_priors = [Beta(2, 8)],
+                std_prior = HalfNormal(0.25),
```
seabbs, line 60: This seems like a lot? 0.25 a day change in Rt seems like a lot 0.5 change seems like loads?

pipeline/src/constructors/remake_latent_model.jl
```diff
@@ -0,0 +1,99 @@
...
+            return ar
+        elseif igp == DirectInfections
+            ar = AR(damp_priors = [Beta(9, 1)],
+                std_prior = HalfNormal(0.25),
+                init_priors = [prior_dict["transformed_process_init_prior"]])
+            return ar
+        end
+    elseif latent_model_name == "rw"
+        if igp == Renewal
+            rw = RandomWalk(
```
seabbs, line 76: So the renewal shares priors everywheere?

seabbs (commented):
I've looked again at the code and it follows what we discussed. At a minimum there needs to be information in the doc string about the specific choices that are hardcoded. I would like to see an issue to update this implementation as its very brute force and hard to reason across.

seabbs (dismissed):
@SamuelBrand1 the prior predictive plotting is super super useful here and I like how they are looking across models. I also like the use of a time scale for the non-stationary processes. I am happy to go forward with these choices with my only required action being documenting the logic near the hard coding

pipeline/src/constructors/remake_latent_model.jl
SamuelBrand1: This is a good point. I'm worried the ugly brute force approach here might lead to errors.

SamuelBrand1: I'm just writing more logic

SamuelBrand1:
Really good feedback.

Ongoing, I'm writing more logic into the doc strings :

This function sets the target standard deviation for an infection generating process (igp)
based on whether the latent process representation of its dynamics are stationary or non-stationary.

## Stationary Processes

For Renewal process `log(R_t)` in the long run a fluctuation of 0.75 (e.g. ~ 75% of the mean) is not unexpected.
For Exponential Growth Rate process `r_t` in the long run a fluctuation of 0.1 is not unexpected e.g. going from
`rt = 0.1` (7 day doubling time) to `rt = -0.1` (7 day halving time) is a 0.1 fluctuation.
For Direct Infections process `log(I_t)` in the long run a fluctuation of 1.0 (i.e an order of magnitude) is not unexpected.

## Non-Stationary Processes

For Renewal process `log(R_t)` in a single time srep run a fluctuation of 0.025 (e.g. ~ 2.5% of the mean) is not unexpected.
For Exponential Growth Rate process `r_t` in a single time step run a fluctuation of 0.005 is not unexpected.
For Direct Infections process `log(I_t)` in a single time step run a fluctuation of 0.025 is not unexpected.

SamuelBrand1:
And I'm not convinced the hard code version matches the scientific reasoning. Just doing something to check that before any commits

SamuelBrand1:
> And I'm not convinced the hard code version matches the scientific reasoning. Just doing something to check that before any commits

To address this I've got #566 to merge into this branch. I think that this makes the logic much clearer, and therefore implements what we want to implement. Albeit it needs scientific review.

seabbs (approved):

PR #574, opened 2025-01-20
trigger CI to test CI duration -- Do not merge

SamuelBrand1:
This is a test to trigger the CI and record the duration. The reason is that #569 CI is now taking a very long time, despite no evidence of slower model calls from our benchmarking.

NB: I'm not finding a slow down on running the test suite from local, and the inference examples in the documentation seem to run and render in normal time. Quite mysterious.


SamuelBrand1:
Closing this test PR with the note that all the CI runs where < 30mins.

PR #576, opened 2025-01-20
Increment DPPL to 0.31 and reduce testenv deps

SamuelBrand1:
Given the (currently) unexplained CI slowdown seen in #569; this PR simply bumps the minor version of `DynamicPPL` by 1.


SamuelBrand1:
I noted that the Test environment had a significant number of dependencies that are not currently used in testings. I have extended this PR to reduce the bloat.

SamuelBrand1:
I think the problem with very slow CI is the `depwarn=yes` argument (see #579 ). I'm inclined to close this PR if that is fixed by #580, and concentrate on bumping to DPPL 0.32.

PR #578, opened 2025-01-20
Issue 577: Remove `doctest`

### LINKED ISSUE #577: Remove `doctest` from testing
We have moved towards only checking that the documented code blocks run without failure rather than returning a specific value/string.

The documentation build action already covers this testing purpose, and the correctness of the modelling is covered elsewhere (e.g. simulation data checking reasonable parameters are inferred).

I think we can remove the `doctest` step in the testing (along with the `Documenter` dep).

SamuelBrand1:
This is a small PR which does as per title and closes #577


SamuelBrand1:
There is a problem here with `StatsPlots` not building correctly. This is no longer a used dependency in the test env so I'm going to reduce the dep bloat.

SamuelBrand1 (commented):
Self-review: Passes CI and is a minor change

Issue #573, opened 2025-01-17
Replace `@submodel` with `to_submodel`


SamuelBrand1:
As we can see from #569 the doc tests are failing only because we are getting an unexpected warning. But this might be the spur to moving to the new submodel approach of Turing?

PR #581, opened 2025-01-21
DynamicPPL bump to 0.32

SamuelBrand1:
This small PR bumps DPPL to 0.32 using the new CI approach from #580 .


SamuelBrand1:
Self review: The reason the checks didn't pass was because of timeout induced by `depwarn` not being suppressed. Enough tests passed to be confident that this worked so I am merging.

The best solution will be resolving #573

